PR #12, opened 2021-03-24
Add PicoProbe support

me-no-dev:
This PR Adds board definition and upload command to be used with PicoProbe as programmer for RP2040.

- requires OpenOCD with Pico support to be added
- not sure if the OpenOCD configs should be kept separate or not. Command currently expects them to be in the `system/openocd` folder next to the `openocd` binary itself


earlephilhower:
Since this has picoprobe support and general fixes, I've split it up.  I've manually moved the picoprobe stuff to #15 which will merge after testing and building openocd for all archs, but the remainder can be pulled in ASAP.  Thx!

PR #22, opened 2021-03-26
Add attach/detachInterrupt support

### LINKED ISSUE #21: Compile Error for attachinterrupt()
I assume the attachInterrupt() function has not yet been added, it would be great if you add it, as interrupts using micro python are really slow and thanks a lot for this repo.

earlephilhower:
Fixes #21


aayushgavande:
Works great thanks a lot.

PR #40, opened 2021-03-28
Fix INPUT_PULLUP/DOWN, I2C 0-len probes, timeout

### LINKED ISSUE #38: zero-byte writes in arduino Wire must be handled by bitbanging
a common thing done in arduino libraries is performing a zero-byte-write, like this
https://github.com/arduino-libraries/ArduinoECCX08/blob/master/src/ECCX08.cpp#L440
or in address scanning/checking

the RP2040 I2C peripheral does not permit that, so for writes of 0 bytes, it must be bitbanged, we've done so in CircuitPython and MicroPython does the same

right now, because of this bug, a lot of libraries (including a lot of Adafruit ones) will not work, they will fail on the address detection step!

(cc @sandeepmistry in case Arduino's core needs it too)

earlephilhower:
Fix the drive logic for pullup and pulldown inputs.

Add a bitbanged I2C ACK checker for 0-byte I2C transfers (since the HW
doesn't support it).

Add a timeout on I2C reads and writes so they don't hang forever if a
device is not present.

Fixes #38

### DIFF
--- a/cores/rp2040/wiring_digital.cpp
+++ b/cores/rp2040/wiring_digital.cpp
@@ -21,6 +21,8 @@
 #include "Arduino.h"
 #include <hardware/gpio.h>
 
+static PinMode _pm[30];
+
 extern "C" void pinMode( pin_size_t ulPin, PinMode ulMode ) {
     switch (ulMode) {
         case INPUT:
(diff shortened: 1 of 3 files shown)


ladyada (approved):
this fixes Adafruit BusIO based libraries!

PR #44, opened 2021-03-28
Adafruit Feather, overclocking, flash FS shims

### LINKED ISSUE #42: Need new pico-sdk with fixed configuration for Adafruit Feather, custom boot2 and libpico
Hi,

I would appreciate if you'll support Adafruit Feather RP2040 which has 8MB flash.
I've found in the code where it says that the pico has only 2048k flash, but I don't know if changing this to 8096k will make it work.

Thanks
Yoram

earlephilhower:
Add support for Adafruit Feather RP2040 (8MB).  Identified with a unique
USB PID so it displays properly once programmed one time.  Moved LED to
pin 13 (per the website docs), but have no board to test.

Add over/underclocking menus, applied at boot.  Use at your own risk, as
usual.

Add shims to allocate flash space for a filesystem (but not implemented
yet).

Add a "generic" RP2040 board

Fixes #42


yoramrotbach:
Hi Earl,
I've installed 0.9.10 and changed the board both to Feather and Generic but no luck

Yoram

earlephilhower:
What do you expect to change, other than the 8MB flash?  If you re-flash usingf the Feather board and re-upload a script, the IDE will report it as the an Adafruit Feather in the Ports section, but that's pretty much the only difference.  Oh, and the flash size reported at the end of builds will be 8MB not 2MB.

yoramrotbach:
Well I've expected sketches to start working. Nothing different is observed after compiling and uploading code generated for the Feather RP2040. The sketch doesn't seem to run, the LED isn't blinking and no /dev/tty<whatever> is available.
From displayed text perspective you are right, it does show the board name and flash amount as being 8MB

PR #55, opened 2021-03-29
Add picoprobe upload support, cleanup up boards

earlephilhower:


boards.txt
```diff
@@ -94,6 +100,103 @@ rpipico.menu.dbglvl.All=All
 rpipico.menu.dbglvl.All.build.debug_level=-DDEBUG_RP2040_WIRE -DDEBUG_RP2040_SPI -DDEBUG_RP2040_CORE
 rpipico.menu.dbglvl.NDEBUG=NDEBUG
 rpipico.menu.dbglvl.NDEBUG.build.debug_level=-DNDEBUG
+rpipicopicoprobe.name=Raspberry Pi Pico (Picoprobe)
+rpipicopicoprobe.vid.0=0x2e8a
+rpipicopicoprobe.pid.0=0x000a
```
me-no-dev, line 105: pid of picoprobe is `0x0004`

earlephilhower: Yes, but this is the VID/PID of the board we're uploading to/reading the serial port of and used to populate the Tools->Port menu.  We don't actually need to identify the picoprobe board at all...openocd finds it itself (by pid, I guess).

me-no-dev: the only difference will be that the PicoProbe will show with name in the ports menu

earlephilhower: Yes, but unfortunately then the Serial Monitor won't work (will connect to the picoprobe port).  I think maybe you're running a HW UART and not USB, but I think most people are using the USB one, so IMHO this is the more common case...

me-no-dev: When I am using the PicoProbe, then I am using HW UART0 on the target (through the PicoProbe) I do not think that it's uncommon practice either. Why use USB UART if HW one is available? And what to do with USB Host sketches?

me-no-dev: Also, since you have board definitions for USB Pico, it will show in the ports named as such anyway (and the probe will show as probe)

earlephilhower: D'oh!  I seriously did not know you could use the UART ports over SWD, too.  That's what I get for reading the documentation at 3AM in a fit of insomnia.

You make good sense.  Let me see about swapping things around in the next release or so.

PR #61, opened 2021-03-30
Update wiring_analog.cpp

smischny:
Fixed the calculation of the pwm clock divider to be the system clock hz divided by the product of the analog frequency and scale.


earlephilhower:
Thanks!

PR #62, opened 2021-03-30
fix: make the package manager JSON arduino-cli friendly

urish:
The Arduino CLI chokes when trying to install the current version. It spits the following error:

```
Error updating core and libraries index: invalid package index in https://github.com/earlephilhower/arduino-pico/releases/download/global/package_rp2040_index.json: json: invalid number literal, trying to unmarshal "\"\"" into Number
```

You can reproduce it by installing the arduino CLI, then adding the URL to your `arduino-cli.yaml` file:

```yaml
board_manager:
  additional_urls:
  - https://github.com/earlephilhower/arduino-pico/releases/download/global/package_rp2040_index.json: json
```

It seems like it gets confused by several files which has size set to "" (an empty string). 

This PR sets size to "0" in case it's empty, making the Arduino CLI accept the JSON file. Not sure why the size was empty in the first place - perhaps there's a deeper bug that hides here, but I wasn't able to get the `build_boards_manager_package.sh` script to run locally, so I couldn't investigate further.

p.s. thanks for making this Pico Core for Arduino! 

I'm planning to talk about it next week in a [live stream where I code an emulator for the pico](https://hackaday.io/project/177082-raspberry-pi-pico-emulator)

### DIFF
--- a/package/build_boards_manager_package.sh
+++ b/package/build_boards_manager_package.sh
@@ -130,7 +130,7 @@ jq_arg=".packages[0].platforms[0].version = \"$visiblever\" | \
 
 if [ -z "$is_nightly" ]; then
     jq_arg="$jq_arg |\
-        .packages[0].platforms[0].size = \"$size\" |\
+        .packages[0].platforms[0].size = \"${size-0}\" |\
         .packages[0].platforms[0].checksum = \"SHA-256:$sha\""
 fi
 


earlephilhower:
Thanks for the find and PR!

However this patches over a symptom and not the cause, which is actually the JSON is busted because my toolchain build had a hiccup.  The files do exist, so I think putting a 0-byte size will make a tool somewhere blow up when they do an integrity check on the OpenOCD bins.

Let me close this, manually put in the proper sizes, and re-upload the JSON.

For your emulator, do you have any idea when you're going to get to the PIOs?  I have lots of ideas, but debugging PIO code is difficult/impossible in my experience...

urish:
Thanks for the speedy response! 

Since I'm only working on the Emulator on the weekly stream, it's been going a bit slow. The good news is that PIO is pretty high on the list. Right now the contenders are: 

- GPIO output (and maybe input too? that's a lower priority, but probably makes sense to implement before PIO)
- USB, so we can get UART-over-USB and display debug prints, etc.
- Timers (so we can get stuff like blink to work?)
- PIO 

It's hard to tell how long each item is going to take, but there's a good chance we can start working on PIO some time during April. So I wouldn't hold my breath, but 4 to 6 weeks until we get a sounds like a reasonable time frame for me. 

What kind of issues have you been running into with PIOs?

We'll probably need some demo code to test our implementation. There are a few examples on the datasheet, and I see you also used it for tone and the Servo library.

earlephilhower:
FYI, the JSON was just updated with the real sizes so if you can convince `arduino-cli` to re-download it you can verify it's good for you.

For PIOs, I'll keep my eyes on your project then.  I won't say I'm having issues other than it's hard to map the FSM I want to build onto the PIO code, and the only way to debug is to run and grab waveforms w/a logic analyzer.  So, no bugs or anything on their side, just an inconvenient way of developing.

urish:
like a charm, thanks!

```
$ arduino-cli update
Updating index: package_index.json downloaded                                                                                                          
Updating index: package_index.json.sig downloaded                                                                                                      
Updating index: package_esp32_index.json downloaded                                                                                                    
Updating index: package_rp2040_index.json downloaded                                                                                                   
Updating index: library_index.json downloaded                                                                                                          
```

> For PIOs, I'll keep my eyes on your project then. I won't say I'm having issues other than it's hard to map the FSM I want to build onto the PIO code, and the only way to debug is to run and grab waveforms w/a logic analyzer. So, no bugs or anything on their side, just an inconvenient way of developing.

That's super useful input. So basically, you'd want a way to debug the PIO code in the emulator, right?

earlephilhower:
> That's super useful input. So basically, you'd want a way to debug the PIO code in the emulator, right?

If I could get a register dump and pin status every cycle that would be awesome.  Right now, I've got the GPIO output pins to look at (on the logic analyzer) and nothing else.

urish:
> If I could get a register and fifo dump and pin status every cycle that would be awesome. Right now, I've got the GPIO output pins to look at (on the logic analyzer) and nothing else.

👍

Doesn't sound too far-fetched. I will shoot you a note once we start working on the PIO part

urish:
> Doesn't sound too far-fetched. I will shoot you a note once we start working on the PIO part

Heads up: I'm starting to work on the PIO tomorrow in the live stream: https://youtu.be/HUW8q__jqr4

earlephilhower:
Awesome, good luck!

PR #73, opened 2021-04-02
Add I2S output support and I2S class/library

earlephilhower:
Using the PIO-driven I2S from pico-extras, add I2S output support.

Be sure to `git submodule update --init` to get the new directory


blurfl:
in libraries/rp2040/keywords.txt, shouldn't 'readAnalogTemp' be 'analogReadTemp'?

earlephilhower:
> in libraries/rp2040/keywords.txt, shouldn't 'readAnalogTemp' be 'analogReadTemp'?

Yes it should be.  Want to throw a PR?

PR #89, opened 2021-04-07
Add SDK multicore example

Bodmer:
Based on an online tutorial, this example uses both RP2040 processors.

### DIFF
--- a/libraries/rp2040/examples/SDK/Multicore/Temperature/Temperature.ino
+++ b/libraries/rp2040/examples/SDK/Multicore/Temperature/Temperature.ino
@@ -1,0 +1,98 @@
+/*
+  Temperature
+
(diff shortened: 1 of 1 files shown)


earlephilhower:
I think this is a good idea, but I really do want to add a safe way of handling this before merging.  It would replace the naked multicore_launch with one that enabled a doorbell IRQ to idle CORE1 whenever CORE0 needs to write to flash.

This example is safe, but if they were to, say, do some math on a core and write out a temp result file it would go boom in a very strange way. :(

Bodmer:
Yes, I understand. No problem.

earlephilhower:
Thanks again, but closing since the setup1()/loop1() is now preferred and lets me add the flash safety bits.

PR #103, opened 2021-04-15
Set default pins for peripherals per datasheet

### LINKED ISSUE #92: Default pins for Serial/SPI/I2C not set properly
I was having issues getting any i2c sensor to work so after a lot of testing and searching i came across this issue witch is unrelated but did help me [https://github.com/earlephilhower/arduino-pico/issues/38#issuecomment-808835731](url) the tester that was pointed out also did not work.

no i saw that PICO_DEFAULT_I2C_SCL_PIN and PICO_DEFAULT_I2C_SDA_PIN ar defined but when correctly connected it does not work.

i fixed this by setting the pins at the beginring of my program 
`  Wire.setSCL(PICO_DEFAULT_I2C_SCL_PIN);//gp5
  Wire.setSDA(PICO_DEFAULT_I2C_SDA_PIN);//gp4`

now i don't know if this is an issue but seeing that they are defined you would expect it to use them if not explicitly told to use different pins right? I figured i would let you know.

earlephilhower:
Using the official Raspberry Pi Pico datasheet and the Adafruit Feather
RP2040 schematic, set the default pins for peripherals to match.

Fixes #92

### DIFF
--- a/cores/rp2040/SerialUART.cpp
+++ b/cores/rp2040/SerialUART.cpp
@@ -179,8 +179,8 @@ SerialUART::operator bool() {
     return _running;
 }
 
-SerialUART Serial1(uart0, 0, 1);
-SerialUART Serial2(uart1, 4, 5);
+SerialUART Serial1(uart0, PIN_SERIAL1_TX, PIN_SERIAL1_RX);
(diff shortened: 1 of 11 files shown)


nuwaveit:
shouldn't SPI1s MOSI be 11, not 15?

PR #114, opened 2021-04-26
PWM functions definitions missing

oguzhanbaser:
Some pwm functions could not usable without this definitions.


earlephilhower (approved):
Thanks!

PR #130, opened 2021-05-13
SF_printable_fix

drmpf:
User requested SafeString library support be added for this project.
https://forum.arduino.cc/t/libraries-updated-to-support-new-rp2040-based-boards-raspberry-pi-pico-etc/860353/3
Current SafeString library works for Arduino Mbed OS RP2040 V2.0.0 board install

This pull request fixes missing Printable.h file
But in order to complete adding the support need a unique define (-D... = .. ) that can be detected by SafeString.h so that the
necessary 
namespace arduino {
can be set
 Arduino Mbed OS RP2040 V2.0.0 has a -DARDUINO_ARCH_MBED_RP2040  define

Can you add a -D to your compile line (and when Arduino detects libraries)
Also dtostrf( ) declaration not found by SafeString library (works with Arduino pico, fixed by an extern ... in SafeString.h )

### DIFF
--- a/cores/rp2040/Printable.h
+++ b/cores/rp2040/Printable.h
@@ -1,0 +1,1 @@
+#include "api/Printable.h"


earlephilhower:
I've submitted the -D additions with
https://github.com/PowerBroker2/SafeString/pull/35

I don't understand the `dtostrf` request, it is part of the deprecated AVR stuff and in the Arduino-supplied header already.  I also don't see it being used in the SafeString library.

earlephilhower:
````
earle@server:~/Arduino/hardware/pico/rp2040$ grep -r dtostrf ~/Arduino/libraries/SafeString/
earle@server:~/Arduino/hardware/pico/rp2040$ grep -r dtostr ~/Arduino/libraries/SafeString/
earle@server:~/Arduino/hardware/pico/rp2040$ 
````

earlephilhower:
#131 adds the non ISO stuff to standard Arduino includes, so covers dtostrf, even though I don't see it used...

Issue #88, opened 2021-04-07
No HID support for keyboard emulation


LegendEag1e:
It looks like a great tool for the pico but there is no support for hid

earlephilhower:
Fixed with #132.  The Arduino example code and my new ones work well now.

PR #132, opened 2021-05-14
Add support for shared USB Serial, Keyboard, Mouse

earlephilhower:
Use a shared infrastructure based on TinyUSB, allow users to use sketches
with ported Arduino Keyboard and Mouse libraries.


earlephilhower:
Fixes #88 

Simple testing of the Circle and Password examples seems to run just fine, w/o any disruption to the shared USB Serial port.

PR #137, opened 2021-05-16
Massively reduce generated code and RAM size

earlephilhower:
Older builds included C++ locale information in RAM/flash as well as a
bunch of exception code which can't get called since they're not enabled.

Remove them by adjusting final link command and not instantiating them
in main.cpp in the first place.

Blink went from:
> Sketch uses 215604 bytes (10%) of program storage space. Maximum is 2093056 bytes.
> Global variables use 18152 bytes (6%) of dynamic memory, leaving 243992 bytes for local variables. Maximum is 262144 bytes.

To:
> Sketch uses 56112 bytes (5%) of program storage space. Maximum is 1044480 bytes.
> Global variables use 12152 bytes (4%) of dynamic memory, leaving 249992 bytes for local variables. Maximum is 262144 bytes.

For a savings of **155KB of flash** and **6KB of RAM**


urish:
Thanks for this Earle! 

I updated the core version on wokwi.com and tested [a project with Pico + LCD1602](https://wokwi.com/arduino/projects/297323005822894602) and this is a massive improvement indeed. The program code is almost 4 times smaller!

Before:
> Sketch uses 215080 bytes (10%) of program storage space. Maximum is 2093056 bytes.
> Global variables use 17120 bytes (6%) of dynamic memory, leaving 245024 bytes for local variables. Maximum is 262144 bytes.

Now:
> Sketch uses 56960 bytes (2%) of program storage space. Maximum is 2093056 bytes.
> Global variables use 12184 bytes (4%) of dynamic memory, leaving 249960 bytes for local variables. Maximum is 262144 bytes.

earlephilhower:
No problem!  It really bugged me and I had a bout of insomnia to dig into it.

Basically every sketch should free up ~6000 bytes of RAM and ~155KB of flash with this change.

urish:
> No problem! It really bugged me and I had a bout of insomnia to dig into it.

Yeah, I know that feeling. And an usual, it turns out to be a relatively minor change in the code that leads to a huge difference...

PR #151, opened 2021-05-20
Fix debug level "Wire" compile error

blurfl:
if Debug Level "Wire" is chosen, the compiler posts error

arm-none-eabi-g++: error: DDEBUG_RP2040_WIRE: No such file or directory

~~Adding the missing '-' in boards.txt corrects the issue.~~
The correction must be made in the makeboards.py tool which generates boards.txt.


earlephilhower (changes requested):
The generator script needs to be fixed and rerun.  If we just update the boards.txt, changes will be lost on a rebuild.

https://github.com/earlephilhower/arduino-pico/blob/0d593252e622a9268feeff725b09ef1c8d2b2240/tools/makeboards.py#L27

earlephilhower (approved):
Thx!

blurfl:
I seem to have wandered out of my lane here and come afoul of the initial stages of CI development. It might be best to close this PR so I don't muddy the waters.

Issue #167, opened 2021-05-25
compile error when using Serial with USB stack set to TinyUSB


skybrian:
I get a compile error if I print to Serial with TinyUSB enabled:

    //#include <Adafruit_TinyUSB.h>
        
    void setup() {
          pinMode(LED_BUILTIN, OUTPUT);
          Serial.begin(115200);
          while (!Serial) {}
          Serial.println("Ready.");
    }
        
    void loop() {
          digitalWrite(LED_BUILTIN, HIGH);
          delay(1000);
          digitalWrite(LED_BUILTIN, LOW);
          delay(1000);
    }

Here are the errors:

    sketch_may24f.ino.cpp.o: in function `setup':
    sketch_may24f.ino:5: undefined reference to `_ZN17Adafruit_USBD_CDC5beginEm'
    sketch_may24f.ino:6: undefined reference to `_ZN17Adafruit_USBD_CDCcvbEv'
    sketch_may24f.ino:8: undefined reference to `Serial'

The workaround is easy: include the TinyUSB header file in your sketch (uncomment the include).

hathach:
It is a must, Serial is based on USB CDC which is also provided by TinyUSB library as well. Therefore you need to include the library header for it to compile and link. For one code fit all you should use

```C
#ifdef USE_TINYUSB
#include <Adafruit_TinyUSB.h>
#endif
```

earlephilhower:
I think it may be best just to put that ifdef inside `Arduino.h` to simplify things.  I was going to put a PR in later today.

earlephilhower:
Arduino won't scan libraries if they're called from the cores/... folder, as you said.  The only option would be to move the library into cores/rp2040/... and then bracket every single source file with `#ifdef USE_TINYUSB` which is not really acceptable.

Thanks @skybrian  for #168 doc update!

hathach:
> Arduino won't scan libraries if they're called from the cores/... folder, as you said. The only option would be to move the library into cores/rp2040/... and then bracket every single source file with `#ifdef USE_TINYUSB` which is not really acceptable.
> 
> Thanks @skybrian for #168 doc update!

one of the tip is just add it to Wire and SPI and other commonly used libraries within the included cores, since those are commonly used, it will trigger arduino to link the library in most sketches 

https://github.com/adafruit/ArduinoCore-samd/blob/master/libraries/SPI/SPI.cpp#L25

earlephilhower:
[...]

PR #168, opened 2021-05-26
Add TinyUSB usage note.

skybrian:
See #167.

### DIFF
--- a/docs/usb.rst
+++ b/docs/usb.rst
@@ -26,12 +26,20 @@ Examples are provided in the Adafruit_TinyUSB_Arduino for the more
 advanced USB stack.
 
 To use Serial with TinyUSB, you must include the TinyUSB header in your
-sketch to avoid a compile error. If you need to be compatible with the
+sketch to avoid a compile error.
+
+.. code:: cpp
+
+    #include <Adafruit_TinyUSB.h>
+
+If you need to be compatible with the
 other USB stack, you can use an ifdef:
 
-#ifdef USE_TINYUSB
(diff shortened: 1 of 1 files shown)


earlephilhower (approved):

docs/usb.rst
```diff
@@ -25,7 +25,23 @@ Adafruit TinyUSB Arduino Support
...
+If you need to be compatible with the
+other USB stack, you can use an ifdef:
+
+.. code:: cpp
+
+    #ifdef USE_TINYUSB
+    #include <Adafruit_TinyUSB.h>
+    #endif
+
+Also, this stack requires sketches to manually call
```
hathach, line 44: Does manually call Serial.begin() is required. I remember do that in the background , will check it is still the case

earlephilhower: The current main.cpp doesn't call `Serial.begin()` if `USE_TINYUSB` is set:

https://github.com/earlephilhower/arduino-pico/blob/a1c9dfe84d3c5070a2a7828618ebf856f84e6b9f/cores/rp2040/main.cpp#L64-L74

I should also probably put in a baud rate there, too.  D'oh.

hathach: it is a bit complicated, but the lib is designed to be one hook called only. It will start the Serial whenever the device is initialized 
https://github.com/adafruit/Adafruit_TinyUSB_Arduino/blob/master/src/arduino/Adafruit_USBD_Device.cpp#L220

main.cpp can explicitly call Serla.begin() afterwards without any issue if design ( if will just check a bool then return).

PR #173, opened 2021-05-26
docs: clarify pin assignments for i2c

skybrian:
Oops, might want to squash this before merging.


earlephilhower (approved):
Thanks.  I still need to add logic to make things crash when illegal pins are selected, but this will cover for now...

PR #185, opened 2021-06-02
fix tone mutex

### LINKED ISSUE #121: Tone problems /?request?
Thanks for the work to get support for RP2040 working in Arduino IDE. Very much appreciated. Attached is test code. I am using BSP 1.1.0 on a standard / reference PI PICO. Arduino IDE is 1.8.13 (Windows Store 1.8.42.0)

I am having trouble getting tone to work well on the 2040. There appear to be two problems.
1. Processor crashes (not PIO) when tone is in use. I am able to loop through code 14902 times and then the processor stops.  The test code  does only 4 things.
-read and scale analog input (pot)  
-noTone / set tone to scaled analog value
-increment loop counter
-print values to usb console
After processor code execution stops PIO tone generation continues. Jitter and long/short pulses stop as expected - you get a nice jitter free frequency. USB serial monitor goes off line and the IDE cannot reach the 2040 to upload programs. Power cycle and if you start upload before next crash you can load new code. As loop execution time goes down, the number of times through the loop before crash increases. There is no significant variation in loop count (1) for a given code loop.

2. There appears to be no way to do 'synchronous' updates to the waveform. That is to say, change the frequency on the next transition. As a result any update to tone (all require a noTone) cuts off the waveform presently being generated such that sometimes very short or extra long pulses are generated and there is a lot of jitter in the waveform. Changing my code to update only when there is a change in frequency will not solve problem. Possibly lessen but you will still have the same problem when control value is alternating rapidly between values 1 apart. This will not be audible unless there are stray short and long period pulses as is the case now.

It looks like the hardware has sufficient registers to prevent this problem. How exactly to do this in the Arduino IDE is beyond my ability. I am just getting my feet wet. 

I have another very closely related issue. I would like to use a piezo speaker connected to produce audio from tone. With the caveats above, this works. However, as you might expect, the speaker performs less well at higher frequencies than at very low ones. The volume change across the spectrum is substantial. In an effort to help minimize this, I would like to try narrowing the pulse width (less energy available) at lower frequencies as _part_ of a scheme to level and possibly even control volume. I would try this with PWM but I haven't found handles to adjust frequency of PWM waveform to test this idea.

Thanks again and I will be happy to test or post an issue if I should. First I wanted to make certain I am not missing something obvious to those who have more experience.

[ToneTestAlpha0.0.1.txt](https://github.com/earlephilhower/arduino-pico/files/6438262/ToneTestAlpha0.0.1.txt)

hathach:
this pr fix mutex issue when calling tone() on the same pin with different frequency/duration e.g
- tone(pin, 100, 1000) then
- tone(pin, 100, 2000) it will then call noTone(pin) first before create the new Tone object

The issue is mutex is acquired before calling noTone() causing the mutex failed on noTone() therefore the object is never release. This will cause memory leak in the long run and eventually crash the mcu. Therefore fix #121 

Note: this Pr also add ToneHappyBirth to play the tune on the buzzer that some board is equipped with. This can help to quickly notice the issue in another tone() should it happen.

@ladyada

### DIFF
--- a/cores/rp2040/Tone.cpp
+++ b/cores/rp2040/Tone.cpp
@@ -48,6 +48,11 @@ void tone(uint8_t pin, unsigned int frequency, unsigned long duration) {
         return;
     }
 
+    auto entry = _toneMap.find(pin);
+    if (entry != _toneMap.end()) {
+        noTone(pin);
+    }
+
     // Ensure only 1 core can start or stop at a time
     CoreMutex m(&_toneMutex);
     if (!m) {
@@ -60,10 +65,6 @@ void tone(uint8_t pin, unsigned int frequency, unsigned long duration) {
     }
     // Even phases run forever, odd phases end after count...so ensure its odd
     int phases = duration ? (duration * 1000 / us) | 1 : 2;
-    auto entry = _toneMap.find(pin);
-    if (entry != _toneMap.end()) {
-        noTone(pin);
-    }
 
     auto newTone = new Tone();
     newTone->pin = pin;
--- a/cores/rp2040/Tone.cpp
+++ b/cores/rp2040/Tone.cpp
@@ -48,11 +48,6 @@ void tone(uint8_t pin, unsigned int frequency, unsigned long duration) {
         return;
     }
 
-    auto entry = _toneMap.find(pin);
-    if (entry != _toneMap.end()) {
-        noTone(pin);
-    }
-
     // Ensure only 1 core can start or stop at a time
     CoreMutex m(&_toneMutex);
     if (!m) {
@@ -65,6 +60,10 @@ void tone(uint8_t pin, unsigned int frequency, unsigned long duration) {
     }
     // Even phases run forever, odd phases end after count...so ensure its odd
     int phases = duration ? (duration * 1000 / us) | 1 : 2;
+    auto entry = _toneMap.find(pin);
+    if (entry != _toneMap.end()) {
+        noTone(pin);
+    }
 
     auto newTone = new Tone();
     newTone->pin = pin;
--- a/libraries/rp2040/examples/ToneHappyBirthday/ToneHappyBirthday.ino
+++ b/libraries/rp2040/examples/ToneHappyBirthday/ToneHappyBirthday.ino
@@ -59,26 +59,25 @@ uint8_t const pin_buzzer = PIN_BUZZER;
 
 // Play a note of the specified frequency and for the specified duration.
 // Hold is an optional bool that specifies if this note should be held a
-// little longer, i.e. for eigth notes that are tied together.
+// little longer, i.e. for eighth notes that are tied together.
 // While waiting for a note to play the waitBreath delay function is used
 // so breath detection and pixel animation continues to run.  No tones
 // will play if the slide switch is in the -/off position or all the
 // candles have been blown out.
-void playNote(int frequency, int duration, bool hold=false, bool measure=true) {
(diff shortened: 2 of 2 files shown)


cores/rp2040/Tone.cpp
```diff
@@ -48,6 +48,11 @@ void tone(uint8_t pin, unsigned int frequency, unsigned long duration) {
         return;
     }
 
+    auto entry = _toneMap.find(pin);
```
earlephilhower, line 51: Thanks for the PR.  I don’t know how I missed it.

Unfortunately, though, this section needs to be kept in a mutex protected block because the std::map is not thread safe.  A second Subblock could be done with the CoreMutex here, just using brackets to set the mutex lifetime to only be the find and end check.

libraries/rp2040/examples/ToneHappyBirthday/ToneHappyBirthday.ino
```diff
@@ -0,0 +1,143 @@
+// The speaker will play the tune to Happy Birthday continuously
```
earlephilhower, line 1: Thanks for the additional example!

earlephilhower (changes requested):
I think we still have a minor problem, but a big thanks for catching and helping fix this bug!

cores/rp2040/Tone.cpp
earlephilhower: It's not elegant, but something like this is what I was suggesting:

````
            bool needStop;
            {
                CoreMutex m(&_toneMutex);
                auto entry = _toneMap.find(pin);
                needStop = entry != _toneMap.end();
            }
            if (needStop) {
                noTone(pin);
            }
````

hathach: that works but not really explicit, how about we add explicit enter()/exit() for mutex. There maybe more of this situation later on. Though it is totally up to you.

earlephilhower: FWIW I think the CoreMutex class is actually more explicit/simpler than the enter/exit.  It's a wrapper that, for the object lifetime, grabs the mutex and releases it on any exit of the block.  It's got the anti-deadlock stuff from `pico-sdk`, which is what the `operator bool` is doing, although I'm thinking having it immediately `panic()` on deadlock may instead be the right thing to do...

hathach: I agreed, I mean we should add .enter() and .exit() to the CoreMutex. For this case, we can actively release the mutex before  calling the noTone() and re-claim it afterward. That would make it easier to understand than replying on the destructor()

https://github.com/earlephilhower/arduino-pico/blob/master/cores/rp2040/CoreMutex.h#L50

earlephilhower:
@hathach I do apologize, but your PR fixing my silly bug made me revisit the tone generator completely.  The new PR #186 fixes a longstanding bug here and avoids `tone()` needing to call `noTone()` completely, so I think this PR might just need to be shelved.

Of course, you'll probably just go and find another big in the new tone PIO program/code.  :)

earlephilhower:
Actually, just the `Tone.cpp` bits of the PR won't be valid.  The example would still be very nice to have!

hathach:
> Actually, just the `Tone.cpp` bits of the PR won't be valid. The example would still be very nice to have!

No problem, I have reverted the tone changes. Also tested your #186 with the example, everything is perfect.

earlephilhower (approved):
Thanks again!

PR #191, opened 2021-06-06
Add a new board - the Adafruit Macropad w/RP2040

ladyada:
:)


earlephilhower (approved):
Adafruit may have more RP2040 board variants than there are RP2040 chips in circulation. :)

PR #204, opened 2021-06-10
Move to Arduino API 1.3.0

earlephilhower:


earlephilhower:
Argh, looks like the Arduino examples themselves do not pass compilation with the latest API update.  Closing for now, will revisit when IDE 2.0 is widely available.

PR #213, opened 2021-06-16
Add PDM library for Arduino Nano RP2040 Connect

earlephilhower:
Thanks @Claupio


earlephilhower:
@Claupio, where is the PDM object instantiated in the main MBED version?  Does the app need to make a `PDM` object, or is it part of the core?  Looks like the library probably needs a little patch to instantiate it and make the PDM example pass.

Claupio:
> @Claupio, where is the PDM object instantiated in the main MBED version? Does the app need to make a `PDM` object, or is it part of the core? Looks like the library probably needs a little patch to instantiate it and make the PDM example pass.

The PDM object is declared in the library in PDM.cpp. In the instance declaration there is a check to see If the PDM pins were defined. Before compiling did you replaced the arduino pins file?

earlephilhower:
Yes, that makes sense.  I run CI only on the RPi Pico board, so the PDM example obviously will fail to build there.  I've removed it and when CI goes through fine I'll merge and bump.

PR #219, opened 2021-06-23
use correct tusb_config.h for built-in usb (pico-sdk)

hathach:
I am working on updating the Adafruit TinyUSB integration with rp2040 and make it more portable. And notice that 
- There is potential issue with `usbd_desc_cfg` value may not be initialized as zero.
- `tusb_config.h` currently included is located at `pico-sdk/src/rp2_common/pico_stdio_usb/include` which is not correctly reflect the one that is used to make libpico.a
- I have also updated the `tusb_config.h` to correctly use  OPT_OS_PICO which allow the stack to use pico sdk mutex to prevent race condition. Therefore libpico.a is also rebuilt with `make-libpico.sh` (let me know if you prefer to build it yourself).


lib/platform_inc.txt
```diff
@@ -53,5 +53,5 @@
 -iwithprefixbefore/pico-sdk/src/rp2_common/pico_runtime/include
 -iwithprefixbefore/pico-sdk/src/rp2_common/pico_stdio/include
 -iwithprefixbefore/pico-sdk/src/rp2_common/pico_stdio_uart/include
--iwithprefixbefore/pico-sdk/src/rp2_common/pico_stdio_usb/include
 -iwithprefixbefore/pico-sdk/src/rp2_common/pico_unique_id/include
+-iwithprefixbefore/tools/libpico
```
earlephilhower, line 57: I understand the desire to have a single tush_config.h file, but isn't the libpico `cmake` going to pull it from the SDK directory during its build.  Then with this change it seems like the Arduino core in non-tinyusb mode will have a different header configuration, no?  Isn't that a problem?

hathach: > I understand the desire to have a single tush_config.h file, but isn't the libpico cmake going to pull it from the SDK directory during its build.

I thought `tools/libpico/tusb_config.h` is the one used for making the libpico.a since it is added by the time you enable the built-in keyboard/mouse. If yes, then it should be the one being used by built-in stack (picosdk) when compiling user sketch as well.  Maybe I misunderstand it !! 

AFAIK, sdk doesn't have any other use of tinyusb except for the sdio-usb (mostly printf retarget), which this core doesn't need at all since it use the tud_cdc() directly. 

For "Adafruit TinyUSB" library, yes, it use an different config file (already as it is now), but it shouldn't be an issue, since I will make sure the correct tusb_config.h is used for each user stack mode and linker will pick the function api by the library instead of one within libpico.

PS: I am trying to move the tusb_config.h from https://github.com/earlephilhower/arduino-pico/tree/master/cores/rp2040/TinyUSB (used by Adafruit TinyUSB) to the library itself to make it more portable and easier for user to modify the config (e.g multiple CDCs or more HIDs), without touching file in the core.

earlephilhower: Let me give this a try over the weekend.  I compile the TUSB library into the `libpico.a` and I'm pretty sure that needs a tusb_config.h file, and that the SDK Cmake uses the one in the stdio dir, but this might override.  I'll double-check the libpico build and then how it's linked w/the main core...

hathach: @earlephilhower take your time, I mostly guess since you add both new tusb_config.h with `CFG_TUD_HID=2` along with new libpico in the keyboard mouse PR https://github.com/earlephilhower/arduino-pico/pull/132/files . But yeah, I am not entirely sure.

earlephilhower: I gave it some testing locally and am finding no issue, great!  No need to rebuild, the `.a` should be identical (minus debug path info) since the gcc binaries you and I are using would be identical.

earlephilhower (approved):
Thanks!

hathach:
> Thanks!

Superb ! Thanks for merging :+1:

PR #226, opened 2021-06-28
 Add CPU cycle counter accessors

earlephilhower:
Use the 24-bit SYSTICK peripheral, wrapped in logic to extend it to a
full 32 or 64bits.  W/o the wrapper, SYSTICK will wrap around in ~100ms.

Adds rp2040.getCycleCount() and rp2040.getCycleCount64()

Clean up the libpico build process as crt0.S from the pico-sdk should
be directly used.


Andy2No:
That could be useful.  What data types do they return?

What's the approximate loop/repeat time for each one?  

Also, is there a way to get the current CPU clock frequency or cycle time, to use to convert the return value to a number of microseconds?  Not essential, but it would be nice to be able to write code that still gave correct time periods, based on that, without needing to assume the CPU clock speed, since it can be changed.

PR #228, opened 2021-06-29
Update tinyusb 1.3.0

hathach:
This PR update the core to use TinyUSB library 1.3.0 which
- Reduce the dependency from having compatible tusb_config.h within the core (previously `cores/TinyUSB/tusb_config.h`) to have it within the libarry. Make it easier for user to customize the usb config without touching files in the core (easy uprgade).
- Menu flag now has explicit include path to usb stack provided by libpico or adafruit library (previously it is implicit where the include path of the adafruit is inserted before the libpico within gcc command).

PS: For actual modification and fixes to the library, please check out  https://github.com/adafruit/Adafruit_TinyUSB_Arduino/pull/100

Note: since it include some changes in CDC interface (to better support multiple instance of CDCs), a new core release might (or might not) be needed to work with latest library release.

@ladyada


earlephilhower (approved):
Very nice cleanup!

PR #233, opened 2021-07-01
Add options to not enable USB or USB Serial

tvlad1234:
Adding additional options to the 'USB Stack' menu, which allow the user to disable USB Serial or USB altogether.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -115,6 +115,12 @@ rpipico.menu.usbstack.picosdk=Pico SDK
 rpipico.menu.usbstack.picosdk.build.usbstack_flags="-I{runtime.platform.path}/tools/libpico"
 rpipico.menu.usbstack.tinyusb=Adafruit TinyUSB
 rpipico.menu.usbstack.tinyusb.build.usbstack_flags=-DUSE_TINYUSB "-I{runtime.platform.path}/libraries/Adafruit_TinyUSB_Arduino/src/arduino"
+rpipico.menu.usbstack.picosdknoser=Pico SDK (No Serial)
+rpipico.menu.usbstack.picosdknoser.build.usbstack_flags=-DDISABLE_USB_SERIAL "-I{runtime.platform.path}/tools/libpico"
+rpipico.menu.usbstack.tinyusbnoser=Adafruit TinyUSB (No Serial)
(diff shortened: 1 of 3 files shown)


boards.txt
```diff
@@ -115,6 +115,12 @@ rpipico.menu.usbstack.picosdk=Pico SDK
 rpipico.menu.usbstack.picosdk.build.usbstack_flags="-I{runtime.platform.path}/tools/libpico"
 rpipico.menu.usbstack.tinyusb=Adafruit TinyUSB
 rpipico.menu.usbstack.tinyusb.build.usbstack_flags=-DUSE_TINYUSB "-I{runtime.platform.path}/libraries/Adafruit_TinyUSB_Arduino/src/arduino"
+rpipico.menu.usbstack.picosdknoser=Pico SDK (No Serial)
+rpipico.menu.usbstack.picosdknoser.build.usbstack_flags=-DDISABLE_USB_SERIAL "-I{runtime.platform.path}/tools/libpico"
+rpipico.menu.usbstack.tinyusbnoser=Adafruit TinyUSB (No Serial)
+rpipico.menu.usbstack.tinyusbnoser.build.usbstack_flags=-DDISABLE_USB_SERIAL -DUSE_TINYUSB "-I{runtime.platform.path}/libraries/Adafruit_TinyUSB_Arduino/src/arduino"
```
hathach, line 121: Maybe I miss it, why would you want to have No Serial option ? Anyway `DISABLE_USB_SERIAL` does not work with TinyUSB though, however Serial CDC can be omitted by just simply called `Serial.end()` when using TinyUSB like this example https://github.com/adafruit/Adafruit_TinyUSB_Arduino/blob/master/examples/CDC/no_serial/no_serial.ino#L26

Therefore I would suggest to just drop No Serial option for TinyUSB.

tvlad1234: At first I wanted to have the No USB option, then I added No Serial just for the sake of completeness.

I will follow your suggestion and remove No Serial for TinyUSB.

majbthrd:
@tvlad1234, I've submitted a PR #239 that might be of interest to you.  Thanks to the hard work of @earlephilhower, there are newer OpenOCD binaries that include CMSIS-DAP support.  The PR adds a dedicated pico-debug board types (just as there presently is for Picoprobe), and the pico-debug options have No USB like you originally proposed.

tvlad1234:
@majbthrd, the effort for #239 is much appreciated, and it's a way more elegant alternative to what I was trying to accomplish here. I will close this PR.

PR #245, opened 2021-07-11
Remove exception handling code, save RAM and FLASH

earlephilhower:
Use a libstdc++ compiled with -fno-exceptions to avoid including the
code needed to unwind C++ exceptions.

Saves 4K RAM and 6K flash.


earlephilhower:
WIP tag added.  There's something going on during boot with this new stdlib causing hangs randomly.  Need to debug. :(

earlephilhower:
Tested working well now.  Shrinks Blink.INO to `52476 bytes flash` and `7440 bytes RAM` used.
`master` gives `57768 bytes flash` and `11640 bytes RAM` used.
`MBED` gives `80072 bytes flash` and `52920 bytes RAM` used (!!!).

So, this PR will save >5K flash and >4K RAM on most projects vs. `master`.

Compared to `MBED`, it used 91% less RAM(!!!), about 71K less.

PR #258, opened 2021-07-21
Tweak README git setup instructions

obra:
git submodule update has a --recursive flag, which can potentially save users a few steps, but it looks like it may check out a bunch of extra HAL drivers inside tinyusb. 

I'm not sure if the slightly increased disk usage and clone time outweighs the easier onboarding. Please feel free to reject this PR if you deem it undesirable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,12 +51,8 @@ To install via GIT (for latest and greatest versions):
 mkdir -p ~/Arduino/hardware/pico
 git clone https://github.com/earlephilhower/arduino-pico.git ~/Arduino/hardware/pico/rp2040
 cd ~/Arduino/hardware/pico/rp2040
-git submodule update --init
-cd pico-sdk
-git submodule update --init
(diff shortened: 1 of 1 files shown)


earlephilhower:
Thanks for the PR, but I think it's probably better not to merge.  Doing full recursive pulls in a **massive** amount of stuff (mostly tinyusb sub-modules).  I just ran a quick check on my system:

````
Full Recursive PR258
    files: 38,231
    size: 2.3GB
    time: many mins

Non-recursive, using master README commands
    files: 5,448
    size: 145 MB
    time: ~ 1 min
````

Going from 145M to 2.3GB for the repo seems like a stiff price to pay, especially since folks can cut-n-paste the README to get going...

obra:
That's incredibly embarassing. I'm sorry for not checking just how much extra stuff this was pulling down. Thank you for being so gracious about it.

PR #261, opened 2021-07-23
fixed URL

ubidefeo:
you had "esrle" instead of "earle" in the GH URL :)


earlephilhower (approved):

earlephilhower:
Thanks, that was a silly mistake on my part!

ubidefeo:
my pleasure, @earlephilhower 
this kind of typos are very common, I caught it by accident :)

PR #273, opened 2021-08-10
Increase waveform resolution

slva21:
Increase the PWM resolution possible by allowing the user to enter a floating point number in the writeMicroseconds method by allowing the usToPIOCycles conversion method to accept a floating data type.


cores/rp2040/RP2040.h
```diff
@@ -44,48 +47,64 @@ class _MFIFO {
...
 
-    bool pop_nb(uint32_t *val) {
+    bool pop_nb(uint32_t *val)
+    {
         // Pop from my own FIFO
         return queue_try_remove(&_queue[get_core_num()], val);
     }
 
-    int available() {
+    int available()
```
earlephilhower, line 85: Can you please undo the formatting-only changes?  That breaks the CI style checks.

cores/rp2040/RP2040.h
```diff
@@ -118,89 +142,108 @@ class _MFIFO {
...
 
-    ~RP2040() { /* noop */ }
-
+    ~RP2040()
+    { /* noop */
+    }
 
     // Convert from microseconds to PIO clock cycles
-    static int usToPIOCycles(int us) {
+    static int usToPIOCycles(float us)
```
earlephilhower, line 162: Can you please make this two overloaded functions.  The original one with a `int us` and your addition with `float us` as a parameter instead.

Making it take a float always is very expensive in generated code and slow since in all cases but Servo it's taking an `int`.

cores/rp2040/RP2040.h
```diff
@@ -118,89 +142,108 @@ class _MFIFO {
...
+    {
         // Parenthesis needed to guarantee order of operations to avoid 32bit overflow
-        return (us * (clock_get_hz(clk_sys) / 1000000));
+        float x = (us * (clock_get_hz(clk_sys) / 1000000));
```
earlephilhower, line 165: Do we need to use a floating million to ensure float and not integer division here?  `(float)1000000.0`  Because `clock_get_hz` returns `int` and `10000000` is int, and they're evaluated first due to parenthesis, I think so.  OTW `/` is integer division and not float.

earlephilhower (changes requested):
Minor things, but looks like a good idea, thanks!

cores/rp2040/RP2040.h
slva21: Sure!

cores/rp2040/RP2040.h
slva21: hmm alright. I'll probably have to change the parameter type of the overload to a double for that to work

cores/rp2040/RP2040.h
slva21: Didn't really think about that and I'm not too sure how the division will be done but it wouldn't hurt to add it just in case

slva21:
Pushed those changes now

PR #281, opened 2021-08-12
Fix wire defines on feather

AustinMorris:
In troubleshooting issues using the STEMMA QT connector on my feather, I think I've discovered an issue with the board's variants file, specifically the WIRE lines below. The file seems to associate the i2c0 peripheral with the 2 and 3 pins, and doesn't seem to be attempting to assign any functional pins to i2c1.
https://github.com/earlephilhower/arduino-pico/blob/8451bc21ec4cca39f189a0493b2714749b272816/variants/adafruit_feather/pins_arduino.h#L29-L35

However looking at the Wires library, it doesn't seem like pins 2 and 3 are valid for i2c0 at all, but they are valid for the i2c1 peripheral.
https://github.com/earlephilhower/arduino-pico/blob/8451bc21ec4cca39f189a0493b2714749b272816/libraries/Wire/src/Wire.cpp#L50-L52
https://github.com/earlephilhower/arduino-pico/blob/8451bc21ec4cca39f189a0493b2714749b272816/libraries/Wire/src/Wire.cpp#L67-L69

This led me to try this in my script, which seems to work correctly with both the built in STEMMA QT connector, and the pins labeled SCL and SDA on the board itself. When trying these same pins with the "zero" Wire object instead of Wire1, it seemed like the program would panic, but I don't have a full debugger to 100% confirm that. This suggests to me that Wire.cpp is correct, and the pins page should be made to agree to it.
```C++
Wire1.setSDA(PIN_WIRE0_SDA);
Wire1.setSCL(PIN_WIRE0_SCL);
  ```
  
From Adafruit's pinout page from the board https://cdn-learn.adafruit.com/assets/assets/000/100/740/original/adafruit_products_RP2040Pinout300DPI.png?1615928836, it seems like there are only 4 pairs of adjacent i2c0 supported pins. Specifically 24+25, RX+TX, 12+13, and 6+9. RX+TX seem labeled for other uses, 13 is tied to the built in LED making it not a great choice here, and it seems like it'd be very easy for people to read 6 and 9 from the wrong direction and incorrectly plug their SDA line into the SCL pin and vice-versa. Therefore 24+25 seems like the a sensible default choice for the PIN_WIRE0 defines on this board. I've tested with those values on Wire "zero" and they work as expected.
 
These changes look correct to me. They allow the STEMMA QT connector to work out of the box so long as the code is using Wire1, and it provides default pins for any code that uses "zero" Wire, instead of requiring extra lines of code from the user specifying which i2c0 valid pins they want to use. All that being said, I'm not especially experienced in electronics and microcontrollers so please let me know if I've made a mistake or overlooked anything.

### DIFF
--- a/variants/adafruit_feather/pins_arduino.h
+++ b/variants/adafruit_feather/pins_arduino.h
@@ -27,12 +27,10 @@
 #define PIN_SPI1_SS    (31u)
 
 // Wire
-#define PIN_WIRE0_SDA  (2u)
-#define PIN_WIRE0_SCL  (3u)
-
-// Not pinned out
-#define PIN_WIRE1_SDA  (31u)
-#define PIN_WIRE1_SCL  (31u)
+#define PIN_WIRE0_SDA  (24u)
+#define PIN_WIRE0_SCL  (25u)
+#define PIN_WIRE1_SDA  (2u)
+#define PIN_WIRE1_SCL  (3u)
 
 #define SERIAL_HOWMANY (2u)
 #define SPI_HOWMANY    (1u)

### CODE REVIEW

**Comment by earlephilhower** [comment]
This looks good, thanks.  I have one question about it, though:  Do people expect to need to use `Wire1.xxxx` to use the connector?

I'm thinking maybe the right thing to do about this is to make the i2c0/i2c1 definable inside the variant.  So you could use `Wire.xxx` with this connector, because in the stemma.h file there'd be a `#define __WIRE0 i2c1`.  That's not a complicated change, but one that would need to percolate through all the existing variants...

**Comment by AustinMorris** [comment]
It seems like the Adafruit documentation indirectly touches on that same question. For the feather board, https://learn.adafruit.com/adafruit-feather-rp2040-pico/pinouts says that in CircuitPython, "`board.I2C()` uses SCL/SDA", which would work with the built in STEMMA QT connector. However for their QT Py RP2040 board, https://learn.adafruit.com/adafruit-qt-py-2040/pinouts says "`board.I2C()` uses SCL/SDA (not for the STEMMA QT connector!)". The top of that page also has "NOTE: To use this connector with CircuitPython, you need to use `board.SCL1` and `board.SDA1` in your code!"

So evidently they decided it was acceptable to have a board where the default i2c interface in code didn't use the STEMMA QT connector. But also given the exclamations, it seems like they think some people will expect the default i2c code interface to use the STEMMA QT connector and need to be warned when it won't.

I think it's worth avoiding that confusion if possible though. In your suggestion, would the `stemma.h` file you referenced go alongside the `pins_arduino.h` file in each variants folder? It looks like `Wire.cpp` would also need to be a bit more clever in how it assigns pins in the Wire constructors so that it flips both the i2cX objects and the associated pins.
```C++
TwoWire Wire(i2c0, PIN_WIRE0_SDA, PIN_WIRE0_SCL);
TwoWire Wire1(i2c1, PIN_WIRE1_SDA, PIN_WIRE1_SCL);
```

Could something like a `#define WIRE_INDICES_FLIPPED true` work instead of associating an i2cX object with the define directly? Then we could do
```C++
#if WIRE_INDICES_FLIPPED
TwoWire Wire(i2c1, PIN_WIRE1_SDA, PIN_WIRE1_SCL);
TwoWire Wire1(i2c0, PIN_WIRE0_SDA, PIN_WIRE0_SCL);
#else
TwoWire Wire(i2c0, PIN_WIRE0_SDA, PIN_WIRE0_SCL);
TwoWire Wire1(i2c1, PIN_WIRE1_SDA, PIN_WIRE1_SCL);
#endif
```

earlephilhower:
Thanks for the explanation.  I gave some thought to this and I think your current setup is probably the best way for now.

The thing that bothers me about swapping `Wire` and `Wire1` is that the whole "which pins are legal" starts to get fuzzy when there is not a 1:1 mapping.  Since they've already documented it, I think we should stick with that for the time being.

PR #283, opened 2021-08-14
Add support for reverse numbering of analog pins.

PontusO:
When laying out a PCB it sometimes makes it easier to reverse the connection of the analog pins. For instance when the RP2040 is placed on the back side of the board. Without this change the check for valid pins will fail.


earlephilhower:
@PontusO do we need to change your other Inventec Challenger board pinout as well?  Or was it just this one which needed the change?

PontusO:
@earlephilhower Fortunatly no, on the LTE board the RP2040 was moved to the backside of the board and to simplify routing I reversed the order of the analog pins. The WiFi board shall remain the same.

PR #302, opened 2021-09-03
reset the SDA and SCL pins to INPUT at Wire.end() to prevent multiple…

SirSydom:
… pins connected to I2C when changing pins


earlephilhower (approved):
Thanks!

PR #315, opened 2021-09-17
Change libpico.a to build with PICO_XOSC_STARTUP_DELAY_MULTIPLIER=64

### LINKED ISSUE #312: Adafruit ItsyBitsy RP2040 - sketch doesn't run after reset or unplugging, following upload
After successfully compiling and uploading a sketch to an Adafruit ItsyBitsy RP2040, with the Boot button pressed as it was plugged int, and confirming that the sketch is running immediately after the upload, I've found that it doesn't run after pressing reset or after unplugging the USB cable, then plugging it back in again.

The sketch also doesn't run if I plug the USB cable into a USB power brick, and I've confirmed that it does work that way if I put the code that it came with back on, by dragging "ItsyBitsy_RP2040.uf2" to the USB drive that I get by holding down the Boot button.

Also, after plugging the USB lead in, the serial port for it does not appear in Device Manager - that only happens after uploading a sketch after plugging it in with Boot button pressed.

I just got a couple of Adafruit ItsyBitsy RP2040, today, and it's been a struggle to get this far, due to lack of driver support (they don't appear to have added it to the last release), but the Zadig procedure from issue #96 got the serial port working, immediately after upload.  

I'm using a variation on Blink which also sends messages to the serial port, so at the very least, I should see the red LED blinking - but that only happens immediately after uploading it, until it's reset or power cycled, then it does nothing.

Is it possible that it's only loading the sketch into RAM, so it runs initially but doesn't survive a reset or unplug/re-plug?

dhalbert:
- Fixes #312.

Some samples of some Adafruit boards need a longer XOSC startup delay. We diagnosed this a while ago and PR'd a change to `pico-sdk` that adds a new compile-time constant `PICO_XOSC_STARTUP_DELAY_MULTIPLIER`, which allows lengthening the XOSC delay.

`arduino-pico` uses a single build of `libpico.a` for all boards. To accomodate the Adafruit boards, this PR sets `PICO_XOSC_STARTUP_DELAY_MULTIPLIER=64` for the build, and provides a newly built `libpico.a`. The extra delay is still quite short, and shouldn't interfere with normal startup.

I tested this on an Adafruit QT Py RP2040 that is known to need a longer startup time. THe board would not run after a power cycle or reset before this fix, and now it works fine.

See https://github.com/raspberrypi/pico-sdk/pull/457 and https://github.com/raspberrypi/pico-sdk/pull/401 for some background on this problem.

@earlephilhower I am not sure of your workflow for rebuilding `libpico.a` with `make-libpico.sh`. I built `libpico.a` manually with `make-libpico.sh` inside `.arduino15` using the lastest pico BSP, and then copied it into the commit for this PR. I had to do some manual fiddling with my `PATH` to make sure to use the compiler you include in the BSP. The `make-libpico.sh` script assumes the correct `arm-none-eabi-gcc` is on `PATH`, and also looks for tools in `system`, though my `system` only includes `system/placeholder`. So I had to set things up manually to make the script work.

Tagging @ladyada for FYI.


hathach:
@dhalbert  libpico.a is built by running this script in the tool https://github.com/earlephilhower/arduino-pico/tree/master/tools/libpico . With this change, I think we could change the boot2 of qtpy from boot2_generic_03h_4_padded_checksum to boot2_w25q080_2_padded_checksum for fast running code, I will pull and check it out.

earlephilhower (approved):
LGTM, thanks!

earlephilhower:
@dhalbert the build script actually assumes that you've got a `git clone` copy of the repo where the toolchain is in a subdir of the main core (hence the hardcoded `../../` in the PATH).

In any case, I'll pull this and give it a whirl and merge if all looks well.

PR #318, opened 2021-09-18
Adds board variant definition for Melopero Shake RP2040

leoli51:
Hi, we created and tested a board variant definition for the upcoming Melopero Shake RP2040 development board.


earlephilhower (approved):
LGTM!

PR #323, opened 2021-09-24
Reverted Wire and Wire1

leoli51:
Hi, 
We are sorry we realized that we cannot have Wire on the i2c1 pins so we had to revert the changes we made. Is there any particular reason why Wire cannot work on the i2c1 pins ( same goes with Wire1 on i2c0 pins) ?
Sorry once again for another PR and thank you very much.


earlephilhower:
>  Is there any particular reason why Wire cannot work on the i2c1 pins ( same goes with Wire1 on i2c0 pins) ?

`Wire` is connected to `i2c0` only and `Wire1` is `i2c1` only.  Because the pins are user-assignable, there's no good way to allow swapping them w/o it getting dicey.  The GPIO muxes just have limited options.

PR #324, opened 2021-09-25
Change SDA and SCL to Wire1

Prow7:
Pins 2 and 3 are on I2C1 which is Wire1. The Wire instance (I2C0) does not work with this pin assignment.


earlephilhower:
Looks like #325 supersedes this.  If not, let me know.

PR #325, opened 2021-09-25
Fix Wire pin definitions on ItsyBitsy

apendley:
* Map GPIO 2 and 3 to Wire1.
* Provide sensible defaults for Wire0 (GPIO 24 and 25)
* This is the same fix that was applied for the Feather in issue #281.


Prow7:
ah, basically the same fix I submitted, except you enabled set Wire0 to GPIO 24 and 25. Yes, this is probably ideal since it is consistent with the feather.

Should the #define WIRE_HOWMANY change to 2 then? Not sure it's used anywhere.

apendley:
Oh! I didn't realize that someone else had submitted a fix for the same issue! I was just banging my head against this tonight and remembered that the Feather had the same issue recently, so I dug into it a little. Not sure about WIRE_HOWMANY, I don't think it's used anywhere, but it probably wouldn't hurt to update it.

PR #366, opened 2021-12-06
Adds support for Challenger NB RP2040 WiFi board and RPICO32 module

PontusO:
As stated this PR adds support for both the NB version of the RP2040 WiFi board as well as support for the new RPICO32 all in one RP2040/WiFi based module.


earlephilhower (approved):
LGTM, thanks for the clean PR!

PR #367, opened 2021-12-09
Adds a lightweight support class for LTE functionality

PontusO:
The PR adds a few basic support functions such as starting the modem properly and setting its operating mode.


PontusO:
Not sure why the package refs test fails, it runs OK here locally.

earlephilhower:
> Not sure why the package refs test fails, it runs OK here locally.

Sometimes there are network hiccups.  I've restarted all jobs and they should go thru...

earlephilhower (approved):

Issue #380, opened 2021-12-23
build.mcu mentionned twice in boards.txt


rei-vilo:
The `build.mcu` variable is mentioned twice in [boards.txt](https://github.com/earlephilhower/arduino-pico/blob/master/boards.txt).

Example for `rpipico` board

https://github.com/earlephilhower/arduino-pico/blob/7120a1508c09f8d409014552d4b7009106a58623/boards.txt#L20
https://github.com/earlephilhower/arduino-pico/blob/7120a1508c09f8d409014552d4b7009106a58623/boards.txt#L32

`build.mcu` is used for libraries distributed as [pre-compiled binaries](https://arduino.github.io/arduino-cli/0.20/library-specification/#precompiled-binaries). This duplicate entry prevents from using that feature.

rei-vilo:
See PR #381

earlephilhower:
Thanks for the PR!  I've combined your changes with the proper script changes and things should be working fine now w/binary libs.

rei-vilo:
You're welcome!

PR #381, opened 2021-12-23
Allow pre-compiled libraries

rei-vilo:
Changes 
* Removed duplicated `build.mcu` entries from `boards.txt` (issue #380)
* Added `compiler.libraries.ldflags` for linking pre-compiled libraries to `platform.txt`

Reference:
* https://arduino.github.io/arduino-cli/latest/library-specification/#precompiled-binaries


earlephilhower (changes requested):
We seem to have submitted PRs at the same second here!

The change you did manually to boards.txt won't pass CI because that file is machine generated.  You need to change makeboards.py.

I'll integrate your compiler.ldflags setting, too, and we can use #380

rei-vilo:
Thank you!

PR #391, opened 2021-12-26
Add SoftwareSerial-like PIO based UARTs

earlephilhower:
Adds support to the core for PIO-based, software-created UARTs, up to 8 (the number of PIO state machines) possible.

By using a custom program on the PIO state machines, it allows for very high bit rates and does not require CPU or interrupts.

All tested bit widths from 5- to 8-bits, 1 or 2 stop bits, and even/odd/none parity are supported.


ladyada:
fyi @tannewt on uart pio

earlephilhower:
FYI, an IRQ based transfer (or maybe DMA, even) like in #393 will be used here before I do the merge.  Right now you only get 4 byte-times of data before you start losing stuff.

PR #393, opened 2021-12-26
Use interrupts to capture Serial1/2 data, not polling

earlephilhower:
PR #379 was an ugly hack which works only if you poll the Serial port more
frequently than ~8 byte times.

This PR replaces the polling with an IRQ based lockless writer/reader queue
so that even if you only read every 32 byte times, the Serial1/2 port should
not lose data.  It also should use less CPU and allow for somewhat higher
throughput.

### DIFF
--- a/cores/rp2040/SerialUART.cpp
+++ b/cores/rp2040/SerialUART.cpp
@@ -81,19 +81,37 @@ void SerialUART::begin(unsigned long baud, uint16_t config) {
     int bits, stop;
     uart_parity_t parity;
     switch (config & SERIAL_PARITY_MASK) {
-    case SERIAL_PARITY_EVEN: parity = UART_PARITY_EVEN; break;
-    case SERIAL_PARITY_ODD: parity = UART_PARITY_ODD; break;
-    default: parity = UART_PARITY_NONE; break;
+    case SERIAL_PARITY_EVEN:
+        parity = UART_PARITY_EVEN;
+        break;
+    case SERIAL_PARITY_ODD:
+        parity = UART_PARITY_ODD;
+        break;
+    default:
+        parity = UART_PARITY_NONE;
+        break;
     }
     switch (config & SERIAL_STOP_BIT_MASK) {
-    case SERIAL_STOP_BIT_1: stop = 1; break;
-    default: stop = 2; break;
+    case SERIAL_STOP_BIT_1:
+        stop = 1;
+        break;
+    default:
+        stop = 2;
+        break;
     }
     switch (config & SERIAL_DATA_MASK) {
-    case SERIAL_DATA_5: bits = 5; break;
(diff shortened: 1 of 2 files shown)


earlephilhower:
@currentc57, if you have a chance could you also give this PR a try?  It supersedes the earlier one and does a real IRQ-based FIFO so it should be much more stable.  Don't mind the failing CI...it seems GH is having issues with `git clone` requests and failing randomly the past day or so.

currentc57:
> @currentc57, if you have a chance could you also give this PR a try? It supersedes the earlier one and does a real IRQ-based FIFO so it should be much more stable. Don't mind the failing CI...it seems GH is having issues with `git clone` requests and failing randomly the past day or so.

I gave it a quick test last night.  Didn't see any problems, but it was far from an exhaustive test.  Won't have much time this week.  Busy catching up with work after the holiday.

earlephilhower:
Thanks again for the additional sanity check.  If there's a problem, I'm sure folks will let me know about it.

currentc57:
Okay, I found one part of my program that's not working with the new code.  Some variables I'm updating over Serial2 are getting clobbered.  Don't know why yet.  It might take me a while.  I'm using both cores, multiple interrupts, Serial and Serial2.  A lot of stuff to test.

earlephilhower:
Argh, please do keep us updated in your own time.  It's entirely possible my 10-minute lockless circular queue code has some interaction that I've missed.  It seemed elegant and simple, but I'm an EE not a CS...

What core are you calling `Serial2.begin()` from, and which are you doing the `Serial2.read()`s on?  The way the interrupts work on the RP2040 the IRQ handler will be called on the core they were installed from.

currentc57:
I've think I've narrowed the problem down to this section of the library I'm using.
`  while(_serial->available() < 8){                  // Waiting for bytes to come to Serial, 
                                                    // we are waiting for 8 bytes
    if((millis() - _tmr1) > 400UL){                // Waiting... But not forever...after the timeout 
       _numberValue = 777777;
       break;                                  // Exit the loop due to timeout. Return 777777
    }
  }`
It's timing out even if I change `while(_serial->available() < 8)` to 'while(_serial->available() < 1)'
This only happens with the IRQ version.  Your previous version works without any problems.

currentc57:
Looks like the timer interrupt I was using was causing the new IRQ Serial to fail.  I changed my IRQ usage and your new version seems to be working well.

SirSydom:
it seems that this feature somehow affects timing or data integrity.

I use the [thelsing/knx stack](https://github.com/thelsing/knx) and I just found out that in some special cases communication does not work anymore (1.10.0). So I rolled back and found out that 1.9.9 worked, but 1.9.10 does not work.
No Core1 is involved. 

Problem is, that i'm just a user of this stack and my knowledge about whats going on there is limited. So tracking this down will be hard...

PR #397, opened 2021-12-30
Fix some pin definitions for Adafruit QT Py RP2040

Kealper:
This PR corrects the pin definitions for pins A0, A1, A2, and A3 on the Adafruit QT Py RP2040.

On this board, the four pins above are reversed compared to what's defined in the common pin configuration. The current way this PR is fixing this is to just use preprocessor `#define` in the board-specific `variants/adafruit_qtpy/pins_arduino.h` to override the `static const` declarations from `variants/generic/common.h` of those four pins. This method appears to work fine and compiles without warnings while keeping the general board variant style alive.

Another option for correcting this problem would be to copy all the pin definitions from `variants/generic/common.h` into the board's specific file with the required changes and just not include that `common.h` in it, but that way doesn't fit the style of how other boards are defined, so is this method in the PR okay with you, or would you prefer it be done more like this other method?

### DIFF
--- a/variants/adafruit_qtpy/pins_arduino.h
+++ b/variants/adafruit_qtpy/pins_arduino.h
@@ -39,3 +39,10 @@
 #define WIRE_HOWMANY   (2u)
 
 #include "../generic/common.h"
+
+// Pin overrides specific to the QT Py RP2040
+// Note, these must go below the common.h include to work correctly
+#define A0 (29u)
+#define A1 (28u)
+#define A2 (27u)
+#define A3 (26u)
--- a/variants/adafruit_qtpy/pins_arduino.h
+++ b/variants/adafruit_qtpy/pins_arduino.h
@@ -38,11 +38,10 @@
 #define SPI_HOWMANY    (1u)
 #define WIRE_HOWMANY   (2u)
 
-#include "../generic/common.h"
-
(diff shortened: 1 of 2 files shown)


earlephilhower (changes requested):
Thanks for the fix!

You bring up a good question for which I really don't have a good answer.

However, I'm thinking the best way to do this is to modify common to allow overrides w/a "well known" `#define __PIN_A0 XXX`

So common.h would be something like:
````
#ifdef __PIN_A0
const int A0 = __PIN_A0;
#else
const int A0 = (26u);
....
````

or
````
#ifndef __PIN_A0
#define __PIN_A0 (26u)
#endif
....
const int A0 = __PIN_A0;
...
````

And your patch would be
````
#define __PIN_A0 (29u)
....
#include "../generic/common.h"
````

Does that seem reasonable?

earlephilhower (approved):
Thanks for the update!  I think this way will be better in the long run.

Kealper:
I think that way of modifying common to create a bit of a "standardized" method of overriding A0-A3 is the best way to go about this as well, since it provides a way for future variants that might need them changed to go about that this way instead of everyone coming up with different solutions down the road.

dhalbert:
Tagging @ladyada for interest.

ladyada:
awesome!

PR #407, opened 2022-01-03
Add the RP2040 Stamp board

arturo182:
Thanks so much for this port!

More info on the Stamp available here: https://www.solder.party/docs/rp2040-stamp/


tools/makeboards.py
```diff
@@ -173,5 +173,6 @@ def MakeBoard(name, vendor_name, product_name, vid, pid, pwr, boarddefine, flash
 MakeBoard("challenger_nb_2040_wifi", "iLabs", "Challenger NB 2040 WiFi", "0x2e8a", "0x100b", 500, "CHALLENGER_NB_2040_WIFI_RP2040", 8, "boot2_w25q080_2_padded_checksum")
 MakeBoard("ilabs_rpico32", "iLabs", "RPICO32", "0x2e8a", "0x1010", 250, "ILABS_2040_RPICO32_RP2040", 8, "boot2_w25q080_2_padded_checksum")
 MakeBoard("melopero_shake_rp2040", "Melopero", "Shake RP2040", "0x2e8a", "0x1005", 250, "MELOPERO_SHAKE_RP2040", 16, "boot2_w25q080_2_padded_checksum")
+MakeBoard("solderparty_rp2040_stamp", "Solder Party", "RP2040 Stamp", "0x1209", "0xa182", 500, "SOLDERPARTY_RP2040_STAMP", 8, "boot2_generic_03h_4_padded_checksum")
```
earlephilhower, line 176: Have you tried other boot2 options?  I think the one you've selected should work fine on anything, but it's the slowest one (/4 clock, and a 1 bit SPI interface IIRC).  Not a blocker, but if you do find another one that works, we can update then.

arturo182: The chip used is the GD25Q64C, I didn't find a definition for that one so I went with the default. But I will check each of them and if any non-generic one works, I'll send another PR. Thanks for the quick review!

PR #424, opened 2022-01-20
Update to latest upstream Keyboard/Mouse

earlephilhower:
Keyboard class now has support for non-en_US layouts, and documentation has
been added to the Mouse class.


earlephilhower:
Thanks to @rico0260 for letting me know about the update, and for the initial PR!

PR #428, opened 2022-01-22
Add interpolater include path

### LINKED ISSUE #427: Include SDK interpolator support
I'm currently using  arduino-pico 1.9.14 on Arduino 1.8.19.

While trying to use the RP2040 interpolator with SDK functions, I noticed that the hardware interpolator directory was not used for the search path. That is, doing the recommended
`#include "hardware/interp.h"`
didn't work. After much guessing-and-checking (the directory structure of the SDK is a mystery to me) I found that the issue was solved by adding the line
`-iwithprefixbefore/pico-sdk/src/rp2_common/hardware_interp/include`
to lib/platform_inc.txt. I have done some quick tests, and the hardware seems to be active and behave as it should.

I suppose this should be a pull request, but I'm relatively inexperienced with git and wouldn't trust myself to do it right.

Thanks!

earlephilhower:
Fixes #427.  The libpico.a build already includes the HW interpolator
sources, but the include was missing from the Arduino path.

### DIFF
--- a/lib/platform_inc.txt
+++ b/lib/platform_inc.txt
@@ -28,6 +28,7 @@
 -iwithprefixbefore/pico-sdk/src/rp2_common/hardware_flash/include
 -iwithprefixbefore/pico-sdk/src/rp2_common/hardware_gpio/include
 -iwithprefixbefore/pico-sdk/src/rp2_common/hardware_i2c/include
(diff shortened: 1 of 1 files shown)


earlephilhower:
Thanks @omion

Issue #436, opened 2022-01-25
Tone library glitch


jhmaloney:
I'm seeing some strange behavior in the tone() function when I attempt to change the output frequency on a pin. Here's an example:

```
tone(10, 200);
delay(250);
tone(10, 250);
```

Expected behavior: the frequency starts at 200 Hz then changes to 250 Hz after 250 milliseconds. After that, the output should be a nice, symmetric 250 Hz square wave.

Actual behavior observed with an oscilloscope:

<img width="572" alt="Screen Shot 2022-01-25 at 12 14 06 PM" src="https://user-images.githubusercontent.com/4073169/151026089-91d30d0b-5a30-49d6-a4b5-0104533d4d4f.png">

Note the narrow pulses. It behaves as if there are two instances of the state machine outputting to the pin at different frequencies.

After doing some debugging, the issue appears to be that the find() operation is failing to find the existing tone entry for that pin:

```
    auto entry = _toneMap.find(pin);
    Tone *newTone;
    if (entry == _toneMap.end()) { // this is always true, even when there was already an entry for the given pin
```

As a result, it creates a new tone instance and associated state machine.

The code looks correct, but I don't think it is behaving as intended. Unfortunately, I'm not enough of C++ expert to know why find() isn't returning the existing entry.

Andy2No:
> ```
> tone(10, 200);
> delay(250);
> tone(10, 250);
> ```

It might be best if you post the entire sketch, assuming it's fairly short.

Is there a delay after the second tone()?  If not, it immediately does the first one again, creating a short glitch, before switching back to 200Hz.

jhmaloney:
Those lines are in setup(). loop() is empty.

earlephilhower:
@jhmaloney I am unable to reproduce anything like what you're showing.  Can you post your exact `setup`?
````
void setup() {
  tone(10, 200);
  delay(250);
  tone(10, 250);
}
void loop() { }
````
Gives me this output on my logic analyzer:

Right at startup after the 1st tone but before the 2nd:
![image](https://user-images.githubusercontent.com/11875/151052484-1f189d0a-95da-464b-8d38-65da7df8b0e9.png)

And at the transition:
![image](https://user-images.githubusercontent.com/11875/151052844-7fe4a840-62ec-4d6f-a1b7-7e0bbcb20dd5.png)

And a few seconds later:
[...]

PR #440, opened 2022-01-26
Fix libraries and RAM size in PlatformIO build

maxgerhardt:
This fixes two things:
* the `ram_size` variable which is fixed at 256K. Previously PlatformIO's board definitions would have 264K there (+4K scratch x/y), which would generate a linker script with the content
```
MEMORY
{
    FLASH(rx) : ORIGIN = 0x10000000, LENGTH = 1044480
    RAM(rwx) : ORIGIN =  0x20000000, LENGTH = 264k
    SCRATCH_X(rwx) : ORIGIN = 0x20040000, LENGTH = 4k
    SCRATCH_Y(rwx) : ORIGIN = 0x20041000, LENGTH = 4k
}
```
i.e., the `RAM` section length would be off by 8K. 
* links **only** against `lib/libpico.a`, not `lib/libstdc++.a` by reworking the linkerflags (no `-L` flag to add to the search directory and `-lpico -lstdc++`, only link `libpico.a` by full path and source `libstdc++.a` from compiler). The `libstdc++.a` has weird errors which make the the system crash & burn when attempting to use even the smallest bits of the C++ standard library, such as creating a `std::string` or a `std::map`. I am not sure why this file exists here, the Arduino IDE build also does not link against it.

Related to https://github.com/maxgerhardt/pio-pico-core-earlephilhower-test/issues/3 and #436.

### DIFF
--- a/tools/platformio-build.py
+++ b/tools/platformio-build.py
@@ -87,7 +87,7 @@ env.Append(
     LIBSOURCE_DIRS=[os.path.join(FRAMEWORK_DIR, "libraries")],
 
     # do **NOT** Add lib to LIBPATH, otherwise 
-    # erronous libstdc++.a will be found that crashes! 
+    # erroneous libstdc++.a will be found that crashes! 
     #LIBPATH=[
     #    os.path.join(FRAMEWORK_DIR, "lib")
     #],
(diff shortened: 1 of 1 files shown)


earlephilhower:
I'm very surprised the binaries didn't just crash immediately with a different stdc++.a interface/structures vs. the interface used in the headers used to compile the code.  Good debug!

PR #444, opened 2022-01-28
 Correctly use earlephilhower.varant if variant not given, ensure USB…

maxgerhardt:
… power macro is always there

Cherry-pick of https://github.com/maxgerhardt/arduino-pico/commit/beec911fce6151cee3f1306421d21af5528f35c7

Fixes a crash of the builder script, `board.get("build.variant", None)` will throw an exception if the `build.variant` was not found (and not return `None` as the fallback value which was the wanted behavior), but an empty string works, so check against that.

USB power defines moved directly into the board files, but still ensure that the macro always exists (with a default fallback value) to not fail the build.

### DIFF
--- a/tools/platformio-build.py
+++ b/tools/platformio-build.py
@@ -148,9 +148,12 @@ def configure_usb_flags(cpp_defines):
         ("USB_PID", usb_pid),
         ("USB_MANUFACTURER", '\\"%s\\"' % usb_manufacturer),
         ("USB_PRODUCT", '\\"%s\\"' % usb_product),
-        ("SERIALUSB_PID", usb_pid),
-        ("USBD_MAX_POWER_MA", 250)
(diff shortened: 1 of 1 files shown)


earlephilhower:
I believe there's still a pico-quick-toolchain bug or PR left to fully close this stuff out, yes?

maxgerhardt:
>I believe there's still a pico-quick-toolchain bug or PR left to fully close this stuff out, yes?

You mean in regards to developing the PlatformIO further so that it is ready to merge into mainline PlatformIO / platform-raspberrypi? Yes, the whole packaging for toolchain-pico, tool-openocd-raspberry and framework-arduinopico is yet to be done. Either manually in the first round to get them in now or automated. The released OpenOCD package still has some issues on Windows (https://github.com/earlephilhower/pico-quick-toolchain/issues)

PR #467, opened 2022-02-07
Add retries for serial port init. Change order of tests to favor /run…

gsexton:
This PR implements 3 changes:

1) Favor /run/media over /media. My OpenSUSE box has both paths but udiskctl uses /run/media. This is what actually kept it form working on OpenSUSE 15.3.
2) Add retries for serial port init. This seems to make things a little more robust.
3) The udiskctl was mounting the volume and then trying to find it. I was having trouble here and saw that it prints the mount point, so added a regex to find the path. If the regex doesn't match,it falls back to the old behavior of searching for the mount point.

### DIFF
--- a/tools/uf2conv.py
+++ b/tools/uf2conv.py
@@ -317,23 +317,17 @@ def main():
 
     if args.serial:
         if str(args.serial).startswith("/dev/tty") or str(args.serial).startswith("COM") or str(args.serial).startswith("/dev/cu"):
-            print("Entering serial port reset")
-            for tries in range(5):
+            try:
+                print("Resetting " + str(args.serial))
                 try:
-                    print("Resetting " + str(args.serial) + " try " + str(tries))
-                    try:
-                        ser = serial.Serial(args.serial, 1200)
-                        ser.dtr = False
-                        print("Reset succeeded. Sleeping 10s")
-                        time.sleep(10)
-                        break
-                    except:
-                        print("Caught exception during reset!")
-                    # Probably should be smart and check for device appearance or something
-                    print("Sleeping 30 seconds")
-                    time.sleep(30)
+                    ser = serial.Serial(args.serial, 1200)
+                    ser.dtr = False
                 except:
-                    pass
+                    print("Caught exception during reset!")
+                # Probably should be smart and check for device appearance or something
+                time.sleep(10)
+            except:
+                pass
     if args.list:
         list_drives()
     else:
--- a/tools/uf2conv.py
+++ b/tools/uf2conv.py
@@ -236,9 +236,9 @@ def get_drives():
             if len(words) >= 3 and words[1] == "2" and words[2] == "FAT":
                 drives.append(words[0])
     else:
-        rootpath = "/media"
+        rootpath = "/run/media"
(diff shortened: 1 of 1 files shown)


tools/uf2conv.py
```diff
@@ -317,18 +317,23 @@ def error(msg):
...
     if args.serial:
         if str(args.serial).startswith("/dev/tty") or str(args.serial).startswith("COM") or str(args.serial).startswith("/dev/cu"):
-            try:
-                print("Resetting "+str(args.serial))
+            print("Entering serial port reset")
+            for tries in range(5):
                 try:
-                    ser = serial.Serial(args.serial, 1200)
-                    ser.dtr = False
+                    print("Resetting " + str(args.serial) + " try " + str(tries))
```
earlephilhower, line 323: This makes it take ~2 minutes to upload whenever you use the BOOTSEL to make it upload.  I'm not sure what retrying the serial (which should be steady available anyway for a long time before the upload process) is really buying you, but that;s just too painful.

tools/uf2conv.py
```diff
@@ -236,9 +236,9 @@ def get_drives():
             if len(words) >= 3 and words[1] == "2" and words[2] == "FAT":
                 drives.append(words[0])
     else:
-        rootpath = "/media"
+        rootpath = "/run/media"
```
earlephilhower, line 239: I'll need to check this doesn't kill Ubuntu or CentOS-likes...

tools/uf2conv.py
```diff
@@ -359,12 +364,31 @@ def error(msg):
...
+                rpidisk = glob.glob(globexpr)
+                if len(rpidisk) == 0:
+                    print("Unable to find disk by ID using expression: {}".format(globexpr))
+                else:
+                    try:
+                        cmd = ["udisksctl", "mount", "--block-device", os.path.realpath(rpidisk[0])]
+                        proc_out = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                        if proc_out.returncode == 0:
+                            stdoutput = proc_out.stdout.decode("UTF-8")
+                            match = re.search(r'Mounted\s+.*\s+at\s+([^\.]*)', stdoutput)
```
earlephilhower, line 377: Is this legit on internationalized installs?  I've been bitten before looking for human-readable strings that come from `.po`s...

earlephilhower (changes requested):
Thanks for the PR, but I have a few concerns noted above.  I'll try the `/media/run` change locally and close the discussion if it's all good...

tools/uf2conv.py
earlephilhower: Testing looks OK to me

tools/uf2conv.py
gsexton: OK. I'll revert that out.

tools/uf2conv.py
gsexton: I've reviewed the source for udisckctl and captions aren't localized. Even if the caption were localized, the code would fall back to the previous behavior of using get_drives() to find the new mount point.

earlephilhower (approved):
LGTM now, thanks!

PR #473, opened 2022-02-08
Add SerialUART::setPollingMode()

### LINKED ISSUE #472: Add polled-mode UART Serial
As discussed with @SirSydom in #468, it may make sense to have a non-FIFO polled only mode for the UART ports when using `setFIFOSize(0)`.

Re-implement the polled mode option in the UART driver when `_fifoSize==0`

earlephilhower:
Fixes #472

Instead of using interrupts, explicitly call the IRQ handler dueing Serial
read/peek/available calls.

### DIFF
--- a/cores/rp2040/SerialPIO.cpp
+++ b/cores/rp2040/SerialPIO.cpp
@@ -122,7 +122,7 @@ void __not_in_flash_func(SerialPIO::_handleIRQ)() {
             if (next_writer == _fifosize) {
                 next_writer = 0;
             }
-            asm volatile("" ::: "memory"); // Ensure the reader value is only written once, correctly
+            asm volatile("" ::: "memory"); // Ensure the writer value is only written once, correctly
             _writer = next_writer;
         } else {
             // TODO: Overflow
--- a/cores/rp2040/SerialPIO.cpp
+++ b/cores/rp2040/SerialPIO.cpp
@@ -114,15 +114,13 @@ void __not_in_flash_func(SerialPIO::_handleIRQ)() {
             }
         }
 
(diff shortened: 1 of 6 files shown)


earlephilhower:
@SirSydom would appreciate you giving this a whirl and reporting back, please.

earlephilhower:
Loopback testing with a simple script shows polled mode working properly and immediately returning bytes as receives, while IRQ-based does it in chunks of 4 bytes until the final set.

SirSydom:
This was at the right time :)
I told you, that #471 fixed my problem with the knx stack, well, that was true for an old version of this stack.
With the current master (of the knx branch) it is screwed again, BUT, when I activate polling it works. Great job.

One question:

do you always get 4byte chunks? also when communication is paused? that would mean somehow, the "32bit times IRQ" does not work properly.

earlephilhower:
No, the last bit will be at 32-bit times out.  I connect up pins GP0 and GP1 and run the following script:
````
void setup() {
  Serial.begin(115200);
  Serial1.setPollingMode(false);
  Serial1.begin(110);
  delay(5000);
}

void loop() {
  Serial1.printf("hello\n");
  while (Serial1.available() < 6) {
    Serial.printf("Avail: %d\n", Serial1.available());
  }
  Serial.printf("read: ");
  while (Serial1.available()) {
    Serial.printf("%c", Serial1.read());
  }
  Serial.printf("\n");
  delay(1000);
}
````

For polled you see Avail go 0..1..2..3..4..5..print
For IRQ you see Avail go 0...4...print

IF the 32-bit timeout did not work, then it would never advance because Serial1 only sends out 6 bytes per loop.

PR #493, opened 2022-02-18
Added ability to reset and relaunch core 1

magnusnordlander:
Since the Pico SDK multicore API supports restarting core 1, by first resetting it, then launching it anew, it would be nice to expose that functionality through the RP2040 class.


earlephilhower (approved):
Thanks for the PR!  I'm curious, do you have a particular use case for this?

earlephilhower (changes requested):
Well, it does need to compile first. :)

Also, I believe we should reset the _MFIFO between the cores if we reset core1 as well.

magnusnordlander:
Right. That was a silly mistake. I think the build should pass now. 😅

As for the use case, I'm building a [controller for an espresso machine](https://github.com/magnusnordlander/smart-lcc). Core 0 is running anything safety critical (or well, there are safety valves and thermostats too), e.g. the steam boiler. Anything not safety critical runs in core 1. I have a watchdog that will reboot the system if core 0 stops working, and right now it does if core 1 stops working too, but another option that I much prefer is for core 0 to act as a watchdog for core 1, by restarting core 1 if it stops responding to messages.

Regarding resetting the _MFIFO, I assume you mean just purging both the queues?

earlephilhower:
> As for the use case, I'm building a [controller for an espresso machine](https://github.com/magnusnordlander/smart-lcc). Core 0 is running anything safety critical (or well, there are safety valves and thermostats too), e.g. the steam boiler. Anything not safety critical runs in core 1. I have a watchdog that will reboot the system if core 0 stops working, and right now it does if core 1 stops working too, but another option that I much prefer is for core 0 to act as a watchdog for core 1, by restarting core 1 if it stops responding to messages.

Very cool.  There is a real [hardware watchdog](https://raspberrypi.github.io/pico-sdk-doxygen/group__hardware__watchdog.html) you can use, check the docs.  I may not have included the libs in the `libpico.a` but it's a trivial addition if needed.  

> Regarding resetting the _MFIFO, I assume you mean just purging both the queues?

Yes, that's correct.  Just popping both until you can't pop anymore should be good (between the core1-reset and core1-start just in case core1 is doing something w/the queue at that time).

magnusnordlander:
> Very cool. There is a real [hardware watchdog](https://raspberrypi.github.io/pico-sdk-doxygen/group__hardware__watchdog.html) you can use, check the docs. I may not have included the libs in the `libpico.a` but it's a trivial addition if needed.

Absolutely, that's what I'm using to watch over Core 0.

> Yes, that's correct. Just popping both until you can't pop anymore should be good (between the core1-reset and core1-start just in case core1 is doing something w/the queue at that time).

Great, I've updated the branch to do that.

earlephilhower (approved):
Thanks and good luck with the project!

PR #496, opened 2022-02-20
PDM library: Update PIO and interrupt use

Bodmer:
The PIO and state machine were hard wired as PIO 0 and SM 0, so this caused problems if they were not free. The approach used by the Servo library has been adopted to search for a free PIO and SM.

The DMA_IRQ_0 was grabbed exclusively, but this conflicts with the use with SPI DMA. The interrupt is now shared, but has been allocated the highest possible priority.

### DIFF
--- a/libraries/PDM/src/PDM.h
+++ b/libraries/PDM/src/PDM.h
@@ -58,6 +58,12 @@ private:
   int _gain;
   int _init;
 
+// Hardware peripherals used
+  uint _dmaChannel;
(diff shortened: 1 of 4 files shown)


Bodmer:
Since the PDM PIO use is receive only, the PIO state machine RX FIFO's can be joined to reduce DMA interrupt load.

earlephilhower (approved):
Great improvement, thanks!

PR #499, opened 2022-02-21
Updating LoRa board pins and UART hw flow.

PontusO:
Added a few helper macros for the pins and missing Serial2 pins.
Also added support for UART hw_flow.

New methods setRTS and setCTS with the appropriate pin must be called before calling the begin of the selected UART.


cores/rp2040/SerialUART.cpp
```diff
@@ -65,6 +65,39 @@ bool SerialUART::setTX(pin_size_t pin) {
...
+    constexpr uint32_t valid[2] = { __bitset({3, 15, 19}) /* UART0 */,
+                                    __bitset({7, 11, 23, 27})  /* UART1 */
+                                  };
+    if ((!_running) && ((1 << pin) & valid[uart_get_index(_uart)])) {
+        _rts = pin;
+        return true;
+    }
+
+    if (_running) {
+        panic("FATAL: Attempting to set Serial%d.TX while running", uart_get_index(_uart) + 1);
```
earlephilhower, line 78: Minor thing, but the error message needs to be updated here and in the other fcn. ("Attemption to set ...serRTX/CTS")

cores/rp2040/SerialUART.h
```diff
@@ -28,13 +28,16 @@
...
 
+#define UART_PIN_NOT_DEFINED      (255u)
 class SerialUART : public HardwareSerial {
 public:
     SerialUART(uart_inst_t *uart, pin_size_t tx, pin_size_t rx);
 
     // Select the pinout.  Call before .begin()
     bool setRX(pin_size_t pin);
     bool setTX(pin_size_t pin);
+    bool setRTS(pin_size_t pin);
```
earlephilhower, line 39: Could you please add these 2 calls to `docs/pins.rst`?

earlephilhower (changes requested):
Nice addition!  Minor requests, please.

earlephilhower (approved):
Thanks!

Issue #515, opened 2022-03-16
uf2conv.py - re.search() should not match newline


randomllama:
`arduino-cli upload -b rp2040:rp2040:adafruit_feather -p /dev/pico`
```
Converting to uf2, output size: 131584, start address: 0x2000
Traceback (most recent call last):
  File "/home/user/.arduino15/packages/rp2040/hardware/rp2040/1.13.0/tools/uf2conv.py", line 410, in <module>
    main()
  File "/home/user/.arduino15/packages/rp2040/hardware/rp2040/1.13.0/tools/uf2conv.py", line 405, in main
    print("Flashing %s (%s)" % (d, board_id(d)))
  File "/home/user/.arduino15/packages/rp2040/hardware/rp2040/1.13.0/tools/uf2conv.py", line 276, in board_id
    with open(path + INFO_FILE, mode='r') as file:
FileNotFoundError: [Errno 2] No such file or directory: '/run/media/user/RPI-RP2\n/INFO_UF2.TXT'
Error during Upload: Failed uploading: uploading error: exit status 1
```

https://github.com/earlephilhower/arduino-pico/blob/53043830e7747fabf4f7cc2876e3bfad4d9f2fe2/tools/uf2conv.py#L383

earlephilhower:
Can I ask what OS and version you're using?  I'm on Ubuntu 20.04 and not seen this, but you're the second report of spurious newlines, so I do believe there's a problem in the upstream source...

randomllama:
I use Arch, btw.  arduino-pico is v1.13.0.

Changing the mentioned line to:
` match = re.search(r'Mounted\s+.*\s+at\s+([^\r\n]*)', stdoutput) `
makes it work as expected for me.

earlephilhower:
Great, thanks for the info.  I checked and it's actually not upstream code but changes submitted by users here that's calling the utility and grabbing those newlines, so mybad.  Upstream's got issues, but this isn't one of them. :)

Would you like to do a 1-line PR since you not only found the problem, you found a fix?

PR #516, opened 2022-03-16
udisks v2.9.0 dropped the trailing period for "udisksctl mount".

randomllama:
udisks >= 2.9.0 dropped the period at the end.  Adjust regex to look a period or newline.

Fixes: #515
Upstream reference: https://github.com/storaged-project/udisks/commit/6842123525966ad5fd5dfa5eccef43c3ea985c21


earlephilhower (approved):
Thanks for the debug and the fix!

PR #525, opened 2022-03-30
Add RTS/CTS to Arduino Nano Connect Serial2(NINA)

### LINKED ISSUE #524: Nano RP2040 Connect and Bluetooth
My goal was to use this with the ArduinoBLE library. I started with a git pull (today) of this (linux host), blink worked, then I added only the ArduinoBLE library.

Observed behavior is that BLE.begin() returns false. Scoping, nina never responds to commands over the uart (Serial2). There are test points on the bottom of the board exposing tx/rx between 2040 and nina. In both mbed and rp2040 builds, I see the command from 2040 to nina, but only in the mbed build do I see a response. In the rp2040 build, there is no response.

From the schematic, RTS and CTS appear to be connected between the RP2040 and Nina (but they are not on test points). In the mbed installation (windows install, latest) I see:

arduino\hardware\mbed_nano\2.7.2\variants\NANO_RP2040_CONNECT\pins_arduino.h
#define SERIAL2_TX			(digitalPinToPinName(25))
#define SERIAL2_RX			(digitalPinToPinName(26))
#define SERIAL2_CTS			(digitalPinToPinName(27))
#define SERIAL2_RTS			(digitalPinToPinName(28))

arduino\hardware\mbed_nano\2.7.2\cores\arduino\Serial.cpp
#ifdef SERIAL2_RTS
UART _UART2_(SERIAL2_TX, SERIAL2_RX, SERIAL2_RTS, SERIAL2_CTS);
#else
UART _UART2_(SERIAL2_TX, SERIAL2_RX, NC, NC);
#endif

In the rp2040 installation, RTS & CTS are not defined for Serial2 (as far as I can tell), but there are setRTS and setCTS methods on the UART. If I add:

Serial2.setCTS(D27);
Serial2.setRTS(D28);

to my sketch before BLE.begin(), it works in the rp2040 installation, so I think nina must be expecting hw flow control.

With the setCTS and setRTS calls, BLE comes up as far as letting me run a close derivative of the ArduinoBLE "LED" example (I can toggle the nano's led from my phone).

A most-basic sketch to show they make a difference can have just a call to BLE.begin() in setup(); it returns true if RTS and CTS are added (and BLE goes on to work), and returns false if not (and BLE does not work).

I'd create a PR, but I'd almost certainly fix it the wrong way in the wrong place.

I also see other (github) issues implying that others are using wifi on this board, which I would have guessed wouldn't work without RTS/CTS, so I might be missing a different way to make it work.

earlephilhower:
Fixes #524


earlephilhower:
@jdfin can you undo your app chances and check this PR out and report back?  I don't have a Nano Connect to do any testing on, so this is just using your comments and their schematic.

cores/rp2040/SerialUART.cpp
```diff
@@ -339,8 +339,17 @@ SerialUART::operator bool() {
     return _running;
 }
 
+#if defined(PIN_SERIAL1_RTS)
+SerialUART Serial1(uart0, PIN_SERIAL1_TX, PIN_SERIAL1_RX, PIN_SERIAL1_RTS, PIN_SERIAL1_CTS);
+#else
 SerialUART Serial1(uart0, PIN_SERIAL1_TX, PIN_SERIAL1_RX);
+#endif
+
+#if defined(PIN_SERIAL1_RTS)
```
jdfin, line 348: PIN_SERIAL2_RTS

jdfin: PIN_SERIAL2_RTS

earlephilhower: D'oh!  Fixed.

jdfin:
That does it! (with the one #if typo fix)

I also ran one of the examples with the WifiNina library to make sure that worked, and it did. I did not try it without this change.

Issue #526, opened 2022-04-03
Pad drive strength


jlbirccyn:
Hello,

Thanks for this great work!

I looked at the code a bit and did some testing and the drive strength is the default value at reset (4mA), which is pretty low. Would it be possible to set it to 12mA in pinMode? Otherwise we can call the function of the sdk(gpio_set_drive_strength) but it is not the usual simplicity of the Arduino.

Kind regards.

earlephilhower:
I'm loathe to change the default value simply because the core is being used by folks for embedded, battery-powered apps.  Changing to a higher drive strength could change the behavior of their boards and reduce battery life.

Any ideas as to a way of allowing folks to override the default?

Gavin-Perry:
I'm happy with gpio_set_drive_strength() 
It clearly says what it does and does the job.

jlbirccyn:
I understand

Maybe add 4 constants to use with pinMode : OUTPUT_2, OUTPUT_4, OUTPUT_8 and OUTPUT_12? OUTPUT would default to OUTPUT4 for compatibility.

earlephilhower:
Interesting idea, especially since it seems the RP2040 can individually set the drive strengths on a per-pad basis.  Would you like to so a Pull Request?  This should be a pretty simple addition...

jlbirccyn:
Of course I do. I'll handle it.

earlephilhower:
Fixed by #532

PR #532, opened 2022-04-08
Added the ability to set pad current via pinMode: OUTPUT_2MA, OUTPUT_…

jlbirccyn:
Hello,

Here is the pull request corresponding to the issue #526 

Best regards


earlephilhower (changes requested):
Nice PR and thanks for documenting, too.

Could you add the new OUTPUT_xMA keywords to https://github.com/earlephilhower/arduino-pico/blob/master/keywords.txt  so the IDE will hilight them?  (be sure to use TAB not spaces in that file)

Can you add a minor change to the CI system so this PR will pass?

Remove this test:
https://github.com/earlephilhower/arduino-pico/blob/061b38f0b85af7e985f7b6149bc4c199b7b0967c/.github/workflows/pull-request.yml#L49-L52

I can see no way of making this change w/o modifying the API files (i.e. you can't extend an enum in C/C++) or using some very ugly `#define` shenanigans.  So, out goes that API check and it'll be on me to verify any future PRs don't hack them too horribly...

jlbirccyn:
Hello Earle,

Just pushed the keywords.txt with OUTPUT_xMA and updated pull-request.yml

I found a trick here : https://www.codeproject.com/Articles/16150/Inheriting-a-C-enum-type to inherit an enum. Did not try it however.

Best regards.

earlephilhower (approved):
Thanks for the PR!

PR #533, opened 2022-04-09
Add FreeRTOS support thanks to @hfellner

earlephilhower:
Using all the work from @hfellner and others.  Things work but are not
stable (i.e. flash operations will make things go badly, fast; delay()
does not actually release the core for other work, etc.).


hfellner:
Making delay() release the core for other work would be simple by making the default implementation a weak symbol and providing an override in the RP2040_FreeRTOS library that calls vTaskDelay(delay / portTICK_PERIOD_MS). However, this cannot be done for delayMicroseconds(), because the minimum yield period for FreeRTOS on the RP2040 is 1ms.

Issue #564, opened 2022-04-29
Setting configSUPPORT_STATIC_ALLOCATION to 1 results in a compile error


ogatatsu:
Setting configSUPPORT_STATIC_ALLOCATION in FreeRTOSConfig.h to 1 will result in a compile error.

```
C:\Users\ogata\Documents\ArduinoData\packages\rp2040\hardware\rp2040\2.0.0\libraries\FreeRTOS\src\variantHooks.cpp:317:6: error: conflicting declaration of C function 'void vApplicationGetIdleTaskMemory(StaticTask_t**, StackType_t**, uint16_t*)'
  317 | void vApplicationGetIdleTaskMemory( StaticTask_t ** ppxIdleTaskTCBBuffer,
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from C:\Users\ogata\Documents\ArduinoData\packages\rp2040\hardware\rp2040\2.0.0\libraries\FreeRTOS\src\task.h:1,
                 from C:\Users\ogata\Documents\ArduinoData\packages\rp2040\hardware\rp2040\2.0.0\libraries\FreeRTOS\src\variantHooks.cpp:41:
c:\users\ogata\documents\arduinodata\packages\rp2040\hardware\rp2040\2.0.0\libraries\freertos\lib\freertos-kernel\include\task.h:1861:10: note: previous declaration 'void vApplicationGetIdleTaskMemory(StaticTask_t**, StackType_t**, uint32_t*)'
 1861 |     void vApplicationGetIdleTaskMemory( StaticTask_t ** ppxIdleTaskTCBBuffer,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:\Users\ogata\Documents\ArduinoData\packages\rp2040\hardware\rp2040\2.0.0\libraries\FreeRTOS\src\variantHooks.cpp:336:6: error: conflicting declaration of C function 'void vApplicationGetTimerTaskMemory(StaticTask_t**, StackType_t**, uint16_t*)'
  336 | void vApplicationGetTimerTaskMemory( StaticTask_t ** ppxTimerTaskTCBBuffer,
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from C:\Users\ogata\Documents\ArduinoData\packages\rp2040\hardware\rp2040\2.0.0\libraries\FreeRTOS\src\timers.h:1,
[...]

PR #566, opened 2022-04-30
Fix compile error when configSUPPORT_STATIC_ALLOCATION is set to 1

ogatatsu:
#564


earlephilhower (changes requested):
Thanks for the PR, but I think the right way to fix this is to just set the value in the FreeRTOSConfig.h.  The macro you changed here is used inside `tasks.c` so fixing it here is only a partial solution.

So can we add a
````
#define configSTACK_DEPTH_TYPE uint32_t
````
to https://github.com/earlephilhower/arduino-pico/blob/master/libraries/FreeRTOS/src/FreeRTOSConfig.h , please?

earlephilhower (approved):
Thanks!

PR #586, opened 2022-05-19
Fix Wire I2C restart handling

### LINKED ISSUE #585: i2c slave .onReceive() failure/error when 'restart' rather than 'stop' is transmitted by i2c master
wire.cpp does not generate a "Wire.onReceive()" interrupt for an i2c Slave following transmission of a 'restart' rather than a 'stop' by an i2c Master. This situation can occur frequently when a Master sends a register address followed by a read request for the contents of that register.

This is particularly important in a Multi Master i2c system, where the Master should not allow another Master to obtain control of the bus between its write of the required register and the read of the register content [and therefore explicitly uses a 'restart' rather than a 'stop' follow by a 'start' between it's write and read over the bus].

I have written a short Arduino C++ program to illustrate the issue
i2cmemory.ino.txt

The error is that wire.cpp in arduino-pico/libraries/Wire/src only checks the IC_INTR_STAT register bit 10 (R_START_DET - a 'start' condition) to generate it's interrupt . . . it should also check and act on bit 12 (R_RESTART_DET - i.e. a 'restart' condition)

The attached image shows the required changes to wire.cpp to correct this wire cpp diff

earlephilhower:
Fixes #585

On an I2C bus restart, call the onReceive callback and clear the buffer.

Thanks to @DWiskow for the debug and patch!

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -149,7 +149,7 @@ void TwoWire::onIRQ() {
         _buffLen = 0;
         _buffOff = 0;
         _slaveStartDet = false;
-        _i2c->hw->clr_reset_det;
+        _i2c->hw->clr_restart_det;
     }
     if (_i2c->hw->intr_stat & (1 << 10)) {
         _buffLen = 0;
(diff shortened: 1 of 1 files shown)


DWiskow:
All good

PR #602, opened 2022-06-01
Add attachInterruptParam

### LINKED ISSUE #601: Missing attachInterruptParam definition
I was trying to us an interrupt service routine with a parameter, but I noticed that this core doesn't seem to implement `attachInterruptParam`.

A minimal setup to replicate this is:

```C++
#define INT_PIN 5

void isr(String* out) {
  Serial.println(*out);
}

void setup() {
  pinMode(INT_PIN, INPUT);
  String str = "hello";
  attachInterrupt(digitalPinToInterrupt(INT_PIN), isr, RISING, &str);
}

void loop() {
  
}
```

Compilations leads to this linker error:
```
packages\rp2040\hardware\rp2040\2.0.2\cores\rp2040/api/Interrupts.h:39: undefined reference to `attachInterruptParam'
```

Would it be possibile to implement this?

Marco

earlephilhower:
Fixed #601

Allows apps to provide a callback parameter in the `attachInterrupt` call
to be passed back to the function when the interrupt happens.

### DIFF
--- a/cores/rp2040/wiring_private.cpp
+++ b/cores/rp2040/wiring_private.cpp
@@ -45,25 +45,39 @@ extern "C" void noInterrupts() {
 // Only 1 GPIO IRQ callback for all pins, so we need to look at the pin it's for and
 // dispatch to the real callback manually
 auto_init_mutex(_irqMutex);
-static std::map<pin_size_t, voidFuncPtr> _map;
+class CBInfo {
+public:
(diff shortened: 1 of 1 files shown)


marc0777:
Perfect, thank you very much!

I've implemented this on my code and it compiles correctly. Will test it on the actual hardware as soon as possible.

PR #608, opened 2022-06-05
Fix WizFi360-EVB-Pico pins_arduino.h

FeuerSturm:
Slight difference to generic RP2040
PIN_SERIAL2_TX + PIN_SERIAL2_RX are
switched with PIN_WIRE0_SDA + PIN_WIRE0_SCL


FeuerSturm:
Sorry @earlephilhower I just noticed my mistake right after you already published a new version :(

PR #610, opened 2022-06-06
Try and upload ASAP instead of hard-waiting 10sec

earlephilhower:
Fixes #609 and a niggling slowness in uploads.  Try to find the drive
in a loop for 10 seconds, instead of only checking once after 10 seconds.


earlephilhower:
I've tested on Windows and Linux and both now upload significantly faster under normal circumstances.

PR #634, opened 2022-06-17
Update PlatformIO board files, add PIO CI

maxgerhardt:
* Pushes `arduino` object higher in JSON file as the first entry in `boot` to keep alphabetical ordering, done in feedpack to PR (https://github.com/platformio/platform-raspberrypi/pull/36#discussion_r899927068)
* Adds PlatformIO to CI
  * builds the Multcore.ino, Fade.ino and CDC_Multi.ino (Adafruit TinyUSB) examples for Pico and Adafruit Feather boards
* upgrades the checkout version to latest stable v3

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -181,9 +181,12 @@ jobs:
 
 # Build a few examples with PlatformIO to test if integration works
   build-platformio:
+    name: Build PlatformIO Examples
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
+    - with:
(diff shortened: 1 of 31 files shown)


earlephilhower (approved):
Great, thanks for adding CI tests!

I'll see if I can use your PIO script to clean up the release publish process.  I didn't think `python` still worked on Ubuntu and you needed `python3`, for example.

maxgerhardt:
`python` on regular Ubuntu is indeed Python2 still, but by using that `setup-python@v2` action, it sets python = python 3.10 (3.10.5 to be exact per logs). A test script should indeed call into Python3 there if that action is not used.

PR #656, opened 2022-06-25
docs gpio (digital.rst)

sstaub:
- moved Pad Strength from pins.rst to digital.rst
- add information about input modes


earlephilhower (approved):
LGTM.  Always happy to have help with the docs!

PR #659, opened 2022-06-26
Add notes on TinyUSB documentation and quirks

### LINKED ISSUE #648: Document quirks on Adafruit TinyUSB Examples
I just went to try out the very first Adafruit USB example linked in the library, which is https://github.com/adafruit/Adafruit_TinyUSB_Arduino/blob/9c03087bcdbe1dc4faff3e2a41c3bc109ee2a554/examples/CDC/cdc_multi/cdc_multi.ino. The example is supposed to create two USB serial devices.

To my surprise this did not work at all -- reason being, that the Adafruit USB CDC class checks the config

```cpp
void Adafruit_USBD_CDC::begin(uint32_t baud) {
  (void)baud;

  // already called begin()
  if (isValid()) {
    return;
  }

  // too many instances
  if (!(_instance_count < CFG_TUD_CDC)) {
    return;
  }
```

and `CFG_TUD_CDC` is the maximum allowed number of CDC devices, which in the RP2040 configuration

https://github.com/adafruit/Adafruit_TinyUSB_Arduino/blob/9c03087bcdbe1dc4faff3e2a41c3bc109ee2a554/src/arduino/ports/rp2040/tusb_config_rp2040.h#L58

is **1**. Changing this to `2` and reuploading made the two serial devices appear properly.

Maybe this will confuse some other people too that at least one example of that library is not executable on the RP2040 with the default configuration and documentation should be added to it?

Another would be to make the USB configuration file, or at least the number of maximum devices, configurable with a macro / custom user config header file.

maxgerhardt:
Closes #648 by documentating said findings.

PlatformIO specific instructions were tested locally and found to work.

### DIFF
--- a/docs/usb.rst
+++ b/docs/usb.rst
@@ -46,3 +46,61 @@ Also, this stack requires sketches to manually call
 sketch upload from the IDE.  If a sketch is run without this command
 in ``setup()``, the user will need to use the standard "hold BOOTSEL
 and plug in USB" method to enter program upload mode.
+
+Adafruit TinyUSB Configuration and Quirks
+-----------------------------------------
+
+The Adafruit TinyUSB's configuration header for RP2040 devices is stored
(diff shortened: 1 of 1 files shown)


earlephilhower (approved):
Thx for the doc update!

PR #663, opened 2022-06-29
Add support for new SD/RTC and SubGHz boards

PontusO:
This PR adds support for our new boards Challenger RP2040 SubGHz and Challenger RP2040 SD/RTC.


earlephilhower (changes requested):
I just updated `makeboards.py` to make the JSON update automatic.  If you can do a merge with `master` and re-run `tools/makeboards.py` we should be good here.

Thx!

earlephilhower (approved):

PR #693, opened 2022-07-20
Support original Pico with ROM B0

### LINKED ISSUE #689: Floats in print stream cause processor crash in v2.3.x
Using the print stream with floats causes the processor to crash, for example:

Serial.println(1234.56);

or 

tft.println(1234.56);

This affects v2.3.0, v2.3.1 and v2.3.2. Works OK in v2.2.2

earlephilhower:
The SDK disabled float/double support for some functions by default on the
original B0 ROMs.  Manually re-enable it.

Fixes #689


earlephilhower:
I just dug up an unused Pico from release day and was able to a) reproduce the crash, and b) test the fix here.  All good now!

PR #686, opened 2022-07-19
Correct PlatformIO build for Pico W, add to CI

maxgerhardt:
Successfull compilation was prevented by missing to add the 

```
compiler.netdefines=-DPICO_CYW43_ARCH_THREADSAFE_BACKGROUND=1 -DCYW43_LWIP=0 -DLWIP_IPV6=1 -DLWIP_IPV4=1 -DLWIP_IGMP=1 -DLWIP_CHECKSUM_CTRL_PER_NETIF=1
```
flags only present in the `platform.txt` to the `platformio-build.py` (PIO does not automatically read the `platform.txt`).

Also adds the WiFi scan example to the CI with the newly added `rpipicow` board definition.

PR #698, opened 2022-07-21
Add USB VID/PID to Platform.IO HWIDs

earlephilhower:
https://github.com/earlephilhower/arduino-pico/pull/686#issuecomment-1191348503


maxgerhardt:
Maybe I'm paranoid but all the other boards (also in [ststm32](https://github.com/platformio/platform-ststm32/blob/develop/boards/bluepill_f103c8.json#L11-L14)) have the VID and PID on uppercase hex, could we do a `.upper()` on VID and PID?

earlephilhower:
> Maybe I'm paranoid but all the other boards (also in [ststm32](https://github.com/platformio/platform-ststm32/blob/develop/boards/bluepill_f103c8.json#L11-L14)) have the VID and PID on uppercase hex, could we do that to VID and PID?

Done!

maxgerhardt:
Looking good now.

PR #711, opened 2022-07-25
WIP - Enable OTA

earlephilhower:
Insert a stub app that will check for work to be done, and if so do it.  If no
work to be done, jump to original app.

Uploads, including password protected and signed, work through the IDE now.

DONE:
- [X] Replace boot.o with OTA.o (which will include the appropriate boot2)
- [X] Define area to store work to be done, format of work block
- [X] Add in R-O LittleFS code
- [X] Add in UZLIB compression lib
- [X] After upload erase work block and reboot
- [X] Make constants less magic in OTA code
- [X] Add simple file example
- [X] Add compressed example
- [X] Add mDNS for Arduino IDE OTA support
- [X] Add ArduinoOTA class on core and uploader Python script
- [X] Add checksum
- [X] Add Updater() to allow signed updates
- [X] Make uploads power fail safe!
- [X] Add docs


tools/signing.py
```diff
@@ -0,0 +1,87 @@
...
+    parser.add_argument('-p', '--publickey', help='Public key file');
+    parser.add_argument('-s', '--privatekey', help='Private(secret) key file');
+    return parser.parse_args()
+
+def sign_and_write(data, priv_key, out_file):
+    """Signs the data (bytes) with the private key (file path)."""
+    """Save the signed firmware to out_file (file path)."""
+
+    signcmd = [ 'openssl', 'dgst', '-sha256', '-sign', priv_key ]
+    proc = subprocess.Popen(signcmd, stdout=subprocess.PIPE, stdin=subprocess.PIPE, stderr=subprocess.PIPE)
```
maxgerhardt, line 24: For people that want to use signed updates, we should declare the OpenSSL dependency somewhere or substitute with a Python library?

earlephilhower: I'd rather stick w/an OpenSSL dependency.  We ship our own Python3 for the Arduino IDE, and pulling in all the requirements to get a X509 signing module running would be a beast.

How would that requirement be set for P.IO?

maxgerhardt: I wouldn't modify the dependency chain, just forward that responsibility to the user. PIO can also access all the program's in the user's `PATH`, so if they want signed OTA updates, PIO will just call into the same `singing.py` script and it will require `openssl` to be installed. The "standard" user won't turn on signing anyways. I could of course create a OpenSSL package make PlatformIO use it, but unless I get a feature request for that, I wouldn't start doing it.

earlephilhower: Ah, good.  We'll go with letting the (rare) user who wants to sign make sure he has the tools to do so.  It's in the documentation (not in this PR yet) as a requirement, and you'd generally run `openssl` to generate the public and private key anyways.

PR #713, opened 2022-07-26
Added to possibility to have extra compiler directives for a board.

PontusO:
As the title suggests this PR adds functionality for adding an extra compiler -D directive for each board. This adds the possibility to have some control of the compilation phase for libraries that requires different configurations for different platforms.


maxgerhardt:
Could you also add this to the PlatformIO board generation logic at https://github.com/earlephilhower/arduino-pico/blob/1f5139d20942027d9236dca61185fa3edab27f7f/tools/makeboards.py#L191-L203 with the build flags being added to `extra_flags` if present?

JAndrassy:
this would alow only one -D but the build.extra_flags allows any compilator command line parameters

tools/makeboards.py
```diff
@@ -122,6 +122,8 @@ def BuildHeader(name, vendor_name, product_name, vidtouse, pidtouse, vid, pid, p
     print("%s.build.pid=%s" % (name, pid))
     print('%s.build.usb_manufacturer="%s"' % (name, vendor_name))
     print('%s.build.usb_product="%s"' % (name, product_name))
+    if extra != None:
+        print('%s.build.extra_flags=-D%s' % (name, extra))
```
maxgerhardt, line 126: Yeah, maybe for future expansion `extra` should be a `list` and you can just do `for x in extra:`.

PontusO: Agreed, I'll look into that during the week.

earlephilhower: Note this change should also apply to the JSON creation section below, of course.  I'll hold off on this merge until the list addition is in.

Another thing, not related to this PR, @PontusO, is now that there is a network infrastructure, would it make sense to integrate the different network interfaces your boards have to make use of the LWIP/etc.?

For example, on the ESP8266 using the lwip-ethernet class we've added the WizNet boards and an ENC-xxxx wired ethernet board into the LWIP infra so they can use `WiFiClient`, `WiFiClientSecure`, etc.

JAndrassy: @earlephilhower that would require some kind of MAC level firmware for esp32

earlephilhower: > that would require some kind of MAC level firmware for esp32

Well, the WiFi negotiation/etc., would definitely be from scratch, but if there is a naked Ethernet packet send/receive option already available it's pretty close to fitting in to the existing lwip-ethernet/CYW43 framework.

> does this core support w5500 or enc28j60 as network interface? 

It's very close to supporting it.  I stole the infrastructure from the ESP8266 where are drivers for those chips, but as usual I hit the code with a hatchet so probably broke some things in a *minor* way.  They won't run with no code changes, but they should go with only minimal ones.

Only so many hours in the day, and right now my focus is on OTA, but importing those is in the queue (esp. since I got a donated W5500 board and have an ENCJ boad hanging around already)...

PontusO:
I think so, this allows for a list of macros to be defined.

tools/makeboards.py
PontusO: > Note this change should also apply to the JSON creation section below, of course. I'll hold off on this merge until the list addition is in.
> 
> Another thing, not related to this PR, @PontusO, is now that there is a network infrastructure, would it make sense to integrate the different network interfaces your boards have to make use of the LWIP/etc.?
> 
> For example, on the ESP8266 using the lwip-ethernet class we've added the WizNet boards and an ENC-xxxx wired ethernet board into the LWIP infra so they can use `WiFiClient`, `WiFiClientSecure`, etc.

@earlephilhower It makes perfect sense to integrate support for our boards into the work you have made. I did some initial investigation and it looks easy enough to get the wifi control stuff in place. Lets discuss what needs to be done.

JAndrassy:
how would you enter something like this?
```
mkrwifi1010.build.extra_flags=-DUSE_ARDUINO_MKR_PIN_LAYOUT {build.usb_flags} -D__SAMD21G18A__
```

maxgerhardt:
The `-D` is forced, so injecting `{build.usb_flags}` wouldn't be possible with this mehod. But I guess they can add that stuff in `platform.txt` compile commands directly? Hm.

JAndrassy:
the `{somename}` can be a key with a value selected in Tools menu with a menu defined in boards.txt

PR #741, opened 2022-08-06
add board ElectronicCats Hunter Cat NFC

sabas1080:
Hi

please add Electronic Cats HunterCat NFC RP2040

- https://electroniccats.com/store/hunter-cat-nfc/

Thanks


earlephilhower (changes requested):
You need to update `makeboards.py` and add it to the PR or else CI will fail.

earlephilhower (approved):
LGTM!

PR #745, opened 2022-08-08
Fix typo in XIAO board name

maxgerhardt:
https://github.com/maxgerhardt/platform-raspberrypi/pull/12


earlephilhower (approved):
Thanks again, as always!

Issue #751, opened 2022-08-11
FreeRTOS : configTIMER_TASK_STACK_DEPTH is too low which is causing system hang


freeasabeer:
Currently configTIMER_TASK_STACK_DEPTH is set to 80 which is below configMINIMAL_STACK_SIZE which is set to 256.
This too low value is causing an exhaustion of the available stack for timer stuff resulting in a system hang while using FreeRTOS timers.
At least the timer stack should be set to 256 but I would recommend an even higher value like 1024 (for reference it is 2048 on the ESP32).

Great job done anyway on this project !

earlephilhower:
Good observation.  Each SIZE is actually 4-bytes, but even so 80 is probably a bit small especially if there are interrupts coming in.   1024 = 4K = the standard Pico SDK hardware stack size, so that seems like a reasonable size.

Would you like to do a PR changing this value?

freeasabeer:
Yes, I will do.

PR #752, opened 2022-08-11
Increase FreeRTOS timer task stack to 1024

freeasabeer:
PR to fix issue #751 FreeRTOS : configTIMER_TASK_STACK_DEPTH is too low which is causing system hang
Original value of configTIMER_TASK_STACK_DEPTH is 80 which is too low.
Proposal is to increase to 1024.


earlephilhower (approved):
Thanks!

PR #779, opened 2022-08-20
Port ESP8266 DNSServer

earlephilhower:
Add a simple DNS server for AP mode


khoih-prog:
It seems all is well.

Could you change line 1 of `DNSServer.cpp` to

```
#include <WiFi.h>
```

from

```
#include <ESP8266WiFi.h>
```

to avoid confusion.

PR #785, opened 2022-08-21
Report "no filesystem" on OTA uploads w/o a FS

earlephilhower:
Give a meaningful error when OTA is attempted against a chip which does
not have a filesystem configured.


earlephilhower:
````
Detected Signed Update. /tmp/arduino_build_69357/SignedOTA.ino.bin.signed will be uploaded instead.
13:18:59 [ERROR]: Bad Answer: ERR: no filesystem
````

PR #787, opened 2022-08-21
Ensure OTA errors are sent in single UDP packet

earlephilhower:
OTA text error messages were getting lost because they were sent in
multiple UDP packets, one per print(). Now collect the full error and
report in a single print, allowing text messages to appear in ESPOTA.


earlephilhower:
Before this patch:

````
14:23:47 [ERROR]: ERROR[12]: 
````

After the patch
````
14:58:05 [ERROR]: ERROR[12]: Signature verification failed
````

PR #793, opened 2022-08-23
Fix functions in docs: RP2040 Helper Class

Phoenix1747:
Very small change, but just for completeness I thought I'd just add the missing stuff right now.


earlephilhower (approved):
Thx!

PR #806, opened 2022-08-28
Identify boards when in compound USB device mode

earlephilhower:
Allows the IDE to detect boards when using the Keyboard, Joystick,
or Mouse libraries.

Thanks @DaleMitchell for the idea!


DaleMitchell:
Thanks for implementing it! 😊

PR #812, opened 2022-08-30
Fix WiFi MAC Address string output

cfilipem:
This PR reverses the printing order of the MAC Address octets in methods `softAPmacAddress` and `macAddress` of the WiFi class.


earlephilhower (approved):
LGTM.  Thanks!

PR #819, opened 2022-08-31
Add bidirectional bulk SPI transfer, update SdFAT

earlephilhower:
Should speed up SD transfers significantly.

See #801


FerGT50:
Thank you. You are always supportive. Arduino-pico is a game changer.

PR #825, opened 2022-09-01
Make USB, UART, SPI, and I2C objects non-copyable

earlephilhower:
Objects that correspond to a single physical device shouldn't be copyable
as their states could end up becoming completely confused if the copy
changes them and the main global object is used later on.

Avoid it by removing the copy and copy move constructors of the
SPI, SerialUSB, SerialUART, and Wire objects.

This mirrors the Arduino MBED core (which uses template NonCopyable<>)


earlephilhower:
Looking around it seems other cores bend over backwards to allow this (storing all object data in a std::shared_ptr<> for example) so we should probably do the same here.

Issue #839, opened 2022-09-06
[Feature Request] Enabling static allocation in FreeRTOS


godario:
Hi,
I was trying to create a task using static allocation ([`xTaskCreateStatic`](https://www.freertos.org/xTaskCreateStatic.html)) but I saw that this feature is currently disabled: Is there any particulare reason why this feature is currently disabled?
Taking into account what is written [here](https://www.freertos.org/a00110.html#configSUPPORT_STATIC_ALLOCATION) and looking around into your FreeRTOS implementation, currently everything should be in order to enable static allocation without further changes.
I tried to enable `configSUPPORT_STATIC_ALLOCATION` in `FreeRTOSConfig.h` and write a small example of statically created task and everything seems to work.

Is there anything that escapes me that can be implemented in order to fully support static allocation?
Thank you.

[static_task_example.zip](https://github.com/earlephilhower/arduino-pico/files/9495973/static_task_example.zip)

earlephilhower:
@godario would you like to make the PR with this change and your example?  I had it disabled only because I thought that it made it *mandatory* to statically allocate, not *optional.*  So, I'm fine with enabling the add'l support if it doesn't break anything.

earlephilhower:
Thanks for the PR!

PR #842, opened 2022-09-07
Enabled static memory allocation with example

godario:
Referred to issue #839


earlephilhower (commented):
Thanks!  Some minor things w/CI are causing failure.

Can you please add `(void) pvParameters;` to line 53 and 65 to avoid the "unused argument warning", and run the `./tests/restyle.sh` to get it into the proper formatting to match the rest of the core?

Issue #861, opened 2022-09-20
Watchdog timer support?


brentru:
@earlephilhower The Pico SDK provides [hardware_watchdog](https://raspberrypi.github.io/pico-sdk-doxygen/group__hardware__watchdog.html) support. Is this avaliable in Arduino Pico? Would be useful for Raspberry Pico W projects for hardware reset/reboot.

If not, I'd be OK with adding it. I have familiarity with the Pico SDK but not arduino-pico core, if you can point me in the right direction in the core?

brentru:
@earlephilhower  I am volunteering, we (adafruit) need this functionality. 

I do have a  follow up Q, in the `RP2040` class, adding [a call for the low-power sleep mode](https://raspberrypi.github.io/pico-sdk-doxygen/group__sleep.html#gae617a6842d4f3a192064c4354b88fcff) as `#include <pico/sleep.h>` throws:

```
/Users/brentrubell/Library/Arduino15/packages/rp2040/hardware/rp2040/2.5.2/cores/rp2040/RP2040Support.h:31:10: fatal error: pico/sleep.h: No such file or directory
   31 | #include <pico/sleep.h>
      |          ^~~~~~~~~~~~~~
```

Is this package not included from the SDK? Am I including it incorrectly?

PR #862, opened 2022-09-20
Add WDT and sleep functions to RP2040 Helper

brentru:
Addresses and resolves https://github.com/earlephilhower/arduino-pico/issues/861 by adding two function helpers for basic hardware watchdog access and one function for entering low-power sleep (WFE) mode.

cc @earlephilhower

### DIFF
--- a/cores/rp2040/RP2040Support.h
+++ b/cores/rp2040/RP2040Support.h
@@ -319,7 +319,7 @@ public:
     }
 
     void wdt_reset() {
-        watchdog_reset();
+        watchdog_reboot();
     }
 
     const char *getChipID() {
--- a/cores/rp2040/RP2040Support.h
+++ b/cores/rp2040/RP2040Support.h
@@ -319,7 +319,7 @@ public:
     }
 
     void wdt_reset() {
-        watchdog_reboot();
+        watchdog_update();
     }
 
     const char *getChipID() {
--- a/cores/rp2040/RP2040Support.h
+++ b/cores/rp2040/RP2040Support.h
@@ -28,6 +28,7 @@
 #include <hardware/structs/systick.h>
 #include <pico/multicore.h>
 #include <pico/util/queue.h>
+#include <pico/sleep.h>
 #include "CoreMutex.h"
 #include "ccount.pio.h"
 #include <malloc.h>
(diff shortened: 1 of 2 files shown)


cores/rp2040/RP2040Support.h
```diff
@@ -314,6 +315,18 @@ class RP2040 {
...
 
+    void wdt_begin(uint32_t delay_ms) {
+        watchdog_enable(delay_ms, 1);
+    }
+
+    void wdt_reset() {
+        watchdog_reset();
+    }
+
+    void sleep_until_ms(uint32_t ms) {
```
earlephilhower, line 326: This isn't needed.  `delay()` already does `sleep_ms`

earlephilhower (changes requested):

cores/rp2040/RP2040Support.h
earlephilhower:
````
In file included from /home/runner/arduino_ide/hardware/pico/rp2040/cores/rp2040/Arduino.h:112,
                 from /home/runner/work/arduino-pico/arduino-pico/build.tmp/sketch/AnalogReadSerial.ino.cpp:1:
/home/runner/arduino_ide/hardware/pico/rp2040/cores/rp2040/RP2040Support.h:31:10: fatal error: pico/sleep.h: No such file or directory
   31 | #include <pico/sleep.h>
      |          ^~~~~~~~~~~~~~
compilation terminated.
exit status 1
````

`pico/time.h` has the `sleep_ms` call definition BTW.  If you do need `pico/sleep.h` then we need to update `tools/libpico/Cmakefile.txt`, rebuild libpico, and update the `lib/platform_inc` file.

**Review by earlephilhower — APPROVED** [review]
LGTM, thanks!

**Comment by brentru** [comment]
@earlephilhower  Sweet! Could you create a release when merged? Need this as a dependency for Adafruit_SleepyDog.

**Comment by brentru** [comment]
Missed that, thanks. Was in my local pico BSP but not upstream.

File: cores/rp2040/RP2040Support.h [hunk 2]
```diff
@@ -314,6 +314,14 @@ class RP2040 {
         reboot();
     }
 
+    void wdt_begin(uint32_t delay_ms) {
+        watchdog_enable(delay_ms, 1);
+    }
+
+    void wdt_reset() {
+        watchdog_reboot();
```
earlephilhower, line 322: Err, don't you mean `watchdog_update()`?  This call should also fail because it actually resets the chip (see the `RP2040::reboot()` method in this file) and needs add'l params.

earlephilhower (changes requested):

cores/rp2040/RP2040Support.h
brentru: Yeah, should be watchdog_update (https://raspberrypi.github.io/pico-sdk-doxygen/group__hardware__watchdog.html#ga329b748919954f8b48c58049115a5c54), to reload the counter with the specified time, not reset it.

earlephilhower (approved):

PR #905, opened 2022-10-09
Clear LWIP started flag on LWIPIntfDev::end

### LINKED ISSUE #884: If WiFi.softAPConfig called after WiFi.begin, softAP ip is not correct.
Hi, First of all, Thank you for this project.

When I call WiFi.softAPConfig after calling WiFi.begin, DNS not works, and softAP ip is not correct.
If you need more information please let me know.

use case 1: The app tries to connect to wifi and fails, so starts config manager.
  - With non existing access point credentials or the credentials not yet set. (WIFI_SSID, WIFI_PASS)

use case 2: The app tries to connect to wifi and succeeds, User starts the configuration manager for setting change.
  - With existing access point credentials. (WIFI_SSID, WIFI_PASS)
  
result: IP is not set correctly.  
expected ip :172.217.28.1, shown ip: 192.168.42.1

Even sometimes , when I try to connect to pico w from android, android can't get an ip address from pico w.

if  these lines commented and only `setServer();` is called, DNS works and softAP ip is correct.
```C++
  WifiConnected = wifiSetup(WIFI_SSID, WIFI_PASS);
  if (!WifiConnected) setServer();
```

This is the code that I tested with Raspberry Pi Pico W
 
```C++

// based on 
//https://github.com/earlephilhower/arduino-pico/blob/master/libraries/DNSServer/examples/DNSServer/DNSServer.ino
#include <WiFi.h>
#include <DNSServer.h>
#include <WebServer.h>

const byte DNS_PORT = 53;
IPAddress apIP(172, 217, 28, 1);
DNSServer dnsServer;
WebServer webServer(80);

const char* WIFI_SSID = "xxxxxxx";
const char* WIFI_PASS = "12345678";  

bool serverOn = false;
bool WifiConnected = false;
int WifiDisconTime = 10; //5sec.

bool wifiSetup(const char* ssid, const char* pass) {
  bool connected = false;
  int tryCount = 10;
  // Set WIFI module to STA mode
  WiFi.mode(WIFI_STA);

  // Connect
  Serial.printf("[WIFI] Connecting to %s ", ssid);
  WiFi.begin(ssid, pass);

  // Wait for wifi connection
  while (1) {
    //check wifi connection
    static unsigned long last = millis();
    if (millis() - last > 500) {
      last = millis();
      if (WiFi.status() == WL_CONNECTED) {
        // Connected!
          Serial.printf("\n[WIFI] STATION Mode, SSID: %s, IP address: %s\n", \
                          WiFi.SSID(), WiFi.localIP().toString().c_str());
        connected = true;
        break;
      }
      Serial.print(".");
      tryCount--;
    }
    
    if (tryCount <= 0) break;
    delay(100);
  }
  Serial.println("");
  return connected;
}

void setup() {
  Serial.begin(115200);
  Serial.println("");
  delay(1000);

  // to disable soft AP after restart
  WiFi.mode(WIFI_OFF);
  delay(100);

  WifiConnected = wifiSetup(WIFI_SSID, WIFI_PASS);
  if (!WifiConnected) setServer();
}

void setServer(){
  WiFi.disconnect();
  WiFi.end();
  WifiConnected = false;

  WiFi.mode(WIFI_AP);
  WiFi.softAPConfig(apIP, apIP, IPAddress(255, 255, 255, 0));
  WiFi.softAP("picow", "12345678");
  delay(500);
  Serial.print("AP IP address: ");
  Serial.println(WiFi.softAPIP());

  // modify TTL associated  with the domain name (in seconds)
  // default is 60 seconds
  dnsServer.setTTL(300);
  // set which return code will be used for all other domains (e.g. sending
  // ServerFailure instead of NonExistentDomain will reduce number of queries
  // sent by clients)
  // default is DNSReplyCode::NonExistentDomain
  dnsServer.setErrorReplyCode(DNSReplyCode::ServerFailure);

  // start DNS server for a specific domain name
  dnsServer.start(DNS_PORT, "*", apIP);

  // simple HTTP server to see that DNS server is working
  webServer.onNotFound([]() {
    String message = "Hello World!\n\n";
    message += "URI: ";
    message += webServer.uri();

    webServer.send(200, "text/plain", message);
  });
  webServer.begin();

  serverOn = true;
  Serial.println("Server setup done");
}

void loop() {
  if (WifiConnected) {
    static unsigned long last = millis();
    if (millis() - last > 500) {
      last = millis();
      Serial.print(".");
      WifiDisconTime--;
    }
    if (WifiDisconTime <= 0) {
      setServer();
      WifiConnected = false;
    }
  }
  delay(100);
  if (serverOn){
    dnsServer.processNextRequest();
    webServer.handleClient();
  }
}
```

earlephilhower:
When moving between different modes or even WiFi.begin/ends, any setting of the IPs will fail because the internal flag _started was not cleared. Clear _started on a ::end call.

Fixes #884

### DIFF
--- a/libraries/lwIP_Ethernet/src/LwipIntfDev.h
+++ b/libraries/lwIP_Ethernet/src/LwipIntfDev.h
@@ -408,6 +408,7 @@ void LwipIntfDev<RawDev>::end() {
     RawDev::end();
     netif_remove(&_netif);
     memset(&_netif, 0, sizeof(_netif));
+    _started = false;
 }
 
 


earlephilhower:
Debugging showed the issue clearly...

W/o the fix, code from #884:
````
..........
LwipIntfDev: use config() then begin()
AP IP address: 192.168.42.1
Server setup done
````

W/the fix:
````
..........
AP IP address: 172.217.28.1
Server setup done

````

**Comment by MrGreensWorkshop** [comment]
Hi, It is working. Thank you. How about adding here too. 
```C
    LwipIntfDev(int8_t cs = SS, SPIClass& spi = SPI, int8_t intr = -1) :
        RawDev(cs, spi, intr), _mtu(DEFAULT_MTU), _intrPin(intr), _started(false), _default(false) {
        memset(&_netif, 0, sizeof(_netif));
        _started = false;
    }
```

earlephilhower:
Thanks for the fast feedback;

That bit is already done with the C++ initializer `_started(false)`, so we should be good as is.

MrGreensWorkshop:
Thanks to you!
silly me I didn't saw that. :)

PR #910, opened 2022-10-11
Fix Mac CI size reporting

earlephilhower:
Thanks to @mcspr, import build.sh size reporting fix for Mac which doesn't have `declare` in their bash version.

https://github.com/mcspr/esp8266-Arduino/blob/2651b65135d47bc2ed12295959344734b51d5fc2/tests/common.sh#L63-L90

### DIFF
--- a/tests/common.sh
+++ b/tests/common.sh
@@ -47,7 +47,7 @@ END {
 '
 
     if [ -z "$elf_file" ]; then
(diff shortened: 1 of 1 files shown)


earlephilhower:
CI passes but the reporting is still incorrect.  Closing until a more comprehensive PR can be made.

Issue #917, opened 2022-10-15
Increase frequency limit of AnalogWrite() ?


fjansson:
Would it be possible to increase the upper frequency limit of AnalogWrite()?

Here
https://github.com/earlephilhower/arduino-pico/blob/69ab736cf8fbbba7bf2345accd7a5b03699eefaa/cores/rp2040/wiring_analog.cpp#L45
the frequency is capped at 1 MHz. I think there's no reason for the limit except that the PWM resolution becomes low at high frequencies. 

I want to use AnalogWrite as an easy way to set up PWM as a pulse generator, and I'd like it to reach a few MHz at least. For a project for just myself I would use the Pico SDK's pwm functions, but this time I want to publish the code and keep it as simple as possible.

earlephilhower:
Can you up the frequency limit on that line to 8MHz or 16MHz and give it a test and report back/give a PR?

I don't mind increasing the limit, but I am a little worried about some of the math later over/underflowing and winding up with division by 0s or something causing it to fail when actually setting the parameters.

fjansson:
Testing that, I discovered two problems in analogWrite() as it currently is:
* `pwm_config_set_wrap(&c, analogScale);` should be `pwm_config_set_wrap(&c, analogScale-1);`
The wrap value is not the counter period but the highest counter value before wrapping. Adding '-1' here gives better frequency accuracy, and also makes the output stay high all the time when outputting the full-scale value.

* `(float)(analogScale * analogFreq)` can overflow, since the multiplication is done in uint32. Then the frequency set is wrong.
`((float)analogScale * analogFreq)` works better.

I'll try to make a pull-request soon.

Btw, I found that analogWrite to one pin changes the PWM settings for *all* PWM blocks, which I didn't expect. I wanted to use another PWM block as a frequency counter for testing, and found I had to set up that after the analogWrite.

earlephilhower:
Closed via #918

Issue #10, opened 2021-03-24
EEPROM does not work


marcelvanherk:
Hi,

I managed to compile my Data General Nova simulator with your platform (without tft support). Impressive work! However, EEPROM does not work. I then compiled you eeprom write test, which reports a "failed to commit".

Marcel

PR #918, opened 2022-10-16
Pr analog write

fjansson:
PR for issue #917.

First commit fixes the overflow and wrap value. 

Second commit extends the frequency range to 10 MHz. For this to work also at low system clocks (down to 50 MHz, since that's what is available in the IDE menu), I had to decrease the lower bounds of the range and resolution. Also changes the adjustment loops to allow clkdiv values down to 1.0, to make it possible to manually set
well-chosen frequency and range values to avoid using the prescaler.

I've tested at system clocks 50MHz, 133MHz, 150MHz and with different frequencies. From what I've seen it always works, however the frequency obtained can be off by several percent when ending up with small and non-integer prescaler values. This happens when requesting a high frequency and a large range.

### DIFF
--- a/cores/rp2040/wiring_analog.cpp
+++ b/cores/rp2040/wiring_analog.cpp
@@ -42,9 +42,9 @@ extern "C" void analogWriteFreq(uint32_t freq) {
     if (freq < 100) {
         DEBUGCORE("ERROR: analogWriteFreq too low (%d)\n", freq);
         analogFreq = 100;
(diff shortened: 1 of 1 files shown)


earlephilhower (approved):
Thanks, this looks good.

khoih-prog:
Hi all,

It's better to just use

```cpp
```

```cpp
```

instead of

```cpp
```

PR #920, opened 2022-10-16
Fix #919

### LINKED ISSUE #919: PWM glitching/optimization

It's better to just use

```cpp
```

```cpp
```

instead of

```cpp
```

khoih-prog:
Fix disturbing the PWM operation of remaining pins

Use the MRE in #919 to test. PWM is still working after `analogWrite()`


khoih-prog:
Close to recreate to fix error

PR #921, opened 2022-10-16
Fix #919

### LINKED ISSUE #919: PWM glitching/optimization

khoih-prog:
Fix disturbing the PWM operation of remaining pins

### DIFF
--- a/cores/rp2040/wiring_analog.cpp
+++ b/cores/rp2040/wiring_analog.cpp
@@ -73,14 +73,14 @@ extern "C" void analogWriteResolution(int res) {
 
 extern "C" void analogWrite(pin_size_t pin, int val) {
     CoreMutex m(&_dacMutex);
-    
+
     static bool pwmInitted[PINS_COUNT] = { false };
     static bool pwmScaleInitted = false;
 
     if ((pin > PINS_COUNT - 1) || !m) {
         DEBUGCORE("ERROR: Illegal analogWrite pin (%d)\n", pin);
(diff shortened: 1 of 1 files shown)


khoih-prog:
HI @earlephilhower 

I rechecked but still I couldn't figure out why `Style, Boards, Package (pull_request)` check failed.

earlephilhower:
> I rechecked but still I couldn't figure out why `Style, Boards, Package (pull_request)` check failed.

Looks like space/tab and empty lines w/trailing spaces.  You can run `./tests/restyle.sh` to run astyle to clean up the source changes.

khoih-prog:
Thanks. Restyled done and OK now.

earlephilhower (changes requested):
I think there are a couple problems here.

* The `pwmInitted` global is still there but ignored.
* The local static bool array[] (which probably should just be a `uint32_t` that's a bitmask to save 24 bytes of RAM) only ever can go from 0->1, but never back to 0.
* Same for the `pwmScaleInitted`, it's a 1-way trip and you can never change the scale or range.

I can do a PR with the `pwm_init` change later in the week (traveling now), so no worries about fixing this one if you're pressed for free time.

khoih-prog:
Good that you'll take care of this. I'm currently on something else. Thanks.

earlephilhower:
Superseded by #926

PR #926, opened 2022-10-17
Only init PWM on requested pin

### LINKED ISSUE #919: PWM glitching/optimization

earlephilhower:
Avoid glitches on other pins, fix #919


revell1:
I suspect that this change is causing my previously working code to stop.

Hardware I am using is:
kitronik.co.uk/collections/raspberry-pi-pico-robotics-products/products/5335-autonomous-robotics-platform-for-pico

I have code that was working under V2.6.1 library release, but V2.6.2 and V2.6.3 break my code.
The motors nolonger rotate, just buzz quietly for low speed selection or louder for full speed.
When working, the useful range was from 20% to 100%, any lower than 20%, the motors do not have enough power to rotate the gearbox.

Each motor uses 2 output pins.
Motor 1 uses pins 20,19
Motor 2 uses pins 6,7

My code is doing the following (which worked for V2.6.1)

Onetime initialisation:

analogWriteFreq(100);  // 100 Hz base frequency
pinMode(20, OUTPUT);
pinMode(19, OUTPUT);
pinMode(6, OUTPUT);
pinMode(7, OUTPUT);
analogWriteRange(1023); // attempt to set 10 bit PWM

Example code fragment to control a motor:

Stop motor:
analogWrite(20, 0);
analogWrite(19, 0);

Run motor foward:
analogWrite(20, controlvalue);  // Where controlvalue is in range 0 to 1022
analogWrite(19, 0);

Run motor reverse:
analogWrite(20, 0);
analogWrite(19, controlvalue);  // Where controlvalue is in range 0 to 1022

PR #929, opened 2022-10-18
Add W25Q64JV QSPI /4 Boot2 file

### LINKED ISSUE #927: Pimoroni PGA2040 wont accept Arduino-Core
Hi, 
Thank you for making such a great resource! 

I am having issues getting the core to properly install on a PGA2040, 
Plugging in the PGA2040 as I connect BS to GND, Ardunio IDE uploads the initial sketch fine, reboots the PGA2040 but then the board is not seen by windows or the IDE, unplugging and plugging back in (to expect the IDE to see a port and windows to play "Device Connected Sound") but nothing happens. 
I have to repeat the BOOTSEL - GND connection and replug the board, taking me back to square one of uploading the initial sketch.
I have tried numerous sketches including the RP2040 examples and just a blank one but none seem to install the core after the initial upload..

I have been using the Generic RP2040 board from the board-list, with the 2MB Flash setting, although the board has 8MB..

Would it be worth trying to make a dedicated Variant for the PGA2040?

Thank you for your help
Kind Regards

maxgerhardt:
A simple addition. Expands only the "Generic" variant by a "W25Q64JV QSPI /4" Boot2 option.

I was very tempted to also just add
```py
# Pimoroni
MakeBoard("pimoroni_pga2040", "Pimoroni", "PGA2040", "0x2e8a", "0x1008", 250, "PIMORONI_PGA2040", 8, "boot2_w25q64jv_4_padded_checksum")
```
but the `pimoroni_pga2040` would have just been a copy of the `generic` variant folder -- felt weird.

Resolves #927 more cleanly by giving them the possibility to use their flash chip on Quad-SPI mode ([clock div 4](https://github.com/adafruit/circuitpython/blob/main/ports/raspberrypi/Makefile#L380)) as intended and e.g. used by CircuitPython.


earlephilhower (approved):
Thanks for digging into this new flash.  I think I will do the `makeboards` addition you mentioned just to avoid having the same problem pop up again.  Having a variant just for the flash type isn't unique to this board, IIRC, and the bits are free...

PR #934, opened 2022-10-26
Add a way to return the RSSI value

mecparts:
Added a way to return the RSSI value as shown in this thread:
https://forums.raspberrypi.com/viewtopic.php?t=341774


earlephilhower (approved):
This is great, thanks!  You might want to submit it to the cyw43 driver proper as well...

PR #947, opened 2022-10-31
Async scan + bugfixes

drzony:
This PR:
1. Fixes buffer overflow in getChipId
2. Adds support for asynchronous WiFi scanning
3. Fixes some ESP compatibility issues
4. Fixes SSID() so that a copy of SSID name is kept inside WiFi class. In case the original value is deleted


earlephilhower (approved):
Thx!

PR #948, opened 2022-10-31
cores/rp2040: _freertos.cpp: Avoid memory leak and unnecessary allocation

kholk:
There's a memory leak around the corner: creating the mutex means that we allocate memory for it, but if we cannot find any free slot in the FMMap array, we're simply returning a nullptr without ever freeing the previously allocated memory.

Instead of allocating it out of the free-slot check, do it inside. While at it, also check if the mutex creation succeeded before assigning it to a FMMap slot.


earlephilhower (approved):
Thanks, looks good. Even though it does leak memory in the case you noted, though, the core would crash on a `nullptr` deref almost immediately when it tried to init or grab the returned mutex. :)

There's a minor indent-level (4 spaces, not two) error in the astyle run.

PR #950, opened 2022-11-02
fix SerialPIO and update flyboard2040 board pin definition

flyrainning:
fix SerialPIO receive large data error at 115200
update board flyboard2040core pin definition


cores/rp2040/SerialPIO.cpp
```diff
@@ -230,7 +230,7 @@ void SerialPIO::begin(unsigned long baud, uint16_t config) {
         pio_sm_clear_fifos(_rxPIO, _rxSM); // Remove any existing data
 
         // Put phase divider into OSR w/o using add'l program memory
-        pio_sm_put_blocking(_rxPIO, _rxSM, clock_get_hz(clk_sys) / (_baud * 2) - 5 /* insns in PIO halfbit loop */);
+        pio_sm_put_blocking(_rxPIO, _rxSM, clock_get_hz(clk_sys) / (_baud * 2.025) /* insns in PIO halfbit loop */);
```
earlephilhower, line 233: Can you explain how you got the 2.025 factor, please?

flyrainning: I just try may times, (_baud * 2.025) can work normally both 9600 and 115200.
(_baud * 2) - 5 NOT work at 115200, but (_baud * 2) - 7 is ok,  so I think is either (_baud * 2.025) or (_baud * 2) - 7

earlephilhower: Could you please go with the `- 7` instead of the multiplication?  Physically, the PIO is running the same clock for all bauds and the same number of instructions, so I can kind of understand a fixed offset like `-7` vs `-5`, but I can't justify a variable one.

flyrainning: Already done

earlephilhower (approved):

Issue #878, opened 2022-09-25
Using FreeRTOS attachInterrupt() does hang the core.


beegee-tokyo:
I am using FreeRTOS (used to it from ESP32 and nRF52), I can use **`attachInterrupt()`** there without any problems.

With this BSP I have a strange problem that I do not understand. As soon as I add **`#include <FreeRTOS.h>`**, the core that calls attachInterrupt() hangs.

Example IRQ code that works:
```cpp
volatile uint32_t CNT;

void setup()
{
	Serial.begin(115200);
	while (!Serial){};
	Serial.println("Attach IRQ");
	attachInterrupt(digitalPinToInterrupt(27), ISR_func, RISING);
	// or attachInterrupt(27, ISR_func, RISING);
	Serial.println("Attached IRQ");
}

void ISR_func()
{
	CNT++;
}

void loop()
{
	Serial.println(CNT);
	delay(500);
}
```

Adding only the **`#include <FreeRTOS.h>`** in the first line and the core that calls **`attachInterrupt()`** hangs.
```cpp
#include <FreeRTOS.h>

volatile uint32_t CNT;

void setup()
{
	Serial.begin(115200);
	while (!Serial){};
	Serial.println("Attach IRQ");
	attachInterrupt(digitalPinToInterrupt(27), ISR_func, RISING);
	// or attachInterrupt(27, ISR_func, RISING);
	Serial.println("Attached IRQ");
}

void ISR_func()
{
	CNT++;
}

void loop()
{
	Serial.println(CNT);
	delay(500);
}
```

Not sure what I am doing wrong.

beegee-tokyo:
Little digging and if I remove the **`detachInterrupt(pin)`** from **`attachInterrupt()`** function it seems to work.
```cpp
extern "C" void attachInterrupt(pin_size_t pin, voidFuncPtr callback, PinStatus mode) {
    CoreMutex m(&_irqMutex);
    if (!m) {
        return;
    }

    uint32_t events;
    switch (mode) {
    case LOW:     events = 1; break;
    case HIGH:    events = 2; break;
    case FALLING: events = 4; break;
    case RISING:  events = 8; break;
    case CHANGE:  events = 4 | 8; break;
    default:      return;  // ERROR
    }
    noInterrupts();
    // detachInterrupt(pin); // <<<<<<<<<<<<<<<
    CBInfo cb(callback);
    _map.insert({pin, cb});
[...]

PR #959, opened 2022-11-04
Fix GPIO interrupt on freeRTOS

imwoo90:
#878 Fix GPIO interrupt bug on freeRTOS
Please check this changes


earlephilhower:
@imwoo90 Sorry for the delay.  I have been sidetracked recently and will be away from home so won't be able to look at this in the next 2 weeks or so.  Like you said, it's an involved change and I need to spend some time thinking about it and doing some tests myself.

fietser28:
I ran into the problem this merge request fixes. I tested this and for me it works: Tested it with 2 GPIO pins routed to 2 different ISR routines. The routines notify FreeRTOS tasks (of course using ISR safe methods).

earlephilhower (approved):
Thanks for the fix.  I don't do much work w/FreeRTOS so I really appreciate the help!

PR #962, opened 2022-11-05
Fix analogWrite scaling WRT slices

### LINKED ISSUE #955: Problems with default pwm resolution
Hello, using the default resolution:

static uint32_t analogScale = 255;
static uint32_t analogFreq = 1000;

analogWrite(10, 255); outputs about 25 percent on the output pin.

analogwrite_255

analogWrite(10, 512); outputs about 50 percent on the output pin.

analogwrite_512

analogWrite(10, 100); outputs about 99 percent at the output pin.

analog_write_1000

analog write (10, 1020); outputs 100 percent on the output pin.

analogwrite_1020

Is there a way to set the resolution to be 8bit at frequencies below 1.5KHZ?

earlephilhower:
The PWM internals on the RP2040 are based on 8 slices with independent clocking.  Make sure that we init the PWM slice being used only once per change of frequency/range.

Fixes #955


earlephilhower:
Works w/in one slice, but still have issues when skipping slices.  No need to test yet

earlephilhower:
Latest push actually fixes this.  The issue was we changed the internal "analogScale" values to match the final adjusted ones (i.e. after the 2nd pin write the total scale was set to 1020, not 255) causing the PWM values not to actually be scaled.

PR #964, opened 2022-11-07
Add DatanoiseTV PicoADK board

DatanoiseTV:


earlephilhower (approved):
LGTM now, thanks!

earlephilhower (changes requested):
Looks like one changed file was missing.  Can you check in `package/package_pico_index.template.json` and re-push?

DatanoiseTV:
Ok, hopefully I got it right now. Did I miss something in the guide or are some hints missing? :)

earlephilhower:
....
> Would be great if we could make this a bit standardised, so your I2S library is "plug-and-play" with the boards supporting I2S codecs out-of-the-box.

I see you added these to your `pins.h`, good!  I suggest we can push people to use your named defines (but whatever pin makes sense, of course) in their own pins.h.  I2S is probably pretty rare so it can be on a case-by-case basis like yours.  We can always push them to use that specific define name, but since there is no real I2S HW onboard it's probably not so big of a deal.

earlephilhower (approved):

DatanoiseTV:
@earlephilhower Do you have an estimate for the next release which will include the PicoADK?

earlephilhower:
Later this week there should be a new release.  I want to try and clear out the PRs/issues that accumulated while I was away the past week or two, if possible.

PR #972, opened 2022-11-10
tools: makeboards.py: Allow disabling USB stack

kholk:
RP2040 supports USB but it's not granted that everyone wants to actually enable it on "regular" builds: allow building without USB support when unneeded to allow saving a few bytes of RAM.

------------------------------------------

I'm proposing this change because my project (having 3 round SPI displays, using 3 UARTs and, fun fact, leaves only one single free GPIO on my busy Pi Pico board) really needs every last byte of RAM that I can get and the only reason why I would ever need USB is to flash the board through the bootloader.

P.S.: That project is going to be open source, currently 90% complete, I'll publish the code somewhere when it reaches 100% completion :))

Cheers!

### DIFF
--- a/tools/makeboards.py
+++ b/tools/makeboards.py
@@ -217,10 +217,10 @@ def MakeBoard(name, vendor_name, product_name, vid, pid, pwr, boarddefine, flash
         BuildExceptions(n)
         BuildDebugPort(n)
         BuildDebugLevel(n)
-        if a == "picodebug":
-            BuildWithoutUSBStack(n)
-        else:
+        if a != "picodebug":
             BuildUSBStack(n)
+        if a != "picoprobe":
+            BuildWithoutUSBStack(n)
         if name == "rpipicow":
             BuildCountry(n)
         BuildIPStack(n)


earlephilhower (changes requested):
This logic doesn't make sense to me and doesn't do what you way you're looking for.

If you're looking for a way to disable the USB stack there is already a way, add  `-DNO_USB=1` to your `platform_local.txt` or your Platform.IO ini file.

kholk:
The rationale behind this change is that while it's easy for a more experienced person to modify the txt manually, or to add a platform_local.txt, then delete the Arduino IDE cache (because sometimes it doesn't work if you don't delete the cache), then finally restart the IDE and enjoy...... 

This change will simply allow to select "No USB" from the IDE menu, allowing to switch between enabling or disabling the USB stack on the fly before rebuilding.

There are surely other ways, but this one will greatly increase the time savings for such usecase.

earlephilhower:
Ah, I see what you wanted to do here.  Isn't the way I'd have done it, but I think it may be OK.   But you need to run the change and check in the modified `boards.txt` to pass CI.

earlephilhower:
Closing as combined w/#1013

Issue #975, opened 2022-11-13
Please add the Waveshare RP2040 Zero


IAmOrion:
Could you please add the Waveshare RP2040 Zero to the board list

[waveshare_rp2040_zero.zip](https://github.com/earlephilhower/arduino-pico/files/10007659/waveshare_rp2040_zero.zip)

I don't understand how to know what I need for the "boot2" part of this:
def MakeBoard(name, vendor_name, product_name, vid, pid, pwr, boarddefine, flashsizemb, boot2, extra = None):

So I only have:

MakeBoard("waveshare_rp2040_zero", "Waveshare", "RP2040 Zero", "0x2e8a", "0x0003", "500", "WAVESHARE_RP2040_ZERO", 2, boot2): 

but obv need to know what to put for boot2

Edit: After looking at the Pico SDK stuff, I realise what the boot2 part is now, but on the Waveshare wiki etc they don't reference it anywhere, other than linking to the official raspberry pi pico stuff, so I think 
```
MakeBoard("waveshare_rp2040_zero", "Waveshare", "RP2040 Zero", "0x2e8a", "0x0003", 500, "WAVESHARE_RP2040_ZERO", 2, "boot2_W25Q16JVxQ_4_padded_checksum")
```

EDIT: The above works fine for me.  Updated to use @maxgerhardt boot2 file linked in the comments below

maxgerhardt:
Schematics of Waveshare RP2040 Zero (https://www.waveshare.com/w/upload/4/4c/RP2040_Zero.pdf) say `W25Q16JVUXIQ` is used. You gave it `boot2_w25q080_2_padded_checksum` and it really worked?

IAmOrion:
> Schematics of Waveshare RP2040 Zero (https://www.waveshare.com/w/upload/4/4c/RP2040_Zero.pdf) say `W25Q16JVUXIQ` is used. You gave it `boot2_w25q080_2_padded_checksum` and it really worked?
> 
> If not, can you try [boot2_W25Q16JVxQ_4_padded_checksum](https://github.com/maxgerhardt/pico-boot2-generator/blob/master/generated/Winbond/W25Q16JVxQ/boot2_W25Q16JVxQ_4_padded_checksum.S)?

Bugger me, "woods for the trees" I reckon! Well found though, my apologies for missing that.  
Yes it's been working fine for me with `boot2_w25q080_2_padded_checksum `.  I will recreate the boards.txt now and test again and report back

[...]

PR #980, opened 2022-11-14
Add Waveshare RP2040 Zero

maxgerhardt:
Adds proposed files per #975 from @IAmOrion.

For https://www.waveshare.com/wiki/RP2040-Zero.

Adds the same QSPI Boot2 file used in Micropython.

Adds "W25Q16JVxQ QSPI /4" to the list of selectable flashes for the "Generic" variant.

grafik


earlephilhower (approved):
LGTM, thanks!

Issue #977, opened 2022-11-13
Blink program stops working when built with -DUSE_TINYUSB


delan:
Hi, thanks for making this really useful core!

I’m trying to get the TinyUSB stack working under PlatformIO, but whenever I add -DUSE_TINYUSB, my Pico no longer blinks or sends any text down Serial, though it still shows up at ttyACM0 and can still be flashed with pio upload. Minimal example below.

platformio.ini
```
[env:pico]
platform = https://github.com/maxgerhardt/platform-raspberrypi.git
board = pico
framework = arduino
board_build.core = earlephilhower
build_flags = -DUSE_TINYUSB
; also tried -DCFG_TUSB_CONFIG_FILE=\"tusb_config.h\" with both
; 1. the upstream config linked in the docs https://github.com/adafruit/Adafruit_TinyUSB_Arduino/blob/08c6ec947ae5671f66b9599214059ec73652ecdb/src/arduino/ports/rp2040/tusb_config_rp2040.h
; 2. the config in this repo https://github.com/earlephilhower/arduino-pico/blob/852219caf583f0d3e5ca7fe2c9c725f0f798019a/include/tusb_config.h
```

src/main.cc
```
#include <Arduino.h>
#include <Adafruit_TinyUSB.h>
// also had <FreeRTOS.h> in my actual program

void setup() {
  // upload still works for me without this, despite the docs?!
  // https://arduino-pico.readthedocs.io/en/latest/usb.html#adafruit-tinyusb-arduino-support
  Serial.begin(115200);

  pinMode(LED_BUILTIN, OUTPUT);
}

void loop() {
  digitalWrite(LED_BUILTIN, HIGH);
  delay(500);
  digitalWrite(LED_BUILTIN, LOW);
  delay(500);
  Serial.write("asdf ");
  Serial.flush();
}
```

pio pkg list
```
Resolving pico dependencies...
Platform raspberrypi @ 1.7.0+sha.fec7ed8 (required: git+https://github.com/maxgerhardt/platform-raspberrypi.git)
├── framework-arduino-mbed @ 3.1.1 (required: platformio/framework-arduino-mbed @ ~3.1.1)
├── framework-arduinopico @ 1.20602.0+sha.ec76310 (required: git+https://github.com/earlephilhower/arduino-pico.git#ec7631096b667b8d40663c6da4667e3e0aea5a97)
[...]

PR #981, opened 2022-11-14
Do not require USB init variable for TinyUSB + FreeRTOS

maxgerhardt:
For #977.


maxgerhardt:
This confirmed works for me and the original poster. 

The rationale is that TinyUSB init seems to be in-line and blocking in the core's main.cpp before, which is different to the USB init path taken for the other USB stack (that creates a new FreeRTOS task), so there's no need to wait for a non-existing parallel task to do USB init work and thus the variable can be ignored in the TinyUSB + FreeRTOS case (and NO_USB as before).

Open for comments.

earlephilhower (approved):
Thanks for working on this while I was away!

PR #983, opened 2022-11-14
SCORPIO changes

PaintYourDragon:
This PR incorporates some changes from arduino-pico that I think must’ve been sent to me as a ZIP, because I’m not seeing these in any branches. Mostly boards.txt updates to add SCORPIO support. HOWEVER, before merging, plz look carefully because I made CHANGES that may or may not be correct and/or incomplete…

I noticed the USB PID for a physical plugged-in SCORPIO board is 0x80f7 (vs 0x80f1 that was originally in boards.txt, and which is correct for regular non-SCORPIO Feather RP2040). So I changed all PIDs for adafruit_feather_scorpio to end in '7'. I was UNSURE if the PIDs in the adafruit_feather_scorpiopicoprobe and adafruit_feather_scorpiopicodebug should be similarly changed (e.g. incrementing by 0x6), but DID change -DSERIALUSB_PID to 0x80f7 in all those cases.

I can back out some or all of those if incorrect, and can change the PIDs for the picoprobe & debug variants if that’s the correct thing to do. Once sorted out, I’ll merge into our main & do a philhower PR. Thx.


PaintYourDragon:
WELL. Seems I’ve PR’d in the wrong place. Plz disregard!
So smooth, I’m gonna be the next James Bond.

PR #991, opened 2022-11-20
Use PIN_I2S_ defines if they exist in board pin definition.

DatanoiseTV:
Make the usage more user friendly and use PIN_I2S_ pin definitions if they are defined in the board pins_arduino.h.

For example: https://github.com/DatanoiseTV/arduino-pico/blob/master/variants/datanoisetv_picoadk/pins_arduino.h#L43


earlephilhower (changes requested):
I like your idea, but this really isn't the way to do it with the way the class is architected now.

These defaults should be se set in the `I2S::I2S` constructor and not the `::begin`.  See https://github.com/earlephilhower/arduino-pico/blob/master/libraries/I2S/src/I2S.cpp#L26-L39

DatanoiseTV:
@earlephilhower Would this work for you?
```cpp
I2S::I2S(PinMode direction) {
    _running = false;
    _bps = 16;
    _writtenHalf = false;
    _isOutput = direction == OUTPUT;
#ifdef PIN_I2S_BCLK
    _pinBCLK = PIN_I2S_BCLK;
#else
    _pinBCLK = 26;
#endif

#ifdef PIN_I2S_DOUT
    if (_isOutput == OUTPUT) {
        _pinDOUT = PIN_I2S_DOUT;
    }
#endif

#ifdef PIN_I2S_DIN
    if (_isOutput == INPUT) {
        _pinDOUT = PIN_I2S_DIN;
    }
#endif
    _freq = 48000;
    _arb = nullptr;
    _cb = nullptr;
    _buffers = 8;
    _bufferWords = 16;
    _silenceSample = 0;
}
```

earlephilhower:
That looks good, yes!

DatanoiseTV:
@earlephilhower The weird thing is just the _pinDOUT confusion. It would be good to support both INPUT and OUTPUT and combined mode for I2S audio codec. Do you have an idea how to solve it?

earlephilhower:
The way the state machine works it's not possible to do both shift directions in the same PIO SM.  So you need 2 separate state machines if you want bidirectional.  (They can, of course, be running completely different bit widths, frequencies, etc.)

earlephilhower:
@DatanoiseTV can you update the PR with what we discussed above?  That way it can be included in the next release this Friday or Saturday...

DatanoiseTV:
> @DatanoiseTV can you update the PR with what we discussed above? That way it can be included in the next release this Friday or Saturday...

Sorry, done just now.

libraries/I2S/src/I2S.cpp
```diff
@@ -27,11 +27,26 @@ I2S::I2S(PinMode direction) {
...
     _writtenHalf = false;
+    _isOutput = direction == OUTPUT;
     _pinBCLK = 26;
     _pinDOUT = 28;
+#ifdef PIN_I2S_BCLK
+    _pinBCLK = PIN_I2S_BCLK;
+#endif
+
+#ifdef PIN_I2S_DOUT
+    if (_isOutput == OUTPUT) {
```
earlephilhower, line 38: I think you mean "`if (_isOutput)` since this is a boolean...

libraries/I2S/src/I2S.cpp
```diff
@@ -27,11 +27,26 @@ I2S::I2S(PinMode direction) {
...
+        _pinDOUT = PIN_I2S_DOUT;
+    }
+#endif
+
+#ifdef PIN_I2S_DIN
+    if (_isOutput == INPUT) {
```
earlephilhower, line 44: Same as above...`if (!_isOutput)`...

earlephilhower (changes requested):

earlephilhower:
Thx, but I think there's a couple minor things as noted.  OTW looks good!

earlephilhower (approved):
Thanks!

PR #1000, opened 2022-11-24
Fix memcpy to unallocated memory in EEPROM.cpp

brabl2:
The memcpy to unallocated memory in EEPROMClass::begin() was done in two cases: 
1) The 'size' parameter was not multiple of 256.
2) The begin() was called two times and the 2nd call 'size' was greater than 1st time 'size'. (e.g. 1st size=256, 2nd size=512)

This simple change fixes both of them.

### DIFF
--- a/libraries/EEPROM/src/EEPROM.cpp
+++ b/libraries/EEPROM/src/EEPROM.cpp
@@ -54,7 +54,7 @@ void EEPROMClass::begin(size_t size) {
     } else if (!_data) {
         _data = new uint8_t[size];
     }
-    
+
     _size = size;
 
     memcpy(_data, _sector, _size);
--- a/libraries/EEPROM/src/EEPROM.cpp
+++ b/libraries/EEPROM/src/EEPROM.cpp
@@ -45,7 +45,7 @@ void EEPROMClass::begin(size_t size) {
         size = 4096;
     }
 
-    _size = (size + 255) & (~255);  // Flash writes limited to 256 byte boundaries
+    size = (size + 255) & (~255);  // Flash writes limited to 256 byte boundaries
 
     // In case begin() is called a 2nd+ time, don't reallocate if size is the same
     if (_data && size != _size) {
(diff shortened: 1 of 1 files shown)


earlephilhower (approved):
Thanks, LGTM.  Did you run into any problems with the unmodified code, or just notice the logic error?

brabl2:
First issue I noticed was hanging core1 during EEPROM.commit() running at core0. 
So I studied the EEPROM class, solved my issue (not related to this bug) and also found this bug. 
I was able to hang the pico with unmodified code using the case 1).

greymfm:
@brabl2 How did you fix the issue hanging core1 during EEPROM.commit() running at core0 - I have the same issue and maybe you can give a hint ;-) Thanks.

brabl2:
I'm pausing the loop1 using simple while loop and make sure loop1 is paused before I call the EEPROM.commit() in loop0.
Calling the multicore_reset_core1() seems a little bit brutal.
```
//loop0 :
pause_loop1=1;
while (!paused_loop1) {
  delay(1);
}
EEPROM.commit();

//loop1 :
while (pause_loop1) {
  paused_loop1=1;
  delay(1);
}
paused_loop1=0;
```

greymfm:
Very nice solution. Perfect!

earlephilhower:
@brabl2 , @greymfm can you please post a failing EEPROM write sketch in a new issue?  The EEPROM library should already stop the other core (and make sure it's running in a RAM ISR handler) before doing any flash writes.  If it doesn't that's a bug in the EEPROM lib that should be fixed.

brabl2:
I studied my hanging code and found, that the issue is not in the EEPROM library.
The important thing in the EEPROM.commit() is that it pauses the other core for approx. 30ms (measured on pi-pico and pi-pico-w).
My problem was that the core1 was doing time calculations based on millis() measurement (affected by the EEPROM.commit()).
And the result of this time calculations was integer underflow (0xffffffff) and then almost never-ending delay(0xffffffff); which looked like core1 hang.
The two waiting loops in the post above ensure, that the other core (core1) waits in the safe part of the code.

@earlephilhower : Adding a note to the EEPROM doc, that the EEPROM.commit() function pauses the other core for approx. 30ms, would be helpful.

PR #1001, opened 2022-11-24
Using NO_USB definition would make rp2040 fifo uninitialized

diesel437:
Using PIO_FRAMEWORK_ARDUINO_NO_USB build flag would cause internal rp2040 fifo not being initialized correct.


earlephilhower (approved):
Thanks, looks good!  I adjusted the logic a while back to move the USB init out of the `if()` but didn't remove the `#ifndef`s.  D'oh!

PR #1017, opened 2022-11-29
Speed up ClientContext::write(Stream) by chunking

### LINKED ISSUE #999: Example:  FSBrowser very very slow ; Response time very long > 60s
it seems that "webserver.streamFile" dos not work well. 

I use the original example "FSBrowser.ino" with LittleFS, 
the response time for e.g. the response of the call "serverIP/edit" is very very  long (> 60s).

The alternate test: 
*******************
With a replacemet in function **handleFileRead()**
 
of
````
 if (server.streamFile(file, contentType) != file.size()) {

````
with
````
    String sHTML = file.readString();
    file.close();
    DBG_OUTPUT_PORT.printf("HTML String bytes to send: %u\n", sHTML.length() );
    server.send(200, contentType, sHTML);
````

the Response time is very well.

it is clear that this does not work for larger content.
Just to test with the /edit-page

Arduino IDE: 1.8.19
Arduino-Pico: 2.6.3 
Board: Raspberry Pi Pico W

earlephilhower:
Instead of sending a single packet per byte, send out larger chunks to TCP/IP while doing a ::write(Stream).

Fixes #999

### DIFF
--- a/libraries/WiFi/src/include/ClientContext.h
+++ b/libraries/WiFi/src/include/ClientContext.h
@@ -374,12 +374,22 @@ public:
             return 0;
         }
         size_t sent = 0;
+        uint8_t buff[128];
         while (stream.available()) {
-            char b;
-            b = stream.read();
-            if (write(&b, 1)) {
-                sent ++;
+            // Stream only lets you read 1 byte at a time, so buffer in local copy
+            size_t i;
+            for (i = 0; (i < sizeof(buff)) && stream.available(); i++) {
+                buff[i] = stream.read();
+            }
+            if (i) {
+                // Send as a single packet
+                int len = write((const char *)buff, i);
+                sent += len;
+                if (len != (int)i) {
+                    break; // Write error...
+                }
             } else {
+                // Out of data...
                 break;
             }
         }


kendo55:
Thanks a lot!

for made that improvement in the 
ClientContext.h
I implemented that, and the Example "**FSBrowser.ino**" with LittleFS,  is more perfomend.
but **is still slow**  --> 55kB in ~ 2 sec.

11:55:12.480 -> handleFileRead: /pins.png
11:55:12.480 -> file opened: /pins.png
11:55:14.607 -> file stream end: /pins.png

The question is:" Why a aquivalent Example in a **ESP8266** enviroment, is extremly much more 'faster' "

earlephilhower:
If I remember correctly, @d-a-v changed the Stream API to add special optimizations for this.

Effectively he sends larger packets and doesn't have to read byte by byte into a temp buffer.  Unfortunately, though, that makes the ESP8266 Arduino API incompatible with the standard one.

It would be possible to create a class local scratch pad of, say, 1k and use that here, but on the stack I'm worried about going much over 128 bytes without running into issues.

PR #1030, opened 2022-12-04
Fix File::readString to work with binary data

elipsitz:
Previously, `File::readString` used a C-style string as an intermediate buffer via the `String` += operator. This treats a NUL byte as a terminator, making this function work incorrectly if the File contains binary data.

This commit switches the function to use `String::concat`, which doesn't treat NUL bytes any differently (and is a bit faster, because it doesn't need to use `strlen`).


earlephilhower (approved):
Thanks, LGTM.

PR #1046, opened 2022-12-16
Modified LowPowerMode functions in WiFiClass.cpp/.h for better connection stability

brabl2:
I was not very happy with pico-w WiFi stability, with huge WiFi traffic around the connection from my AP to pico-w was really unstable. 
After some testing I discovered, that something (probably inside cyw43-driver) is periodically delaying wifi operations.
The ping response suffered from periodical delay which went up to 10 seconds in some cases (see the screenshot of ping from my server to pico-w):
pico-w_ping_pm_CYW43_DEFAULT_PMm
After some research I found that the delay is caused by the power saving behavior of the WiFi chip.
There are existing two functions inside the arduino-pico WiFiClass: lowPowerMode() and noLowPowerMode(). 
Both of them are unfortunately using power saving modes (CYW43_AGGRESSIVE_PM and CYW43_DEFAULT_PM).
After reading this: https://forums.raspberrypi.com/viewtopic.php?t=340050#p2037055
and this: https://raspberrypi.github.io/pico-sdk-doxygen/group__cyw43__driver.html#ga808eabf2e62d713990ad2994596cb7d3
I modified WiFiClass power mode functions to allow real no low power functionality (using 0xA11140 value instead of parameters defined above).
Using modified noLowPowerMode() function the connection is much stable (see the ping):
pico-w_ping_pm_0xA11140m
The noLowPowerMode() function must be called after the WiFi.begin(...) function.
Hope this helps other people.

### DIFF
--- a/libraries/WiFi/keywords.txt
+++ b/libraries/WiFi/keywords.txt
@@ -57,7 +57,8 @@ beginEnterprise	KEYWORD2
 setHostname	KEYWORD2
 end	KEYWORD2
 getTime	KEYWORD2
-lowPowerMode	KEYWORD2
+aggressiveLowPowerMode	KEYWORD2
+defaultLowPowerMode	KEYWORD2
 noLowPowerMode	KEYWORD2
 ping	KEYWORD2
 beginMulticast	KEYWORD2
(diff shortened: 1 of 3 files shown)


earlephilhower:
@brabl2 this is awesome, thanks!  I think it would be even better if you could put in a call to `noLowPowerMode()` in the `WiFiClass::begin()/beginAP()` to make the default usage work better.  Right now I'd assume it's using the middle tier power savings mode on startup in the CYW43 driver which shows not so hot perf as you're listed...

brabl2:
Done. 
The noLowPowerMode() is called in the WiFiClass::begin()/beginAP() after _wifi.begin() .

earlephilhower (approved):
Again, awesome and thanks!

PR #1062, opened 2022-12-23
Degz Mizu changed to Viyalab Mizu RP2040

thlstfs:


tools/json/viyalab_mizu.json
```diff
@@ -9,7 +9,7 @@
     },
     "core": "earlephilhower",
     "cpu": "cortex-m0plus",
-    "extra_flags": "-D ARDUINO_DEGZ_MIZU -DARDUINO_ARCH_RP2040 -DUSBD_MAX_POWER_MA=250",
+    "extra_flags": "-D ARDUINO_VIYALAB_MIZU_RP2040 -DARDUINO_ARCH_RP2040 -DUSBD_MAX_POWER_MA=250",
```
earlephilhower, line 12: Do you want to keep the other define `-D ARDUINO_DEGU_MIZU` too?  That way existing code won't need changes.

thlstfs: No. We changed the brand. I just want to change the name of the board. It's same board.

PR #1072, opened 2022-12-28
Speed up ADC reads by not re-initting

earlephilhower:
Only change the ADC mux/GPIO control when things re not yet set up. See #1070

### DIFF
--- a/cores/rp2040/wiring_analog.cpp
+++ b/cores/rp2040/wiring_analog.cpp
@@ -26,6 +26,8 @@
 #include <hardware/pll.h>
 #include <hardware/adc.h>
 
+void __clearADCPin(pin_size_t p);
+
 static uint32_t analogScale = 255;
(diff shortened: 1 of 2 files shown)


cores/rp2040/wiring_analog.cpp
```diff
@@ -121,6 +124,14 @@ extern "C" void analogWrite(pin_size_t pin, int val) {
 
 auto_init_mutex(_adcMutex);
 static int _readBits = 10;
+static int _lastADCMux = 0;
+static bool _adcGPIOInit[29];
```
maxgerhardt, line 128: Optimization nitpick: A `uint32_t` used as a bitmask could have save 25 bytes of memory.

earlephilhower: Optimizing the optimization?  That's a recipe for an infinite loop or black hole! :laughing:

PR #1075, opened 2022-12-29
Universal analogReadTemp()

Phoenix1747:
Fixes `analogReadTemp()` by automatically using the correct ADC resolution and optionally also a different reference voltage.

This does not introduce any breaking changes, everything will work like it did (except, if you wanted a wrong temperature reading on purpose) as the default Vref is 3.3 V.

Users can supply their own custom voltage directly in the `analogReadTemp(float vref)` call to make this change as simple as possible.

Tested it on my setup, works like expected with 3.0 V reference and 12-bit ADC.

### DIFF
--- a/cores/rp2040/Arduino.h
+++ b/cores/rp2040/Arduino.h
@@ -73,7 +73,7 @@ void noInterrupts();
 
 // ADC RP2040-specific calls
 void analogReadResolution(int bits);
-float analogReadTemp();  // Returns core temp in Centigrade
+float analogReadTemp(float vref = 3.3);  // Returns core temp in Centigrade
 
 // PWM RP2040-specific calls
 void analogWriteFreq(uint32_t freq);
--- a/cores/rp2040/wiring_analog.cpp
+++ b/cores/rp2040/wiring_analog.cpp
@@ -172,7 +172,7 @@ extern "C" float analogReadTemp(float vref) {
     adc_select_input(4); // Temperature sensor
     int v = adc_read();
     adc_set_temp_sensor_enabled(false);
-    float t = 27.0f - ((v * vref / (float)_readBits) - 0.706f) / 0.001721f; // From the datasheet
+    float t = 27.0f - ((v * vref / pow(2,_readBits)) - 0.706f) / 0.001721f; // From the datasheet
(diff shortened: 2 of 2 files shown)


earlephilhower (approved):
Thanks, LGTM.  I'll do a quick doc update after merge

1225g:
In analogReadTemp(),

static uint8_t _readBits = 10;
...
//float t = 27.0f - ((v * 3.3f / **4096.0f**) - 0.706f) / 0.001721f; // From the datasheet
float t = 27.0f - ((v * vref / **pow(2, _readBits)**) - 0.706f) / 0.001721f; // From the datasheet

pow(2, 10) != 4096.0f

Phoenix1747:
> In analogReadTemp(),
> 
> static uint8_t _readBits = 10; ... //float t = 27.0f - ((v * 3.3f / **4096.0f**) - 0.706f) / 0.001721f; // From the datasheet float t = 27.0f - ((v * vref / **pow(2, _readBits)**) - 0.706f) / 0.001721f; // From the datasheet
> 
> pow(2, 10) != 4096.0f

Doesn't matter, does it? The 4096 was wrong, that's why I corrected it. Even if you don't call `analogReadResolution()`, the ADC bits will default to 10 bits, otherwise it'll be what you set it to. Then your highlighted equation is correct.

earlephilhower:
@Phoenix1747 I think that @1225g is right here.  `adc_read` doesn't care what `_readBits` is, it's the unmodified HW result.  This function returns in degrees C, so it shouldn't change no matter what the core's `analogReadResolution` is set to.

Phoenix1747:
> @Phoenix1747 I think that @1225g is right here. `adc_read` doesn't care what `_readBits` is, it's the unmodified HW result. This function returns in degrees C, so it shouldn't change no matter what the core's `analogReadResolution` is set to.
Oh, you're right, sorry. I thought we were using `analogRead()` for this and not the stock `adc_read()`. I can do a quick PR to fix this, if you want.

earlephilhower:
Yes, please. :)

Phoenix1747:
Please have a look at #1092.

PR #1084, opened 2023-01-01
Add stereo support, docs for PWM playback

earlephilhower:
Limited to consecutive pins (i.e. GPIOs on the same PWM slice). "For free" with PWM since no add'l DMA, buffers, or IRQs are needed.


libraries/PWMAudio/src/PWMAudio.cpp
```diff
@@ -84,7 +92,9 @@ bool PWMAudio::begin() {
     pwm_config_set_wrap(&c, _pwmScale);
     pwm_init(pwm_gpio_to_slice_num(_pin), &c, true);
     gpio_set_function(_pin, GPIO_FUNC_PWM);
+    gpio_set_function(_pin + 1, GPIO_FUNC_PWM);
     pwm_set_gpio_level(_pin, (0x8000 * _pwmScale) >> 16);
+    pwm_set_gpio_level(_pin + 1, (0x8000 * _pwmScale) >> 16);
```
maxgerhardt, line 97: But this would always initialize _pin and _pin + 1 even when `stereo = false`?

earlephilhower: Yes, thanks for the review!   Fixed with latest push (was setting GPIO to PWM too by accident,).

PR #1092, opened 2023-01-02
Fix temperature reading resolution

Phoenix1747:
`adc_read()` resolution is always 12 bits. Doesn't change with `analogReadResolution()`.

This is basically a partial reversal #1075.

### DIFF
--- a/cores/rp2040/wiring_analog.cpp
+++ b/cores/rp2040/wiring_analog.cpp
@@ -172,7 +172,7 @@ extern "C" float analogReadTemp(float vref) {
     adc_select_input(4); // Temperature sensor
     int v = adc_read();
     adc_set_temp_sensor_enabled(false);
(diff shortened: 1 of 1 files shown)


Phoenix1747:
We could also use the artificially lowered resolution of `analogRead()` by using this line instead of `adc_read()` only:

https://github.com/earlephilhower/arduino-pico/blob/d718b143d2b4075366af9804b6bb76b7884d0133/cores/rp2040/wiring_analog.cpp#L156

What's your opinion on this? This would emulate full control over the (lower-than-12-bit) resolution range

earlephilhower:
I can't think of a reason to lower the returned resolution.  For the `analogReadResolution` you're doing it to be compat with SW from some other HW which has a different ADC resolution.  Here, this is a floating point number and you want every bit of accuracy possible...

earlephilhower (approved):

Phoenix1747:
Sounds good 👍🏻

PR #1096, opened 2023-01-03
Set Right and Left correctly for lsbj format

tomcombriat:
The word select bit (WS) convention is inverted between I2S (down for left) and LSBJ (down for right). This inverts the behavior of this bit for the LSBJ format.

I am sorry about the whitespaces changes…

### DIFF
--- a/libraries/I2S/src/pio_i2s.pio
+++ b/libraries/I2S/src/pio_i2s.pio
@@ -50,17 +50,17 @@ right:
 
 ;                           +----- WCLK
 ;                           |+---- BCLK
-    mov x, y         side 0b11
-left:
-    out pins, 1      side 0b00
-    jmp x--, left    side 0b01
-    out pins, 1      side 0b00
-
     mov x, y         side 0b01
+left:
+    out pins, 1      side 0b10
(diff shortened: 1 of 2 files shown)


libraries/I2S/src/pio_i2s.pio
```diff
@@ -50,17 +50,17 @@ right:
...
-    out pins, 1      side 0b00
-    jmp x--, left    side 0b01
-    out pins, 1      side 0b00
-
     mov x, y         side 0b01
-right:
+left:
     out pins, 1      side 0b10
-    jmp x--, right   side 0b11
+    jmp x--, left    side 0b11
```
earlephilhower, line 56: Don't you need to swap the WCLK polarity throughout the program, too?  The label changes don't affect anything in the generated code and now, I'd assume, L and R are swapped?

tomcombriat: The diff is misleading (see the lines number), I didn't change the labels but inverted the WCLK values only.

earlephilhower: Got it, thx.  I see the shifted SIDESETs later on.

PR #1126, opened 2023-01-17
Add correct boot source to helios

Conor-Burns:
thank you @maxgerhardt


maxgerhardt:
Can we also add it in the array for the General board selection? 

https://github.com/earlephilhower/arduino-pico/blob/0406affd097525eb7eca0bc134b8cb41f42cc3a6/tools/makeboards.py#L68-L75

You did test a firmware built on this boot2 file on real hw?

earlephilhower:
Ping @Conor-Burns .  Did you have a chance to double check this new flash BOOT2 does work properly on the board?  I can add the missing boot2 options to generic after this, no worries.

Conor-Burns:
Hi @earlephilhower sadly not yet, sorry for beeing unresponive. I just tried to build it locally but I'm at a loss on how to test it / compile.
Could you point me in the right direction?

earlephilhower:
No worries!

If you've got a `git clone` you're running from, just select the board from the menus, run an example (I'd suggest the LittleFS SpeedTest since it's very flash I/O heavy) ans make sure it runs w/the new BOOT2.s (the core will compile the .S automatically, no need to do anything other then point it to the right file like you did).

Conor-Burns:
but how do i run from a git clone?

earlephilhower:
No boards manager used, that's only for d/l'ing releases.

If you place the git clone in the proper spot, it will be detected (and say "RP2040/... (in sketchbook") in the Boards menu.  See https://github.com/earlephilhower/arduino-pico#installing-via-git

Conor-Burns:
nvm got it - thanks

Conor-Burns:
Before 
image
After
image

I'd say we keep it as is - what do you think?

earlephilhower:
The issue is more about timing being out of spec.  The new one is more conservative, probably because the flash itself has slower timings, so it will have a little lower performance.  It'll probably work but at certain temperatures or certain slower-but-legal chips you can have timing problems and fail reads or writes.

So, I'd say use the slower one you've just selected for safety.  We're not normally flash speed limited for microcontroller apps.

Conor-Burns:
Okay thank you for the clarification :+1: 

Then I'd say it's ready to merge.

Conor-Burns:
🙏

PR #1130, opened 2023-01-20
Add serial break

Beirdo:
I need the capability to know when a break signal is received on the hardware UART.  This is my implementation thereof.


earlephilhower:
@Beirdo can you run `./tests/restyle.sh` to fix the formatting issues (i.e. space not tabs, I think)?  Or if you're under Windows then I think manual tab->space will clear things up.

Also, is this the "standard Arduino way" of handling this?  I did a quick search and couldn't find anything for polling on BREAK.  Just curious if we're breaking new ground or just following what's already out there.

earlephilhower:
Ninja'd :)

Beirdo:
> @Beirdo can you run `./tests/restyle.sh` to fix the formatting issues (i.e. space not tabs, I think)? Or if you're under Windows then I think manual tab->space will clear things up.
> 
> Also, is this the "standard Arduino way" of handling this? I did a quick search and couldn't find anything for polling on BREAK. Just curious if we're breaking new ground or just following what's already out there.

I have not found a "official" way to poll on BREAK, but only to send BREAK (if using the new API, it seems).

cores/rp2040/SerialUART.cpp
```diff
@@ -365,6 +366,26 @@ SerialUART::operator bool() {
     return _running;
 }
 
+void SerialUART::setBreakReceived() {
```
earlephilhower, line 369: I don't think there is any reason for user code to fake push a BREAK event (it can't fake push chars in the receive buffer, for example).  There may also be issues w/taking the mutex in an IRQ while in FreeRTOS.

Could we just set the flag in the `_handleIRQ` call instead and drop this method?

earlephilhower (changes requested):
LGTM with the one method removal.

Should we add a method to send a BREAK too?

cores/rp2040/SerialUART.cpp
```diff
@@ -365,6 +366,26 @@ SerialUART::operator bool() {
...
+    if (!_running || !m) {
+        return;
+    }
+
+    _break = true;
+}
+
+bool SerialUART::getBreakReceived() {
+    CoreMutex m(&_mutex);
+    if (!_running || !m) {
```
earlephilhower, line 380: Also, this should probably be the `_fifoMutex` and not the global object one.  The reason is that this flag is essentially now an extension of the object's FIFO.  OTW there is a potential race condition between lines 384 and 385...

earlephilhower: Also, since it's driven off the UART FIFO it should do the same as `::peek` and `::read` to make sure to pump out any chars from the HW FIFO to the SW one (which is where we'd detect the BREAK).

i.e. 
````
    if (_polling) {
        _handleIRQ(false);
    } else {
        _pumpFIFO();
    }
````

OTW I can see a case where the IRQ doesn't actually pump the FIFO because the SW is doing some Serial call that disallows it, leaving the BREAK stuck in the HW side and never detected.

Sorry. BTW, for the piecemeal notes.  The concurrency w/the UARTs is kind of subtle, especially in multi-core, multi-threaded stuff like we have here!

Comment (Beirdo) [reply cores/rp2040/SerialUART.cpp:380]: Good points, both.   I will take a look at redoing this later today.

File: cores/rp2040/SerialUART.cpp (hunk 1 above)
Comment (Beirdo) [reply cores/rp2040/SerialUART.cpp:369]: sure.  That should work.

File: cores/rp2040/SerialUART.cpp [hunk 3]
```diff
@@ -365,6 +366,23 @@ SerialUART::operator bool() {
...
+        return false;
+    }
+
+    if (_polling) {
+        _handleIRQ(false);
+    } else {
+        _pumpFIFO();
+    }
+
+    bool break_received = _break;
```
earlephilhower, line 381: I think the mutex lock needs to be here around only the dangerous bit, and use the raw mutex handler:
````
uint32_t owner;
mutex_enter(&_fifoMutex, &owner);
bool break_received = _break;
_break = false;
mutex_exit(&_fifoMutex);
return break_received;
````

I also think I goofed on my `overflow` implementation due to this same race condition.  I'll fix that after we merge here.  😭  Multicore/thread with shared HW resources is hard!

File: cores/rp2040/SerialUART.cpp [hunk 4]
```diff
@@ -365,6 +366,23 @@ SerialUART::operator bool() {
     return _running;
 }
 
+bool SerialUART::getBreakReceived() {
+    CoreMutex m(&_fifoMutex);
```
earlephilhower, line 370: Drop the `CoreMutex` and adjust the condition only on `_running`, please.  If we lock here the IRQ stuff will be blocked from running since the mutex will be taken,

earlephilhower (changes requested):
Minor adjustment and I believe we're good.  You also helped me find another issue I'll need to fix w/a flag set in the IRQ as well!

cores/rp2040/SerialUART.cpp
Beirdo: Yeah, it is :)  OK, let me fix that.

cores/rp2040/SerialUART.cpp
Beirdo: OK

earlephilhower (approved):
LGTM now, thanks!  I appreciate your going back and fixing stuff so fast.

Beirdo:
> 

Not a problem.  Of course, I missed the _ in the end when releasing, but it seems to be all fixed.  Thank you :)

PR #1160, opened 2023-02-06
pdm re-port(#1156)

TomKong666:
pulled in the changes from arduinocore-mdbed, leaving the dynamics of pio/sm and dma_channel assignment in arduino-pico untouched. Due to this, additional clean-up has to be done in `PDM::end` compared to arduino-mbed.  Begin and end can now be called as often as desired, but works almost ( fail in one of about every 50-100 end-begin cycles ). Error Scenario now is either all data returned by `PDM::read` is -128's or toally rubbish looking like a missing byte swap for the shorts or a mismatch in the byte index to build the shorts. For now, observed only when forcing a movement from one sm to another of the same pio by claiming the former sm. Forcing load of pdm_pio_program to new sm by resetting `_offset` in PIOProgram::prepare after `pio_remove_program` in  `PDM::end` doesn't help. Where is `pdm_pio_programm` after `PDM::end` ? But before diving to deep, please give me a hint if this PR makes sense to you.

### DIFF
--- a/libraries/PDM/src/PDM.h
+++ b/libraries/PDM/src/PDM.h
@@ -38,12 +38,19 @@ public:
 
     //PORTENTA_H7 min -12 max 51
     //NANO 33 BLE SENSe min 0 max 80
+    //NICLA_VISION min 0 max 8
     void setGain(int gain);
     void setBufferSize(int bufferSize);
     size_t getBufferSize();
 
     // private:
     void IrqHandler(bool halftranfer);
+
+    // Hardware peripherals used
+    uint _dmaChannel;
+    PIO _pio;
+    int _smIdx;
+    int _pgmOffset;
 
 private:
     int _dinPin;
@@ -56,11 +63,13 @@ private:
     int _gain;
     int _init;
 
+    int _cutSamples;
+
     // Hardware peripherals used
-    uint _dmaChannel;
-    PIO _pio;
-    int _smIdx;
-    int _pgmOffset;
+    // uint _dmaChannel;
+    // PIO _pio;
+    // int _smIdx;
+    // int _pgmOffset;
 
     PDMDoubleBuffer _doubleBuffer;
 
--- a/libraries/PDM/src/PDM.h
+++ b/libraries/PDM/src/PDM.h
@@ -46,12 +46,6 @@ public:
     // private:
     void IrqHandler(bool halftranfer);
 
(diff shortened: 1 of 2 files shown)


earlephilhower:
I'll need to spend some time looking at it, but since you've actually run this code I'm pretty OK about merging since the existing code is pretty hokey.

That said, if you can run `tools/restyle.sh` (if you're in Linux/Mac) to format things because the astyle job is failing.  4 space indents, not 2, is what I ended up settling on for actual code.  Sketches are kept at 2-space indent like the IDE.  You could also manually open up the failed job and the diffs are pretty simple.

libraries/PDM/src/rp2040/PDM.cpp
```diff
@@ -118,14 +138,30 @@ int PDMClass::begin(int channels, int sampleRate) {
...
+
+    if (_init != 1) {
+        return;
+    }
+
+    dma_channel_set_irq0_enabled(_dmaChannel, false);
     dma_channel_abort(_dmaChannel);
+    dma_channel_unclaim(_dmaChannel);
+    irq_remove_handler(DMA_IRQ_0, dmaHandler);
+    pio_remove_program(_pio,  &pdm_pio_program, _pgmOffset);
```
earlephilhower, line 158: You don't want to do this.  The way `PIOProgram` works is that it cached the uploaded address of a program so that multiple instances can be run without needing to upload multiple programs (i.e. you can do 4 SWSerial ports with only 1 pio-program in memory).  Leave the PIO program in memory and then the next time you begin the `PIOProgram` call will return the already-installed address

libraries/PDM/src/PDM.h
```diff
@@ -38,13 +38,20 @@ class PDMClass {
...
     //NANO 33 BLE SENSe min 0 max 80
+    //NICLA_VISION min 0 max 8
     void setGain(int gain);
     void setBufferSize(int bufferSize);
     size_t getBufferSize();
 
     // private:
     void IrqHandler(bool halftranfer);
 
+    // Hardware peripherals used
```
earlephilhower, line 49: Why would these be made public?  Nobody but the class should mess with the PIO program, no?

earlephilhower (changes requested):
Just some little things here.

earlephilhower (approved):
Thanks!

PR #1178, opened 2023-02-13
Fix SD.H FILE_WRITE mapping

earlephilhower:
O_RDWR != O_READ|O_WRITE.  Posix is weird.  Thanks @mcspr


mcspr:
*(c++ constexpressions silently cry in the corner; maybe some other fix would adopt them)*

earlephilhower:
Sorry.  Only so much time during half-time and nobody every accused me of being anything other than a C99 hack.  :laughing:

PR #1185, opened 2023-02-14
add pyocd with picoprobe upload

hathach:
picoprobe have switched to use CMSIS DAP protocol in latest release 
- https://github.com/raspberrypi/picoprobe/blob/master/src/picoprobe_config.h#L81

The current way of using openocd doesn't work anymore. This PR add pyocd as uploading tool, tested with pyocd 0.31.0 and works well

```
Sketch uses 54364 bytes (2%) of program storage space. Maximum is 2093056 bytes.
Global variables use 9388 bytes (3%) of dynamic memory, leaving 252756 bytes for local variables. Maximum is 262144 bytes.
pyocd flash -t rp2040 /tmp/arduino_build_286990/Blink.ino.elf 
[---|---|---|---|---|---|---|---|---|----]
[========================================]
0007181:INFO:loader:Erased 73728 bytes (18 sectors), programmed 73728 bytes (288 pages), skipped 0 bytes (0 pages) at 12.88 kB/s
```

Submitted as draft PR since I don't know
- How to pack pyocd exe for different platforms, currently simply use pyocd installed in system with pip
- not sure how to use picoprobe.tcl

PS: I think the picoprobe (openOCD) should also change the driver from "interface/picoprobe.cfg" to "interface/cmsis-dap.cfg" to work with latest picoprobe changes. However, a quick try does not seem to work. Maybe the openocd binaries need to be rebuilt

```
Open On-Chip Debugger 0.10.0+dev-ge3428fadb-dirty (2022-07-14-14:57)
Licensed under GNU GPL v2
For bug reports, read
	http://openocd.org/doc/doxygen/bugs.html
Info : auto-selecting first available session transport "swd". To override use 'transport select <transport>'.
Info : Hardware thread awareness created
Info : Hardware thread awareness created
Info : RP2040 Flash Bank Command
Error: unable to find CMSIS-DAP device
Error: No Valid JTAG Interface Configured.
```

From latest getting started with pico

Screenshot from 2023-02-14 10-43-58


earlephilhower:
I have never before heard of pyocd, and looking at the GH repo it looks like it needs dozens of Python imports which makes it almost impossible to package in the way we've packaged other tools.  Is it really necessary?  Doesn't OpenOCD cover the needs (we added CMSIS-DAP with the last toolchain build, in addition to  PicoProbe)?

Does this mean that now they have folks with old PicoProbe boards that need new firmware, as well?

Basically the process to get something like this in would be to add the python sources to the repo and then include all the python dependency packages as well (like we have pyserial already).  A wrapper python script would then add the proper include paths and call the main pyocd.py file.  It's probably doable with a lot of work, but I really hope it's not needed...

hathach:
> I have never before heard of pyocd, and looking at the GH repo it looks like it needs dozens of Python imports which makes it almost impossible to package in the way we've packaged other tools. Is it really necessary? Doesn't OpenOCD cover the needs (we added CMSIS-DAP with the last toolchain build, in addition to PicoProbe)?

pyocd is python OCD, which is nice and easier to work comparing to openOCD, IMHO. However, If it is too much of trouble (I agree python doesn't like to be packed into exe), we could skip it. Upgrading the openocd binary + using cmsis-dap.cfg should cover our need

> Does this mean that now they have folks with old PicoProbe boards that need new firmware, as well?

Yeah, I think so, we can have 2 separate picoprobes, Picoprobe (cmsis-dap), Picoprobe (old firmware) (or something similar) with different cfg file.

> Basically the process to get something like this in would be to add the python sources to the repo and then include all the python dependency packages as well (like we have pyserial already). A wrapper python script would then add the proper include paths and call the main pyocd.py file. It's probably doable with a lot of work, but I really hope it's not needed...

never mind the pyocd as long as we update the openocd binaries and add support for new cmsis-dap.cfg. That would cover the new picoprobe firmware

earlephilhower:
Re: openocd and CMSIS, the latest release should have this already:
https://github.com/earlephilhower/pico-quick-toolchain/blob/0a9de22221b24d0e94869a6213ff1ce1f0400eb4/Makefile#L238-L243

I don't have a TCL script in `/lib` for CMSIS, though, but the actual support files should be installed already by the release and be able to be used from the command line directly.  Have you given it a try?

(FWIW, the OpenOCD shipped is the RPi Pico fork, picoprobe branch, since nothing else supports the Picoprobe)

hathach:
> Re: openocd and CMSIS, the latest release should have this already: https://github.com/earlephilhower/pico-quick-toolchain/blob/0a9de22221b24d0e94869a6213ff1ce1f0400eb4/Makefile#L238-L243

> I don't have a TCL script in `/lib` for CMSIS, though, but the actual support files should be installed already by the release and be able to be used from the command line directly. Have you given it a try?

I did quick try but it doesn't work when picking the "interface/cmsis-dap.cfg" (editing the platforms.txt directly). I haven't looked any closer on why it failed.

> (FWIW, the OpenOCD shipped is the RPi Pico fork, picoprobe branch, since nothing else supports the Picoprobe)

maybe it is good chance to update to 'rp2040' branch, which seems to be the dev branch now (which also incldue the picoprobe branch)  https://github.com/raspberrypi/openocd/commits/rp2040

earlephilhower:
> maybe it is good chance to update to 'rp2040' branch, which seems to be the dev branch now (which also incldue the picoprobe branch) https://github.com/raspberrypi/openocd/commits/rp2040

Sorry, that was a typo.  I am building from `rp2040` as of June 16, 2022:
````
earle@amd:~/src/pico-quick-toolchain/repo/openocd$ git branch 
* rp2040
earle@amd:~/src/pico-quick-toolchain/repo/openocd$ git log | head -20
commit 228ede43db3665e470d2e518730de013a8c74411
Author: Liam Fraser <liam@raspberrypi.com>
Date:   Thu Jun 16 08:58:28 2022 +0100

    Use new submodule URLs from upstream openocd. repo.or.cz was down!

commit 610f137d2b91b794dc9a8999b1e6631a99fcddc7
Author: Rene Kita <git@rkta.de>
Date:   Fri Mar 12 18:38:02 2021 +0100

    build: remove warnings with gcc 11
    
    This removes some warnings which prevent a successful build with -Werror
    which is enabled by default. I'm using gcc 11, so maybe others are not
    getting this warnings yet.
    
    In src/flash/nor/numicro.c the debug messages were misleadingly indented.
    In src/target/arm920t.c the array size where smaller than expected from
    the receiving function.
````

**Comment by earlephilhower** [comment]
The commit log on GH for the repo shows no significant changes from then, only a bugfix for something called "clion":  https://github.com/raspberrypi/openocd/commits/rp2040

**Comment by hathach** [comment]
ah, you are right, the openocd is up-to-date, I forgot to run the `tools/get.py` to get latest. Though somehow I still have issue using with latest picoprobe though. I have no idea why it failed with `CMD_DAP_SWJ_CLOCK`, it could be caused by my pc setup. Let close this pyocd for now. I will do a more clean setup later and make an PR to add picoprobe (cmsis-dap) upload mode after testing it out. Thank you for fast response.

```
/home/hathach/Arduino/hardware/rp2040/rp2040/system/openocd/bin/openocd -f interface/cmsis-dap.cfg -f target/rp2040.cfg -s /home/hathach/Arduino/hardware/rp2040/rp2040/system/openocd/share/openocd/scripts -c program {/tmp/arduino_build_434986/Blink.ino.elf} verify reset exit 

Open On-Chip Debugger 0.11.0-g228ede43d-dirty (2022-12-23-00:38)
Licensed under GNU GPL v2
For bug reports, read
	http://openocd.org/doc/doxygen/bugs.html
Info : auto-selecting first available session transport "swd". To override use 'transport select <transport>'.
Info : Hardware thread awareness created
Info : Hardware thread awareness created
Info : RP2040 Flash Bank Command
Info : Using CMSIS-DAPv2 interface with VID:PID=0x2e8a:0x000c, serial=E660C06213513538
Info : CMSIS-DAP: SWD  Supported
Info : CMSIS-DAP: FW Version = 2.0.0
Info : CMSIS-DAP: Interface Initialised (SWD)
Info : SWCLK/TCK = 0 SWDIO/TMS = 0 TDI = 0 TDO = 0 nTRST = 0 nRESET = 0
Error: CMSIS-DAP command CMD_DAP_SWJ_CLOCK failed.
Error: No Valid JTAG Interface Configured.
Error: No Valid JTAG Interface Configured.
```

PR #1199, opened 2023-02-17
Enable use of uninitialized_ram() macro

### LINKED ISSUE #1188: Move uninitialized_data segment to after heap end
See #1186

earlephilhower:
The uninitted RAM macro would fail because the OTA code would clear out the first 50KB or so of RAM to copy its code and global values.

Move all global values to the end of RAM in OTA, and then align any uninitted vars to a 16KB unit to ensure it won't appear in the 1st part of RAM that is still needed to copy the OTA executable.

In the normal case, without any uninit_ram vars, no additional space is used. When uninit_ram is used, up to 16KB of space may be lost to get that alignment, but it will work properly.

Fixes #1188


earlephilhower:
Tested with the simple sketch:
````
/*static*/ uint32_t __uninitialized_ram(spacecamp)[2];

void setup() {
  delay(5000);
  Serial.printf("%08x %d\n", spacecamp[0], spacecamp[1]);
  if (spacecamp[0] != 0xc0debead) {
    spacecamp[0] = 0xc0debead;
    spacecamp[1] = 0;
  }
  delay(1000);
  spacecamp[1]++;
  rp2040.reboot();
}

void loop() { }
````

PR #1208, opened 2023-02-19
Move PicoW auto-reassignment of LED pin to code

earlephilhower:
Many examples require that LED_BUILTIN be defined as a constant, so we can't use a ternary operator to swap between Pico and PicoW LED pins.

Instead, do the check in the digitalWrite/digitalRead/pinMode calls to cover most of the uses.

### DIFF
--- a/variants/rpipicow/picow_digital.cpp
+++ b/variants/rpipicow/picow_digital.cpp
@@ -24,7 +24,12 @@ extern "C" void __pinMode(pin_size_t pin, PinMode mode);
 extern "C" void __digitalWrite(pin_size_t pin, PinStatus val);
 extern "C" PinStatus __digitalRead(pin_size_t pin);
 
+extern bool  __isPicoW;
+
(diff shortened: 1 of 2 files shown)


earlephilhower:
@oddstr13 would you be able to do some add'l testing and report back, since you found it on your code?  I don't mind doing a hotfix, but I'd rather not have to do multiple hotfixes in a row. :)

oddstr13:
I do like magic that makes things easier to deal with, but I can't help but feel that this is a bit too much, and that it has a potential to be hard to debug down the line.

Personally I think I'd expect having to manage the pin numbers myself, if writing code that should compile and be binary compatable on several dissimilar boards.

I'm not sure of how much the extra overhead added for this magic matters for the Pico, but it is certainly not something I'd want in a avr chip (but, I'd be poking registers directly there when speed matters, so..).

oddstr13:
Still not a huge fan, but fair enough

PR #1210, opened 2023-02-20
Update PlayStereo.ino

crp500:
The PWMAudio lib is expecting an int16_t value but the example passes a sample cast to uint16_t  So any values from the lookup table that are negative are recast to 0 i think. So half the sine wave in the case of the example is lost.


earlephilhower (approved):
Thanks!

PR #1219, opened 2023-02-22
Add timeout to WiFiClient constructor

Beirdo:
- The client has full support for timeouts, but no apparent way to set the timeout to anything but the default.
- This change keeps the same default in place (5000ms), but allows the user to override it.
- This can make it so reconnecting to Wifi-based TCP sockets can happen much quicker, which is crucial if you are trying to reconnect from your main loop, particularly when both cores are running in sub-second timed schedule loops.
- This will also make it more possible to employ the hardware WDT with a reasonable timeout without the board rebooting when a connection takes too long as there is no "yield to kick the watchdog" code in place.

### DIFF
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -59,9 +59,9 @@ template<>
 WiFiClient* SList<WiFiClient>::_s_first = 0;
 
 
-WiFiClient::WiFiClient()
+WiFiClient::WiFiClient(int timeout)
     : _client(0), _owned(0) {
-    _timeout = 5000;
+    _timeout = timeout;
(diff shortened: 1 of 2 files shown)


earlephilhower:
@Beirdo I've not gone thru this in detail, but don't we already support `WiFiClient::setTimeout(xxx)` via the `Stream` inheritance?

Beirdo:
Hmmm, I couldn't find that, but let me give it a shot.  If so, we can just abandon this, although it still irks me to see a hardcoded 5s timeout ;)

Beirdo:
Yes, it seems we do have that.  Well, let's ditch this.  Good catch.  I searched for it for quite a bit there, but I guess my mind is having a second Monday. :)

Issue #1206, opened 2023-02-19
Error expanding macro for PIN_LED: initializer element is not constant


oddstr13:
Regression caused by #1204

```
In file included from /home/oddstr13/Arduino/hardware/pico/rp2040/cores/rp2040/Arduino.h:30,
                 from /tmp/arduino_build_702994/sketch/constants.c:1:
/home/oddstr13/Arduino/hardware/pico/rp2040/variants/rpipicow/pins_arduino.h:9:24: error: initializer element is not constant
    9 | #define PIN_LED        (__isPicoW ? 32u : 25u)
      |                        ^
/home/oddstr13/Arduino/hardware/pico/rp2040/variants/rpipicow/../generic/common.h:9:21: note: in expansion of macro 'PIN_LED'
    9 | #define LED_BUILTIN PIN_LED
      |                     ^~~~~~~
/tmp/arduino_build_702994/sketch/constants.c:5:11: note: in expansion of macro 'LED_BUILTIN'
    5 | int led = LED_BUILTIN;
```

Target board: Raspberry Pi Pico W

earlephilhower:
Fixed by #1208.  CI didn't catch it because that use case (setting a global) wasn't in any of the WiFi-related sketches.  I've manually tested it w/Blink and BlinkWithoutDelay which does properly work on both Pico/PicoW (and uses a global int set to `LED_BUILTIN`)

oddstr13:
I've made a PR for a minimal reproduction, do with it as you wish.

I'm personally not entirely sure of all the implications of handling all of this at runtime as implemented in #1208, I think it has the potential to cause hard to track down bugs (the `PIN_LED 32 -> 25` mapping may not hold true on all boards, Pin 32 may have other meanings on some non-wifi boards).

Might not be an issue with current boards, but it makes me feel a bit uneasy.

PR #1222, opened 2023-02-22
Add reproduction case for #1206

oddstr13:
It was necessary to place the macro expansion in a separate .c file to trigger #1206 (just a .h was not enough)

### DIFF
--- a/libraries/rp2040/examples/Blink_1206/Blink_1206.ino
+++ b/libraries/rp2040/examples/Blink_1206/Blink_1206.ino
@@ -1,0 +1,13 @@
+#include "blink_config.h"
+// This example is intended to be a minimal reproduction test case for earlephilhower/arduino-pico#1206
+
+void setup() {
+  pinMode(led, OUTPUT);
+}
+
+void loop() {
+  digitalWrite(led, LOW);
+  delay(blink_time);
+  digitalWrite(led, HIGH);
+  delay(blink_time);
+}
--- a/libraries/rp2040/examples/Blink_1206/blink_config.c
+++ b/libraries/rp2040/examples/Blink_1206/blink_config.c
@@ -1,0 +1,7 @@
+#include <Arduino.h>
+#include "blink_config.h"
+
+// These variables are stored in a separate file in order to trigger earlephilhower/arduino-pico#1206
+
+int led = LED_BUILTIN;
+long blink_time = 500;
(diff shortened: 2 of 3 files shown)


earlephilhower:
Thanks for adding a test for this, but could we not just modify an existing test to use a global `int led=LED_BUILTIN` and `s/LED_BUILTIN/led/g`?

oddstr13:
If placed into, and imported from a `.c` via a `.h`, I wasn't able to trigger it directly in the `.ino` file, but that *may* have been a caching issue on my side? Placing the `led=` in the header file didn't trigger the bug for sure.

earlephilhower:
Should this not fail testing until a change is merged to master?  The CI ran to completion here w/o incident?

oddstr13:
Yes, I was expecting a failure too, been looking into it a little, but I haven't found out why yet.

oddstr13:
Looks like it passes because the tests get ran against the generic board.

https://github.com/earlephilhower/arduino-pico/actions/runs/4245900238/jobs/7382019361#step:5:318

It passes locally for me too when selecting `Generic RP2040` or `Raspberry Pi Pico`. Fails on `Raspberry Pi Pico W`.

Looks like it could be worth running tests with more permutations, such as with/without WiFi, WizNet and probably also IPv6.
(I had failure for IPv6 a while back, but didn't need it on this particular network, so didn't look more into it and just switched to the IPv4 only stack).

earlephilhower:
The problem is the number of runs involved becomes onerous.  While I thank you for going the extra step to make a unit test for this, I'm not quite sure we want to ship unit tests as examples, and at this point (with BT coming up in a matter of days) I don't want to start multiplying the CI.

The problem w/#1206 was actually a couple things.
1. You had a complete compile failure because of (I believe) a difference in global initialization w/C vs. C++.  This was actually the best possible outcome
2. Even if the (@ global level) `int led = LED_BUILTIN;` compiles in C++ as it seems to do w/other WiFi examples, the value of `__isPico` is **undefined** when the code setting `int led` to some value.  This is an even worse outcome, being a logic bug which slips by compilation and warnings.

The fix returns `LED_BUILTIN` to a constant, and only silently changes it inside a variant-local set of digitalWrite/Read/pinMode.  So for now I think we can skip this add'l test and just keep an eye out for anything similar in incoming board definitions...

oddstr13:
I agree, unit tests should be placed in a separate directory, and include a set of different configurations to test with, and the expected outcome of those.

PR #1225, opened 2023-02-25
Make copy to RAM build configuration for Raspberry Pi PICO

daveythacher:
Attempt to create copy to RAM board configuration for Raspberry Pi PICO. Is this on the right track?

Notes:
Not sure if linker script is correct. (No OTA, no binary_info, etc.)
Only applied changes in python script for Raspberry Pi PICO.
Changes in python script are hackish.
Do not know if I changed the correct configuration lines.
Did not produce corresponding intermediate files.
No testing performed.


daveythacher:
I did not realize this (#1234) was practiced in the project. Nevermind!!!

PR #1242, opened 2023-03-02
Add Adafruit Feather RP2040 DVI

PaintYourDragon:
Forthcoming board; not out yet. Hardware is essentially the same as RP2040 SCORPIO, but with HDMI connector in place of dual-row header. Different PID.

Some video resolutions require extreme overclocking, such that the CPU outpaces flash at the default QSPI div2 clock rate, so a rollover menu here allows selecting div2 or div4. Not just for the new DVI Feather, this is present for any Adafruit RP2040 board w/W25Q080 flash, as some other might be used with DVI breakouts or have other reasons for wanting extreme overclocking (e.g. signal acquisition w/SCORPIO).


earlephilhower:
You'll need to rebase off of the latest release, but OTW seems fine, thanks!

maxgerhardt:
Also the makeboards.py was not executed or at least the PlatformIO JSON definition was not generated.

Issue #1216, opened 2023-02-21
Feather RP2040 Scorpio pins not working as they should


whimsee:
Hi! I just migrated my Feather RP2040 project to its SCORPIO variant and I'm running into some trouble with the I2C and SPI pins not being where they should be since my OLED wasn't connecting on the I2C, as does the SD card using SPI.

I did some digging and comparing with the pinout sheet and pins_arduino.h file for both boards and it looks like the pins aren't lining up right with the bus.

To check: I did go ahead and change the I2C pins from--

SCORPIO:
#define PIN_WIRE0_SDA  (2u)
#define PIN_WIRE0_SCL  (3u)
#define PIN_WIRE1_SDA  (31u)
#define PIN_WIRE1_SCL  (31u)

to

BASE FEATHER RP2040:
#define PIN_WIRE0_SDA  (24u)
#define PIN_WIRE0_SCL  (25u)
#define PIN_WIRE1_SDA  (2u)
#define PIN_WIRE1_SCL  (3u)

And it worked for the OLED (though looking at it now, SDA and SCL should be 16 and 17 respectively). But I'm kind of stuck with the SPI not working for the SD card since it's looking at a different bus (ie. the Base Feather RP2040's SCK0 is now SCK1 in SCORPIO)

earlephilhower:
The Scorpio variant was built by ladyada and team at Adafruit:  https://github.com/earlephilhower/arduino-pico/commit/913dfad1ad3ef48f7270f26e2351ede7f8bc09de .   If there's an issue with the pinout, can you make a PR to fix it?

As for the SPI busses, just use this `SD::begin` to tell it to use SPI1 not SPI0:
https://github.com/earlephilhower/arduino-pico/blob/a97d5eab22ea0edecd01193ada120733ea898e75/libraries/SD/src/SD.h#L38

whimsee:
Thanks for the reply! Since I was using the SdFat library with SPI, I did end up using its SoftwareSPI class instead, which worked nicely with some fiddling. Though I'm grateful for that solution as well in case anyone else runs into this issue.

And sure, I wasn't certain if I could submit a PR because I've never done it before but I'd be glad to. I'm fairly positive about the I2C pins now but I'm still not 100% about the SPI pins; more like I know where SCK should be for both buses, but not so much about MOSI, MISO, and SS.

earlephilhower:
[...]

PR #1255, opened 2023-03-03
Reassign Feather Scorpio I2C pins to correct buses

whimsee:
Referencing #1216 and basing it off these pinouts (https://learn.adafruit.com/introducing-feather-rp2040-scorpio/pinouts) as this board uses Wire1 for the "main" broken-out I2C pins, the stemma port included. Can confirm that's the case for mounted featherwings that use this board's I2C.

Basing Wire0's pins from the base RP2040 pindefs, which match with this board's as well. (Base RP2040 pinout: https://learn.adafruit.com/adafruit-feather-rp2040-pico/pinouts)

Sidenote: This also applies to the SPI pins (should be on SPI1, but otherwise SCK, MOSI, and MISO pins check out) but can't fully confirm the SS pin and it's originally commented as "Not pinned out."


earlephilhower (approved):
Thanks.  Yes, it seems Wire0 and 1 got mixed up in the initial board file.

PR #1258, opened 2023-03-04
Add board definition for Neko Systems BL2040 Mini

nanoparticle:
Hi,

I have designed an open source brushless servo motor driver board (intended for use with the [SimpleFOC](https://github.com/simplefoc/Arduino-FOC) library), and would appreciate the opportunity to add it to the boards recognized by this Arduino core. More details are available [here](https://oshwlab.com/nekosystems/bl2040-mini-alpha-v3), and documentation is currently in the process of being written.  

While adding the board, I had a couple of questions and concerns that weren't immediately answerable by the internet:
1. For the USBD_MAX_POWER_MA value, what exactly does this do? My board allows the user to run a motor when powered just by USB, and is capable of drawing up to 1000mA (limited by a protection IC), since many USB ports can source that much current despite it exceeding the USB spec. A website I found implied that 500 is the maximum USBPWR value, but there didn't seem to be a clear delineation.
2. I couldn't find a good explanation of what exactly the boot2 files do, so I just picked the one that immediately worked when I used this core configured for a generic rp2040 device. I'm using the W25Q32JVSSIQ flash chip, and would be interested in finding out which of the boot2 files utilize it to its fullest potential.

Thanks!


nanoparticle:
A couple more questions occurred to me:
1. Is the pins_arduino.h file a good place to store some useful constant definitions for the board? The effective equivalent resistance/gain of the current sense amplifiers comes to mind as an example of a helpful value to have on hand.
2. What is the best way to test the core (including my changes) on the boards in order to verify functionality?

earlephilhower:
>  For the USBD_MAX_POWER_MA value, what exactly does this do? My board allows the user to run a motor when powered just by USB, and is capable of drawing up to 1000mA (limited by a protection IC), since many USB ports can source that much current despite it exceeding the USB spec. A website I found implied that 500 is the maximum USBPWR value, but there didn't seem to be a clear delineation.

It's what the USB descriptor will contain.  It's been a while since I looked at the guts of it, but IIRC the USB hub can choose to enable/disable ports to fit within its power profile.  Don't set above 500mA as that's the spec limit would be my suggestion.

> I couldn't find a good explanation of what exactly the boot2 files do, so I just picked the one that immediately worked when I used this core configured for a generic rp2040 device. I'm using the W25Q32JVSSIQ flash chip, and would be interested in finding out which of the boot2 files utilize it to its fullest potential.

They configure the QSPI port.  Your flash chip vendor/model and PCB wiring will determine the right one.  If you get the wrong one it may work in most conditions, but there may be corners of V and temp where it doesn't.

> Is the pins_arduino.h file a good place to store some useful constant definitions for the board? The effective equivalent resistance/gain of the current sense amplifiers comes to mind as an example of a helpful value to have on hand.

Yes, that's the spot.  Add your own defines there if needed.  The core won't use it, but your apps can.

> What is the best way to test the core (including my changes) on the boards in order to verify functionality?

Try something flash-heavy like the LittleFS SpeedTest, and a few of the other non-trivial examples would be my suggestion.

nanoparticle:
I'm having a hard time installing the core from my branch in order to test it on hardware.

Steps taken:
1. Follow this guide to install the core in my Arduino folder
    a. git clone my fork
    b. checkout the add-bl2040-board branch
    c. run `git submodule update --init --recursive` to install the ArduinoCore-API as well as some other necessary libraries like lwip
2. Adjust my pins_arduino.h as needed
3. Run Verify step (works fine)
4. Run Upload step (fails, complains about being unable to find /system/python3/python3 in order to execute upload script)

Is there some other step I need to do to help the Arduino IDE find python or recognize the core better? I checked in the arduino-pico core installed by the board manager and the system folder is also empty, so I have no idea how python is found when the released core uploads code.

earlephilhower:
I think you missed the step documented in the readme:
````
cd ../tools
python3 ./get.py
````

That actually pulls down the toolchain and installs it for GIT use...

earlephilhower:
@nanoparticle I see you've been merging w/Master.  Is this still a draft?  If not, please update the PR (the "Ready for Review" button).

nanoparticle:
Sorry about that, I got sidetracked when a pcb design error came up while I was testing the core.
One last question: Is it possible to omit any of the definitions on [this line](https://github.com/earlephilhower/arduino-pico/blob/6f8b3307cc5323fd0ff47ccc6e3514948aa7fbc7/variants/nekosystems_bl2040_mini/pins_arduino.h#L44) and below? I had to define all of them to allow the core to instantiate the default peripheral drivers, but there aren't enough free pins left to give each one a unique unused pin. Currently, some of the defines are set to pins used for motor control functions, but I worry that someone will try using SPI2 or something without changing the pins and end up frying the board.

nanoparticle:
Alright, I think the board definition is ready. Thanks so much for all the help!

earlephilhower (approved):
LGTM!

PR #1259, opened 2023-03-04
Add bluetooth support to Platform.io

sanjay900:
I went for a similar definition to the IPv6 one, `PIO_FRAMEWORK_ARDUINO_ENABLE_BLUETOOTH`


earlephilhower (changes requested):
Thanks!  One thing is missing, though.   We need to add the proper `#defines`,  `ENABLE_CLASSIC=1` and `ENABLE_BLE=1` when using the BT version of the lib.  Can you add them here, please?

earlephilhower (approved):

Issue #1232, opened 2023-02-27
I2S input example doesn't work with common MEMS microphones


timboldt:
I was experimenting with a common I2S MEMS microphone (https://www.adafruit.com/product/3421) and it didn't work.

In retrospect, I think I found the problem. The datasheet (https://www.mouser.com/datasheet/2/218/ph0645lm4h_b_datasheet_rev_c-1525723.pdf) says that the SPH0645 needs a minimum clock rate of 1MHz in order to wake the device.

However, the example (https://github.com/earlephilhower/arduino-pico/blob/master/libraries/I2S/examples/I2SInput/I2SInput.ino) has 22050Hz * 16 bits * 2 channels = 705600Hz = 0.7MHz.

If you are interested, I can update the example to work with a modern I2S MEMS microphone, and send you a PR.

earlephilhower:
Thanks, but the specs of the mikes really do differ.  The one I built on and documented in the code is from the GOOG AI-Yourself pack that's got a nice PCB w/stereo MEMS chips with a nice separation between them, and they happily work at 22Khz, for example.

Maybe you'd like to do a PR adding some comments in the example like, "for the Adafruit mike, or if your mike doesn't seem to be reporting data, use XXKhz"?  That way we cover all bases.  The current comments spec a particular mike, and your add'l ones will cover some other options?

timboldt:
You proposal sound reasonable.

I'm still having trouble getting the I2S data line to respond at all with an RP2040, so I'll do some digging with a Saleae and compare it against a different known-working microcontroller.

timboldt:
It produces data with a Feather M0 and their example program at https://learn.adafruit.com/adafruit-i2s-mems-microphone-breakout/arduino-wiring-and-test:

![image](https://user-images.githubusercontent.com/903212/222935113-14d9949c-9714-4ec4-b0cb-774277f46637.png)

I'll repeat the exercise with an RP2040.

timboldt:
I was able to get the Pico to cooperate finally:
![image](https://user-images.githubusercontent.com/903212/222937450-0b885e77-75ca-4ce2-9951-6cd1798ff5f8.png)

Next, I'll send a PR, adding some comments for how to configure an Adafruit I2S MEMS microphone to work with the example.

timboldt:
[...]

PR #1272, opened 2023-03-05
Add instructions for making I2S input work with Adafruit microphone

timboldt:
As described in #1232, many modern MEMS microphones, like the Adafruit I2S MEMS Microphone (https://www.adafruit.com/product/3421) require a minimum clock rate that is higher than the original example provided.


earlephilhower (approved):
Thanks!

PR #1292, opened 2023-03-10
Fix pin assigment for Waveshare RP2040 1.28 LCD PIN_BAT_ADC

PhilippMolitor:
Tested and verified on my hardware.


earlephilhower (approved):
Thanks!

PR #1296, opened 2023-03-11
add supports for VCC_GND 4MB and 8MB boards

e-tinkers:
add supports for VCC_GND 4MB (YD-RP2040-4N) and 8MB (YD_RP2040-8N) boards, rename original VCC_GND board to 16MB.


earlephilhower (changes requested):
Thanks very much for bringing this up, but I think adding add'l boards just for a different flash size is not the best way to do it.  (We already have a lot of boards in the popup menu!)

In #1297 I've added a Flash Size menu, just like the Generic RP2040, so that you can select between 4, 8, or 16MB boards easily w/o needing a whole new board.

Thx again, though, for the effort!  It's always nice to get PRs rather than bug reports. :laughing:

e-tinkers:
Thanks, and in fact I like this approach for selecting the Flash Size than having 3 boards listing. 

BTW, I noticed that by adding the options for 4,8 and 16MB for the VCC-GND board, you removed the 2MB option that was originally available for the Generic board. 2MB for sure not valid option for VCC-GND board but it could be a valid option for Generic board.

earlephilhower:
I didn't remove the 2MB option, I think the `diff` threw you because `git` sometimes chooses odd comparison lines.

My menus show it here:
image

PR #1301, opened 2023-03-12
Feather DVI peripheral fix

ladyada:
Fix SPI and Wire so even tho we are using the 'second' peripherals, they are usable as `SPI` and `Wire` :)

### DIFF
--- a/variants/adafruit_feather_dvi/pins_arduino.h
+++ b/variants/adafruit_feather_dvi/pins_arduino.h
@@ -37,8 +37,11 @@
 // Wire
 #define PIN_WIRE0_SDA  (2u)
 #define PIN_WIRE0_SCL  (3u)
+#define __WIRE0_DEVICE i2c1
+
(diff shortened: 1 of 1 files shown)


ladyada:
cc @PaintYourDragon

maxgerhardt:
`__SPI0_DEVICE` does not appear in this repo nor in Pico-SDK when doing a Github search, does it affect a library?

PaintYourDragon:
@ladyada Since pins_arduino.h is getting updated here…might be worth adding the DVI CKN/CKP/D0N/D0P etc. pin defs that were desired but overlooked when the board was first added.
ALSO, if the proposed Wire/SPI changes here stick…should the same change be retconned into Feather RP2040 SCORPIO, or is this applicable only to the new “bones” boards?

ladyada:
@maxgerhardt its in the SPI and Wire library files, determines which peripheral is used

PR #1321, opened 2023-03-20
Add Win32 Long Path info to P.IO docs

earlephilhower:
@maxgerhardt Can you verify the new steps?  Seems like this is biting lots of folks now.


maxgerhardt:
Thanks, the docs are absolutely right.

PR #1333, opened 2023-03-28
Add Feather RP2040 RFM, fix pin defs for DVI & SCORPIO

PaintYourDragon:
Former covers several forthcoming Feather boards, only difference among them is different transceiver modules (same pinout on all of them, and same USB PID).
Latter adds or changes some pin names and TWI instance to use for the STEMMA connector (DVI board had this right, SCORPIO did not, both in sync now).


earlephilhower (approved):
Thx!

PR #1334, opened 2023-03-29
Two more incoming Feathers

PaintYourDragon:
ThinkINK (e-paper display) and USB Host.


earlephilhower:
Neat!  You guys are coming out with a ton of cool boards.

Is this PR ready to merge or are you expecting more changes/additions?

PaintYourDragon:
I think that’s it for this round (add RFM, ThinkINK & USB host, fix SCORPIO & DVI pins). Might be more in the future but probably not for some weeks.

variants/adafruit_feather_usb_host/pins_arduino.h
```diff
@@ -0,0 +1,56 @@
...
+// NeoPixel
+#define PIN_NEOPIXEL   (21u)
+#define NEOPIXEL_POWER (20u)
+
+// 'Boot0' button also on GPIO #7
+#define PIN_BUTTON     (7u)
+
+// USB host connector
+#define PIN_USB_DP     (16u)
+#define PIN_USB_DN     (17u)
```
hathach, line 18: @PaintYourDragon late to the party, but there is a typo, this should be DP/DM (not DN though). Since we need to make an PR to correct it, do you think we should rename them to PIN_USB_HOST_DP/DM to make it clearer that this is for host. Since I saw some other board has the same naming but for usb device ?

PaintYourDragon: Good idea, I’ll add HOST in the pin names, and DM instead of DN (I was probably following the lead of the DVI Feather, where it’s [N]egative. Boards are not quite shipping to customers yet so the changes shouldn’t affect anyone.

Issue #1382, opened 2023-04-20
Update Adafruit TinyUSB to latest version


TheRacc2:
Adafruit TinyUSB v2.0.3 includes support for multiple HID devices, which is critical for low-latency situations. The earlphilhower core currently supports Adafruit TinyUSB v1.17

Is updating the TinyUSB version possible?

TheRacc2:
Can be fixed by [pull #1383](https://github.com/earlephilhower/arduino-pico/pull/1383)

earlephilhower:
Wow, that was a fast upstream release.  I checked this morning, less than a day later, and there it was.  Hope it's all working for you now!

PR #1383, opened 2023-04-20
Update Adafruit_TinyUSB_Arduino version (Fixes issue #1382)

TheRacc2:
Tested with PlatformIO, Windows 11.

Worked flawlessly on Pi Pico, no issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,6 @@
   "homepage": "https://arduino-pico.readthedocs.io/en/latest/",
   "repository": {
     "type": "git",
-    "url": "https://github.com/TheRacc2/arduino-pico"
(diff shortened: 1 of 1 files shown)


earlephilhower (changes requested):
First of all, thanks for the PR.

I would request we wait a week or two so that the upstream repo can do a tagged release.  It looks like this is just a link to the latest commit(which I guess added the new feature)

Keeping it on a nice release number makes it easier to support if there a problem down the line, and it looks like they do release pretty frequently.

TheRacc2:
Seeing as this is a one-line git command, I'll close the PR and re-open the issue. Should this be changed in the near future, I'll promptly close my issue. Thank you!

PR #1395, opened 2023-04-23
Protect the HW random generation from FreeRTOS

### LINKED ISSUE #1394: PicoW hang during init when using FreeRTOS
### Discussed in https://github.com/earlephilhower/arduino-pico/discussions/1390

<div type='discussions-op-text'>

<sup>Originally posted by **ruifig** April 22, 2023</sup>
I'm experimenting with adding FreeRTOS tasks.
With a minimal application it works fine but with my actual project, it's freezing at startup with this callstack:

```
sleep_until@0x1003774a (\sleep_until.dbgasm:43)
capture_additional_rosc_samples@0x10045778 (Unknown Source:151)
initialise_rand@0x10045822 (Unknown Source:210)
get_rand_64@0x10045822 (\get_rand_64.dbgasm:66)
get_rand_32@0x100459c2 (\get_rand_32.dbgasm:3)
udp_init@0x10042196 (\udp_init.dbgasm:3)
lwip_init@0x1004794c (\lwip_init.dbgasm:7)
lwip_nosys_init@0x10045550 (\lwip_nosys_init.dbgasm:8)
cyw43_arch_init@0x1003d088 (\cyw43_arch_init.dbgasm:14)
cyw43_arch_init_with_country@0x1003ce82 (\cyw43_arch_init_with_country.dbgasm:5)
initVariant@0x100035cc (c:\Users\Rui\.platformio\packages\framework-arduinopico\variants\rpipicow\picow_init.cpp:65)
main@0x10034326 (c:\Users\Rui\.platformio\packages\framework-arduinopico\cores\rp2040\main.cpp:99)
```

Still investigating to see if I can pinpoint why it works with one and fails with another, but any tips are welcome if this rings any bells. :)

The only thing that crosses my mind is that maybe I have some global objects with non-trivial constructors that somehow messes things up before main is called, but don't think that's the case. 

For context, what I'm trying to do is to make use of the rp2040 watchdog in my application, which works fine using Arduino-Pico's `rp2040` object.
But as far I've noticed, there is no way to disable the watchdog once enabled, so if we call into any third-party code or Arduino-Pico's code that takes too long to complete, like for example connecting to Wifi, then there a good chance it will trigger a watchdog reboot.

The workaround I'm trying is to create a very simple FreeRTOS task that allows me to implement a watchdog pause of sorts.
For example, I created a task like this (stays suspended)

```
void autoResetTask(void* pvParameters)
{
	while(true)
	{
		rp2040.wdt_reset();
		delay(250);
	}
}
```

Under normal circumstances, my application calls `rp2040.wtd_reset()` periodically, then when I know I'm calling code that potentially can take longer than 8300ms, I wrap that code with a vTaskResume / vTasSuspend to enable/disable the watchdog auto reset.
This is all working fine with a minimal(ish) sample application I was playing around with, but with my actual application if freezes at startup as explained above. 
I know this won't work with cooperative multitasking (I think Arduino-Pico's FreeRTOS uses cooperative, right?), but I'm not too bothered, since Wifi code does a few delays when connecting.
</div>

earlephilhower:
Fixes #1394

The Pico_Rand SDK calls gather bits from the HW ROSC at precise intervals. If there is jitter in the sleep_until() call then the ROSC bit collection will always think it's failed to acquire the right bit and retry infintitely.

Avoid by wrapping the HW random number calls and the sleep_until() routine. Only when in FreeRTOS set a flag to silently make sleep_until() into a busy wait loop while in a random number generation step.  When not in the random code, do the normal sleep_until call.


earlephilhower:
@ruifig can you give this a try and report back?  The simple MCVE you gave starts up and runs, printing the millis() values every second.

ruifig:
It fixes the freeze, both on my sample application and my real application, but I see a crash that *might* be related.

When calling WiFiMulti::run, very often it crashes when scanning the available networks.
This is the callstack:

Core0:
```
sleep_until@0x100379b6 (\sleep_until.dbgasm:43)
sleep_us@0x10037a10 (\sleep_us.dbgasm:13)
sleep_ms@0x10037a2a (\sleep_ms.dbgasm:8)
rtosFatalError@0x1002e014 (c:\Users\Rui\.platformio\packages\framework-arduinopico\libraries\FreeRTOS\src\variantHooks.cpp:245)
xQueueSemaphoreTake@0x1002f014 (c:\users\rui\.platformio\packages\framework-arduinopico\libraries\freertos\lib\freertos-kernel\queue.c:1486)
xQueueTakeMutexRecursive@0x1002eb14 (c:\users\rui\.platformio\packages\framework-arduinopico\libraries\freertos\lib\freertos-kernel\queue.c:695)
__freertos_recursive_mutex_take@0x1002e1fe (c:\Users\Rui\.platformio\packages\framework-arduinopico\libraries\FreeRTOS\src\variantHooks.cpp:434)
__retarget_lock_acquire_recursive@0x100330a6 (c:\Users\Rui\.platformio\packages\framework-arduinopico\cores\rp2040\lock.cpp:150)
__malloc_lock@0x10048dd0 (\__malloc_lock.dbgasm:4)
_malloc_r@0x100489da (\_malloc_r.dbgasm:25)
malloc@0x10048994 (\malloc.dbgasm:6)
__wrap_malloc@0x100344dc (c:\Users\Rui\.platformio\packages\framework-arduinopico\cores\rp2040\malloc-lock.cpp:30)
operator new@0x1003a2a2 (Unknown Source:14)
__gnu_cxx::new_allocator<std::_Rb_tree_node<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > >::allocate@0x10026e02 (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\ext\new_allocator.h:115)
std::allocator_traits<std::allocator<std::_Rb_tree_node<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > > >::allocate@0x10026cf6 (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\bits\alloc_traits.h:460)
std::_Rb_tree<unsigned long long, std::pair<unsigned long long const, _cyw43_ev_scan_result_t>, std::_Select1st<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> >, std::less<unsigned long long>, std::allocator<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > >::_M_get_node@0x10026972 (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\bits\stl_tree.h:584)
std::_Rb_tree<unsigned long long, std::pair<unsigned long long const, _cyw43_ev_scan_result_t>, std::_Select1st<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> >, std::less<unsigned long long>, std::allocator<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > >::_M_create_node<std::piecewise_construct_t const&, std::tuple<unsigned long long&&>, std::tuple<_cyw43_ev_scan_result_t&> >@0x1002656c (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\bits\stl_tree.h:634)
std::_Rb_tree<unsigned long long, std::pair<unsigned long long const, _cyw43_ev_scan_result_t>, std::_Select1st<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> >, std::less<unsigned long long>, std::allocator<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > >::_M_emplace_hint_unique<std::piecewise_construct_t const&, std::tuple<unsigned long long&&>, std::tuple<_cyw43_ev_scan_result_t&> >@0x1002630c (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\bits\stl_tree.h:2461)
std::map<unsigned long long, _cyw43_ev_scan_result_t, std::less<unsigned long long>, std::allocator<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > >::emplace_hint<std::piecewise_construct_t const&, std::tuple<unsigned long long&&>, std::tuple<_cyw43_ev_scan_result_t&> >@0x10025ffa (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\bits\stl_map.h:608)
std::map<unsigned long long, _cyw43_ev_scan_result_t, std::less<unsigned long long>, std::allocator<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > >::insert_or_assign<_cyw43_ev_scan_result_t&>@0x10025bba (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\bits\stl_map.h:941)
WiFiClass::_scanCB@0x1002540a (c:\Users\Rui\.platformio\packages\framework-arduinopico\libraries\WiFi\src\WiFiClass.cpp:420)
```

Core1:
```
sleep_until@0x100379b6 (\sleep_until.dbgasm:43)
sleep_us@0x10037a10 (\sleep_us.dbgasm:13)
sleep_ms@0x10037a2a (\sleep_ms.dbgasm:8)
rtosFatalError@0x1002e014 (c:\Users\Rui\.platformio\packages\framework-arduinopico\libraries\FreeRTOS\src\variantHooks.cpp:245)
xQueueSemaphoreTake@0x1002f014 (c:\users\rui\.platformio\packages\framework-arduinopico\libraries\freertos\lib\freertos-kernel\queue.c:1486)
xQueueTakeMutexRecursive@0x1002eb14 (c:\users\rui\.platformio\packages\framework-arduinopico\libraries\freertos\lib\freertos-kernel\queue.c:695)
__freertos_recursive_mutex_take@0x1002e1fe (c:\Users\Rui\.platformio\packages\framework-arduinopico\libraries\FreeRTOS\src\variantHooks.cpp:434)
__retarget_lock_acquire_recursive@0x100330a6 (c:\Users\Rui\.platformio\packages\framework-arduinopico\cores\rp2040\lock.cpp:150)
__malloc_lock@0x10048dd0 (\__malloc_lock.dbgasm:4)
_malloc_r@0x100489da (\_malloc_r.dbgasm:25)
malloc@0x10048994 (\malloc.dbgasm:6)
__wrap_malloc@0x100344dc (c:\Users\Rui\.platformio\packages\framework-arduinopico\cores\rp2040\malloc-lock.cpp:30)
operator new@0x1003a2a2 (Unknown Source:14)
__gnu_cxx::new_allocator<std::_Rb_tree_node<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > >::allocate@0x10026e02 (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\ext\new_allocator.h:115)
std::allocator_traits<std::allocator<std::_Rb_tree_node<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > > >::allocate@0x10026cf6 (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\bits\alloc_traits.h:460)
std::_Rb_tree<unsigned long long, std::pair<unsigned long long const, _cyw43_ev_scan_result_t>, std::_Select1st<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> >, std::less<unsigned long long>, std::allocator<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > >::_M_get_node@0x10026972 (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\bits\stl_tree.h:584)
std::_Rb_tree<unsigned long long, std::pair<unsigned long long const, _cyw43_ev_scan_result_t>, std::_Select1st<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> >, std::less<unsigned long long>, std::allocator<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > >::_M_create_node<std::piecewise_construct_t const&, std::tuple<unsigned long long&&>, std::tuple<_cyw43_ev_scan_result_t&> >@0x1002656c (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\bits\stl_tree.h:634)
std::_Rb_tree<unsigned long long, std::pair<unsigned long long const, _cyw43_ev_scan_result_t>, std::_Select1st<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> >, std::less<unsigned long long>, std::allocator<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > >::_M_emplace_hint_unique<std::piecewise_construct_t const&, std::tuple<unsigned long long&&>, std::tuple<_cyw43_ev_scan_result_t&> >@0x1002630c (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\bits\stl_tree.h:2461)
std::map<unsigned long long, _cyw43_ev_scan_result_t, std::less<unsigned long long>, std::allocator<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > >::emplace_hint<std::piecewise_construct_t const&, std::tuple<unsigned long long&&>, std::tuple<_cyw43_ev_scan_result_t&> >@0x10025ffa (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\bits\stl_map.h:608)
std::map<unsigned long long, _cyw43_ev_scan_result_t, std::less<unsigned long long>, std::allocator<std::pair<unsigned long long const, _cyw43_ev_scan_result_t> > >::insert_or_assign<_cyw43_ev_scan_result_t&>@0x10025bba (c:\users\rui\.platformio\packages\toolchain-rp2040-earlephilhower\arm-none-eabi\include\c++\10.3.0\bits\stl_map.h:941)
WiFiClass::_scanCB@0x1002540a (c:\Users\Rui\.platformio\packages\framework-arduinopico\libraries\WiFi\src\WiFiClass.cpp:420)
```

Note the `rtosFatalError` .
Seems like the same callstack on both, so I'm not sure if these callstacks are reliable or not. I find my Pico debugging experience slightly unreliable.

Still, looking at the headers and source code, I see this:
```
int8_t WiFiClass::scanNetworks(bool async) {
    cyw43_wifi_scan_options_t scan_options;
    memset(&scan_options, 0, sizeof(scan_options));
    _scan.clear();
    if (!_wifiHWInitted) {
        _apMode = false;
        cyw43_arch_enable_sta_mode();
        _wifiHWInitted = true;
    }
    int err = cyw43_wifi_scan(&cyw43_state, &scan_options, this, _scanCB);
    if (err) {
        return 0;
    }
    if (!async) {
        uint32_t now = millis();
        while (cyw43_wifi_scan_active(&cyw43_state) && (millis() - now < 10000)) {
            delay(10);
        }
        return _scan.size();
    } else {
        return -1;
    }
}
```

I have zero direct experience with the Pico-SDK and FreeRTOS, but looks like it launches a task to do the scanning?
Again, not sure this is related. Just thought worth pointing it out.

I'm going to try removing my use of FreeRTOS to see if makes any difference.

ruifig:
It does seem like introducing the FreeRTOS.h header messes up Wifi somehow.
With a sample application without FreeRTOS.h , it normally takes maybe 5-10 seconds to connect.
Once I add the FreeRTOS.h, it doesn't connect at all. 

Sorry I don't have anything else to go by. My sample application is this:
Ignore the "Config.h" header. That's just adding some logging function, wifi secrets and what UART to use (MySerial)

```
#include <Arduino.h>
#include "Config.h"
#include "WiFi.h"

#include <FreeRTOS.h>

WiFiMulti gMulti;

void setup()
{
#if CZ_USE_PROBE_SERIAL
	MySerial.setRX(MySerial_RXPin);
	MySerial.setTX(MySerial_TXPin);
#endif
	MySerial.begin(115200);
	logln();
	logln("Hello world!");

	gMulti.addAP(WIFI_SSID, WIFI_PASSWORD);

	auto start = millis();
	logln("Connecting to WiFi...");
	uint8_t res = gMulti.run();
	auto elapsed = millis() - start;
	logln(elapsed, " ms elapsed");
	if (res == WL_CONNECTED)
	{
		logln("Connected to Wifi");
	}
	else
	{
		logln("Failed to connect to Wifi");
	}
}

void loop()
{
	delay(500);
	MySerial.println(millis());
}
```

This is where it gets stuck:
```
[WIFIMULTI] scanList: SSID: 'XXX' -- BSSID: 'XXX' -- RSSI: -58
[WIFIMULTI] scanList: SSID: 'XXX' -- BSSID: 'XXX' -- RSSI: -74
[WIFIMULTI] Connecting to: SSID: 'XXX' -- BSSID: 'XXX' -- RSSI: -58
```

As far as I can tell, it keeps spinning in `WiFiClass::begin`, here:
```
    // Enable CYW43 event debugging (make sure Debug Port is set)
    //cyw43_state.trace_flags = 0xffff;
    while (!_calledESP && ((millis() - start < (uint32_t)2 * _timeout)) && !connected()) {
        delay(10);
    }
```

earlephilhower:
Core1 cannot access WiFi, so that looks to be the problem there.  WiFi + assorted support is only possible on Core0 (`setup`/`loop` or core-pinned tasks).  

Your trace shows
> Core1
````
WiFiClass::_scanCB@0x1002540a (c:\Users\Rui\.platformio\packages\framework-arduinopico\libraries\WiFi\src\WiFiClass.cpp:420)
````

Actually, I'm not sure how you have both core0 and core1 inside `scanCB` now that I look at it.  Is that a cut-n-paste error?

earlephilhower:
If you're using GDB or have a GDB prompt then you can do 
````
thread 0
where
thread 1
where
````

The FreeRTOS debugging stuff does not really support SMP, which is used here, so if you're using something to decode the stacks that **thinks** it knows how to decode FreeRTOS stacks...it's wrong.  (FWIW the SMP support so really new it didn't even support using any C-lib reent calls in parallel so I had to patch that in this branch.  Current upstream did add support, eventually)

ruifig:
Not a cut-n-paste error. They indeed showed the same callstack.
Like I said, I find my debugger experience a bit unreliable, so I'd take the fact they both show the same callstack with a pinch of salt.
One of these days I'll have to start fresh and see if I can improve the workflow. I'm using one of those recently launched Raspberry Pi Debug Probe (pretty much a commercialized PicoProbe).
I also have a J-Link EDU Mini, but haven't tried that in a while because I had problems with using the official Arduino's RP2040 core, which apparently doesn't play well with the J-Link.

Still regardless of showing on both cores or suspecting of having an unreliable debugger, the fact I got that callstack while having problems with Wifi seems too much of a coincidence.

Lets just ignore this for now, and I'll take your tips into consideration to try to pinpoint the problem, eventually. :)
I might end up avoiding the use of FreeRTOS. Not making any progress with the rest of the application for the last few days, so I'll come back to this at a later time.

earlephilhower:
Oh yeah, there probably is an issue w/WiFi and FreeRTOS.  Basically, FreeRTOS takes every assumption the core is built upon and throws it in the trash.  Code can run on different cores, be stopped essentially randomly (on the tick) and swapped, etc.  With the CYW43 rewrite in SDK 1.5 and other changes it looks like there's quite a few spots where interactions are causing trouble.

If you can go w/o FreeRTOS things will be much more stable.

I don't believe the stack thing is debug HW related.  The J-Link or Picoprobe will be essentially the same thing, just an SWD pipe.  The stack problem is related either to OpenOCD (which doesn't grok the SMP method used on the RP2040...there's an open issue from @maxgerhardt about it somewhere) or the higher level IDE logic.  When using plain OpenOCD and GDB things work fine (but it only knows about the current stack, nothing related other suspended tasks in FreeRTOS).

earlephilhower:
I will merge this for now because it does fix a known problem (PicoW doesn't start under FreeRTOS due to the HWRNG).  At some point we'll have a more comprehensive solution.  FreeRTOS is a tricky beast...

earlephilhower:
@ruifig FWIW, the WiFi issue is similar to #1402 and a logically simple issue but tactically involved problem.  Basically the bare metal WiFi driver uses internal mutexes fully inside the SDK.  Those mutexes don't function properly under FreeRTOS' task switching model so it all will blow up, eventually.  The solution involves doing something like the USB fix, or importing the SDK FreeRTOS async processing on-the-fly.  Since it's in the SDK proper it gets a little more complicated than the 3-line fix for the USB mutex. :(

So, in a nutshell, stay away from FreeRTOS if you need BT or WiFi for now.

PR #1416, opened 2023-05-01
Add PICO_FLASH_SIZE_BYTES to PIO

maxgerhardt:
Complementary to #1414 for PlatformIO's defines.

Tested with 
```ini
board = waveshare_rp2040_plus_16mb
board_build.filesystem_size = 1m
```
and it gives indeed the full 
```
-DPICO_FLASH_SIZE_BYTES=16777216
```
which is exactly `16*1024*1024`.


earlephilhower (approved):
Sorry, slipped my mind to update P.IO, too.  Thx!

maxgerhardt:
No problem that's what I'm here for, the burden of PlatformIO support shouldn't be primarily on you.

PR #1447, opened 2023-05-15
Add Arduino CLI installation instructions

alranel:
This PR adds instructions to install this core using Arduino CLI.


earlephilhower (approved):
Thanks!

PR #1451, opened 2023-05-16
setInverted()/getInverted()

myklemykle:
Thanks again for the guidance! 

I see now there are some whitespace issues. Do you have a preferred linter I can run on this?


earlephilhower:
Can you run the `./tests/restyle.sh` script (astyle) if there are spacing issues?  I'll start the jobs anyway to see if anything else gets caught.

cores/rp2040/SerialPIO.h
```diff
@@ -41,6 +41,9 @@ class SerialPIO : public HardwareSerial {
     void begin(unsigned long baud, uint16_t config) override;
     void end() override;
 
+		void setInverted(bool i);
```
earlephilhower, line 44: Can we make this `setInverted(bool invTx = true, invRx = true)`?  Even if it's not implemented I can throw RX together by inspection quickly and have a good shot at it working.

cores/rp2040/SerialPIO.cpp
```diff
@@ -46,12 +46,17 @@ static pio_program_t *pio_make_uart_prog(int repl, const pio_program_t *pg) {
...
-    if (f == _txMap.end()) {
-        pio_program_t * p = pio_make_uart_prog(bits, &pio_tx_program);
-        _txMap.insert({bits, new PIOProgram(p)});
-        f = _txMap.find(bits);
+static PIOProgram *_getTxProgram(int bits, bool inverted) {
+		int key = bits * (inverted ? -1 : 1);
+    auto f = _txMap.find(key);
+    if (f == _txMap.end()) { 
+				pio_program_t * p;
+				if (inverted)
```
earlephilhower, line 54: Not a fan of ternarys?  The `astyle` will blow this up to 6 lines or so because it requires `{}`s around conditions.  I've been burned in past lives before w/o them, so it's part of the core style rules now.
One-liner:  `pio_make_uart_prog(bits, _inverted ? &pio_tx_inv_program : pio_tx_program);`

cores/rp2040/SerialPIO.cpp
```diff
@@ -364,7 +377,10 @@ size_t SerialPIO::write(uint8_t c) {
     }
     val <<= 1;  // Start bit = low
 
-    pio_sm_put_blocking(_txPIO, _txSM, val);
+		if (_inverted)
```
earlephilhower, line 380: Same ternary comment as before

earlephilhower (changes requested):
Looking good once astyle and a couple minor things.  Thx!

cores/rp2040/SerialPIO.cpp
```diff
@@ -346,6 +351,14 @@ void SerialPIO::flush() {
     delay((1000 * (_txBits + 1)) / _baud);
 }
 
+void SerialPIO::setInverted(bool i){
+		_inverted = i;
+}
+
+bool SerialPIO::getInverted(){
```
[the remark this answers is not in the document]
myklemykle: At some point in my life I decided I shouldn't write setters without matching getters. One hopes the compiler will strip it out if it's not used. In this case I was using it for a flip-state test command, where otherwise I have to duplicate the boolean state externally to know it. 
It would be even simpler to just scope the _inverted property public. 
But whatever meets your design goals, I'm happy to do.  I don't need that getter any more, will remove.

cores/rp2040/SerialPIO.h
myklemykle: good idea. I'll have it ignore rx for now.

cores/rp2040/SerialPIO.cpp
myklemykle: done

myklemykle:
> Can you run the `./tests/restyle.sh` script (astyle) if there are spacing issues? I'll start the jobs anyway to see if anything else gets caught.
> 
> (edit: generally it's just 4 space, not tab, indent, and no empty whitespace after EOL....the `pioasm` output always has excess whitespace which requires running the restyle script).

done.

myklemykle:
Whoops, try again.  (I was testing with the wrong copy of the repo, sorry.)

earlephilhower (approved):
Thx!

earlephilhower:
@myklemykle I added the "invert RX" side to your PR.  Pretty simple addition, just wait for `0` start bit then invert the received bits in the IRQ handler.

PR #1461, opened 2023-05-20
Fix I2S::write(const uint8_t *buffer, size_t size)

LinusHeu:
According to #1450 
Added `|| !_running || !_isOutput` for consistency with the other methods.


earlephilhower (approved):
Thanks for the fix!

PR #1468, opened 2023-05-23
BREAKING: Swap Wire1 and Wire in Adafruit Feather

### LINKED ISSUE #1465: Incorrect I2C port assignments on Adafruit RP2040 Feather
The Adafruit RP2040 Feather board assigns both the SCL/SDA pins on its header and Qwiic connector to RP2040 pins 2,3 which is defined as Wire1 in the file `variants/adafruit_feather/pins_arduino`.
```
// Wire
#define PIN_WIRE0_SDA  (24u)
#define PIN_WIRE0_SCL  (25u)
#define PIN_WIRE1_SDA  (2u)
#define PIN_WIRE1_SCL  (3u)
```
As a result it's necessary to override the default port value in most all interfacing examples which assume Wire/Wire0. This is especially problematic for the miniTFTwing which does not allow overriding the default port in its constructor.

earlephilhower:
Fixes #1465

### DIFF
--- a/variants/adafruit_feather/pins_arduino.h
+++ b/variants/adafruit_feather/pins_arduino.h
@@ -30,10 +30,12 @@
 #define PIN_SPI1_SS    (31u)
 
 // Wire
-#define PIN_WIRE0_SDA  (24u)
-#define PIN_WIRE0_SCL  (25u)
-#define PIN_WIRE1_SDA  (2u)
-#define PIN_WIRE1_SCL  (3u)
+#define __WIRE0_DEVICE i2c1
+#define PIN_WIRE0_SDA  (2u)
+#define PIN_WIRE0_SCL  (3u)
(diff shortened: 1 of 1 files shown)


caternuson:
Note this is a breaking change. All current sketches that are working via use of `Wire1` will now break.

However - this update is OK. It changes things for the better in the long run.

earlephilhower:
Noted and this might wait until 3.2 or 4.0.  It's always hard to figure out when to remove technical debt when it's been worked around by end users...

caternuson:
Yep, from the Adafruit side - OK to merge/release whenever. This change was already being contemplated. Adafruit will update Learn docs etc. and handle issues resulting from the breaking change.

emeb:
I've applied this change to my project and it appears to work as desired.

Thanks to all!

PR #1483, opened 2023-05-27
Added function overload to pass name to setup().

djpearman:
Added a simple function overload to BTstackManager::setup()  to allow a name for the BT instance to be passed.


earlephilhower (approved):
LGTM, thanks!

Issue #1470, opened 2023-05-24
CoreMutex aquiring for FreeRTOS


hreintke:
Hi,

Working with the FreeRTOS library at the moment.

Running with Picoprobe(CMSIS-DAP) as upload method and Pico SDK as USB stack, so using 2 USB ports.

Noticed that I do not get all Serial.printf(...) data on the USB port.

This is my current smallest testprogram : 

```
#include "Arduino.h"
#include "freertos.h"
#include "task.h"
#include <map>

static inline TickType_t MsToTicks(TickType_t milliseconds)
{
    return milliseconds / portTICK_PERIOD_MS;
}

void ledTask(void*);
void setup()
{
    Serial.begin(115200);

    pinMode(16, OUTPUT);
    pinMode(32, OUTPUT);
    xTaskCreate(ledTask,"Led16", 1024, reinterpret_cast<void*>(16), configMAX_PRIORITIES - 1, NULL);
    xTaskCreate(ledTask,"Led32", 1024, reinterpret_cast<void*>(32), configMAX_PRIORITIES - 1, NULL);
}

void ledTask(void* param)
{
	int ledPin = reinterpret_cast<int>(param);
	int taskCount = 0;
	bool ledValue = false;
	char* taskName = pcTaskGetName(NULL);
    for(;;)
    {
        vTaskDelay(MsToTicks(1000));
        digitalWrite(ledPin, ledValue ? HIGH:LOW);
        ledValue = !ledValue;

        if (ledPin == 16)
        {
        	Serial.printf("Led16 %3d\r\n", taskCount++);
        }
        else if (ledPin == 32)
        {

        	Serial.printf("Led32 %3d\r\n", taskCount++);
		}
        else
        {
        	Serial.printf("LedXX %3d\r\n", taskCount++);
        }
    }
}

void loop()
{
} 
```

Which gives (varying) results like : 

```
Led32  12
Led32  13
Led32  14
Led32  15
Led32  17
Led32  19
Led32  20
Led32  21
```

In this case, all output from led16 task is lost.

Found the reason for this behavior : 

In SerialUSB there is in the write function : 

```
size_t SerialUSB::write(const uint8_t *buf, size_t length) {
    CoreMutex m(&__usb_mutex, false);
    if (!_running || !m) {
        return 0;
    }
```

And in CoreMutex.cpp : 

```
CoreMutex::CoreMutex(mutex_t *mutex, uint8_t option) {
    _mutex = mutex;
[...]

PR #1484, opened 2023-05-27
CoreMutex add portYield

hreintke:
This is a PR wich includes :

The same update for Mutex/ISR for https://github.com/earlephilhower/arduino-pico/issues/1470
A possible way to implement a coreMutex_give_fromISR which includes a portYield option.
Not yet tested, can do that after june 5.

### DIFF
--- a/cores/rp2040/CoreMutex.cpp
+++ b/cores/rp2040/CoreMutex.cpp
@@ -30,12 +30,11 @@ CoreMutex::CoreMutex(mutex_t *mutex, uint8_t option) {
     _option = option;
     if (__isFreeRTOS) {
         auto m = __get_freertos_mutex_for_ptr(mutex);
-        if (__freertos_check_if_in_isr()) {
-            __freertos_mutex_take_from_isr(m);
-        } else {
-            if (!__freertos_mutex_try_take(m)) {
+        if (__freertos_check_if_in_isr() && !__freertos_mutex_take_from_isr(m)) {
                 return;
             }
+        else {
+            __freertos_mutex_take(m);
         }
     } else {
         uint32_t owner;
@@ -57,7 +56,7 @@ CoreMutex::~CoreMutex() {
         if (__isFreeRTOS) {
             auto m = __get_freertos_mutex_for_ptr(_mutex);
             if (__freertos_check_if_in_isr()) {
-                __freertos_mutex_give_from_isr(m);
+                __freertos_mutex_give_from_isr(m, true);
             } else {
                 __freertos_mutex_give(m);
             }
--- a/cores/rp2040/_freertos.h
+++ b/cores/rp2040/_freertos.h
@@ -43,10 +43,10 @@ extern "C" {
     extern SemaphoreHandle_t _freertos_recursive_mutex_create() __attribute__((weak));
 
     extern void __freertos_mutex_take(SemaphoreHandle_t mtx) __attribute__((weak));
-    extern void __freertos_mutex_take_from_isr(SemaphoreHandle_t mtx) __attribute__((weak));
+    extern int __freertos_mutex_take_from_isr(SemaphoreHandle_t mtx) __attribute__((weak));
     extern int __freertos_mutex_try_take(SemaphoreHandle_t mtx) __attribute__((weak));
     extern void __freertos_mutex_give(SemaphoreHandle_t mtx) __attribute__((weak));
-    extern void __freertos_mutex_give_from_isr(SemaphoreHandle_t mtx) __attribute__((weak));
+    extern void __freertos_mutex_give_from_isr(SemaphoreHandle_t mtx, bool portYield = false) __attribute__((weak));
 
     extern void __freertos_recursive_mutex_take(SemaphoreHandle_t mtx) __attribute__((weak));
     extern int __freertos_recursive_mutex_try_take(SemaphoreHandle_t mtx) __attribute__((weak));
--- a/libraries/FreeRTOS/src/variantHooks.cpp
+++ b/libraries/FreeRTOS/src/variantHooks.cpp
@@ -58,8 +58,8 @@ extern "C" {
         xSemaphoreTake(mtx, portMAX_DELAY);
     }
 
-    void __freertos_mutex_take_from_isr(SemaphoreHandle_t mtx) {
-        xSemaphoreTakeFromISR(mtx, NULL);
+    int __freertos_mutex_take_from_isr(SemaphoreHandle_t mtx) {
+        return xSemaphoreTakeFromISR(mtx, NULL);
     }
 
     int __freertos_mutex_try_take(SemaphoreHandle_t mtx) {
@@ -70,8 +70,15 @@ extern "C" {
         xSemaphoreGive(mtx);
     }
 
-    void __freertos_mutex_give_from_isr(SemaphoreHandle_t mtx) {
-        xSemaphoreGiveFromISR(mtx, NULL);
+    void __freertos_mutex_give_from_isr(SemaphoreHandle_t mtx, bool portYield) {
+        if (!portYield) {
+            xSemaphoreGiveFromISR(mtx, NULL);
+        }
+        else {
+            BaseType_t pxHigherPriorityTaskWoken;
+            xSemaphoreGiveFromISR(mtx, &pxHigherPriorityTaskWoken);
+            portYIELD_FROM_ISR(pxHigherPriorityTaskWoken);
+        }
     }
 
     void __freertos_recursive_mutex_take(SemaphoreHandle_t mtx) {


cores/rp2040/_freertos.h
```diff
@@ -43,10 +43,10 @@ extern "C" {
     extern SemaphoreHandle_t _freertos_recursive_mutex_create() __attribute__((weak));
 
     extern void __freertos_mutex_take(SemaphoreHandle_t mtx) __attribute__((weak));
-    extern void __freertos_mutex_take_from_isr(SemaphoreHandle_t mtx) __attribute__((weak));
+    extern int __freertos_mutex_take_from_isr(SemaphoreHandle_t mtx) __attribute__((weak));
```
earlephilhower, line 46: This call also needs to include a pointer param to return the portYield bool returned from the takeFromISR call.

cores/rp2040/CoreMutex.cpp
```diff
@@ -30,12 +30,11 @@ CoreMutex::CoreMutex(mutex_t *mutex, uint8_t option) {
     _option = option;
     if (__isFreeRTOS) {
         auto m = __get_freertos_mutex_for_ptr(mutex);
-        if (__freertos_check_if_in_isr()) {
-            __freertos_mutex_take_from_isr(m);
-        } else {
-            if (!__freertos_mutex_try_take(m)) {
+        if (__freertos_check_if_in_isr() && !__freertos_mutex_take_from_isr(m)) {
```
earlephilhower, line 33: Here, the take-from-isr call should pass in a new CoreMutex member variable to store the portYield return value

cores/rp2040/CoreMutex.cpp
```diff
@@ -57,7 +56,7 @@ CoreMutex::~CoreMutex() {
         if (__isFreeRTOS) {
             auto m = __get_freertos_mutex_for_ptr(_mutex);
             if (__freertos_check_if_in_isr()) {
-                __freertos_mutex_give_from_isr(m);
+                __freertos_mutex_give_from_isr(m, true);
```
earlephilhower, line 59: Here, return that stored member variable and I think we're good.  This isolates all changes to the CoreMutex object itself, transparently.

cores/rp2040/_freertos.h
```diff
@@ -43,10 +43,10 @@ extern "C" {
...
     extern int __freertos_mutex_try_take(SemaphoreHandle_t mtx) __attribute__((weak));
     extern void __freertos_mutex_give(SemaphoreHandle_t mtx) __attribute__((weak));
-    extern void __freertos_mutex_give_from_isr(SemaphoreHandle_t mtx) __attribute__((weak));
+    extern void __freertos_mutex_give_from_isr(SemaphoreHandle_t mtx, bool portYield = false) __attribute__((weak));
```
earlephilhower, line 49: I'd drop the default `false` because we'll always track its value in CoreMutex.

libraries/FreeRTOS/src/variantHooks.cpp
```diff
@@ -70,8 +70,15 @@ extern "C" {
         xSemaphoreGive(mtx);
     }
 
-    void __freertos_mutex_give_from_isr(SemaphoreHandle_t mtx) {
-        xSemaphoreGiveFromISR(mtx, NULL);
+    void __freertos_mutex_give_from_isr(SemaphoreHandle_t mtx, bool portYield) {
+        if (!portYield) {
```
earlephilhower, line 74: Here, I'm a little fuzzy.  When we take the mutex, FreeRTOS returns a "higherTaskWoken" bool which,m if true, means we need to do a context switch on the giveFromISR.  That's the `portYield` param here.  https://www.freertos.org/xSemaphoreTakeFromISR.html

But the giveFromISR also seems to return the same flag:  https://www.freertos.org/xSemaphoreTakeFromISR.html  .

So, maybe the logic for this function should be:
````
BaseType_t pxHigherPriorityTaskWoken;
xSemaphoreGiveFromISR(mtx, &pxHigherPriorityTaskWoken);
portYIELD_FROM_ISR( (pxHigherPriorityTaskWoken || portYield) ? pdTRUE : pdFALSE);
````
?

**Review by earlephilhower — CHANGES_REQUESTED** [review]
I think you're on to something here, but a little more work probably needed to get it correct.  FreeRTOS, mutexes, and ISRs can give one nightmares. :)

File: libraries/FreeRTOS/src/variantHooks.cpp (hunk 5 above)
Comment (hreintke) [reply libraries/FreeRTOS/src/variantHooks.cpp:74]: The pxHigherPriorityTaskWoken can be used multiple times.

From the docs

> FreeRTOS API functions can only set *pxHighPriorityTaskWoken to pdTRUE. If an ISR
calls more than one FreeRTOS API function, then the same variable can be passed as the
pxHigherPriorityTaskWoken parameter in each API function call, and the variable only
needs to be initialized to pdFALSE before it is used for the first time.

File: cores/rp2040/_freertos.h (hunk 4 above)
Comment (hreintke) [reply cores/rp2040/_freertos.h:49]: change to BaseType_t* and removed default value

File: cores/rp2040/CoreMutex.cpp (hunk 3 above)
Comment (hreintke) [reply cores/rp2040/CoreMutex.cpp:59]: Changed to use the CoreMutex member variable

File: cores/rp2040/_freertos.h (hunk 1 above)
Comment (hreintke) [reply cores/rp2040/_freertos.h:46]: Done

File: cores/rp2040/CoreMutex.cpp [hunk 6]
```diff
@@ -23,19 +23,20 @@
 
 #include "Arduino.h"
 #include "CoreMutex.h"
+#include "projdefs.h"
```
earlephilhower, line 26: Think we got a stray include here...

libraries/FreeRTOS/src/variantHooks.cpp
```diff
@@ -70,8 +70,14 @@ extern "C" {
         xSemaphoreGive(mtx);
     }
 
-    void __freertos_mutex_give_from_isr(SemaphoreHandle_t mtx) {
-        xSemaphoreGiveFromISR(mtx, NULL);
+    void __freertos_mutex_give_from_isr(SemaphoreHandle_t mtx, BaseType_t* pxHigherPriorityTaskWoken) {
+        if (!pxHigherPriorityTaskWoken) {
```
earlephilhower, line 74: The GiveFromISR might also set the flag, so I think the `if` is not needed or correct, no?

Given the description, should this whole function be
````
            xSemaphoreGiveFromISR(mtx, pxHigherPriorityTaskWoken);
            portYIELD_FROM_ISR(pxHigherPriorityTaskWoken);
````
?

Comment (hreintke) [reply libraries/FreeRTOS/src/variantHooks.cpp:74]: I implemented it that way because I am not sure how `portYIELD_FROM_ISR(NULL);` is working.

Will check.

File: cores/rp2040/CoreMutex.cpp (hunk 6 above)
Comment (hreintke) [reply cores/rp2040/CoreMutex.cpp:26]: I got a compile error on pdFALSE. Tried to find an include that holds the definition.

PS : Same for `BaseType_t` 

Will check.

PS 2 : Will be out for my break in 2 hours time

File: libraries/FreeRTOS/src/variantHooks.cpp (hunk 7 above)
Comment (earlephilhower) [reply libraries/FreeRTOS/src/variantHooks.cpp:74]: Maybe it's as simple as this?  You pass in NULL, it starts at FALSE, OTW it starts at whatever the earlier call was.
````
void __freertos_mutex_give_from_isr(SemaphoreHandle_t mtx, BaseType_t* pxHigherPriorityTaskWoken) {
    BaseType_t hiprio = pxHigherPriorityTaskWoken ? *pxHigherPriorityTaskWoken : pdFALSE;
    xSemaphoreGiveFromISR(mtx, &hiprio);
    portYIELD_FROM_ISR(&hiPrio);
}
````

**Review by earlephilhower — APPROVED** [review]

**Comment by earlephilhower** [comment]
Sorry for the churn on the pushes.  Editing single files at a time from the GH website is a painful process.  That said, I've merged the latest head and my suggestion for the IRQ portYield bit.  If you could take a look at it and give it a try in your own code when you get back, it would be much appreciated.  There are a few FreeRTOS and I2S fixes (including this one!) I'd like to get out in a new release later this week...

**Comment by earlephilhower** [comment]
Interesting, thanks!   I will look into it myself as I get time.  Basic FreeRTOS code seems to run much better w/the `mutex_take` vs `mutex_try_take`, so it seems only the IRQ bit has been munged up.

What puzzles me is that this code should run *exactly* the same path in FreeRTOS since before the PR the `gpioInterruptDispatcher` was hard-setting the "inInterrupt" flag causing CoreMutex to use `__freertos_mutex_take_from_isr`.  Now, only difference is it uses the FreeRTOS internals to determine if it's in an IRQ.  Maybe that `portCHECK_IF_IN_ISR` isn't alwasy correct. :(

**Comment by earlephilhower** [comment]
See #1510.  There was a logic bug, but it would only hit if the IRQ was unable to grab the mutex.

I don't have an LED on pin 16 (just std. Pico), so I ran the following and then grounded pin 17 and the blink went from slow to fast, as expected, with #1510.

````
#include "Arduino.h"
#include "FreeRTOS.h"

volatile int delayValue = 1000;

void intRoutine()
{
 delayValue = 100;
}

void setup()
{
  Serial.begin(115200);
  pinMode(LED_BUILTIN, OUTPUT);
  pinMode(17, INPUT_PULLUP);
  attachInterrupt(17, intRoutine, CHANGE);
}

bool ledOn = true;
void loop()
{
  digitalWrite(LED_BUILTIN,ledOn);
  ledOn = !ledOn;
  delay(delayValue);
}
````

**Comment by hreintke** [comment]
Thanks for solving the issue. 
A little to quick in optimizing code. Glad that I mentioned that it was not fully tested and found the issue during the test.

Did also test with acquiring a mutex that is already taken and that works OK too.

```
volatile int delayValue = 1000;

// using pointer to be sure they are instanciated correctly during setup
mutex_t* tm;
CoreMutex* cm;

void intRoutine()
{
	if (CoreMutex{tm,0}){
		delayValue = 100;
	}
}

void setup()
{
	Serial.begin(115200);
	pinMode(16, OUTPUT);
	pinMode(17, INPUT_PULLUP);
	tm = new mutex_t;
	cm = new CoreMutex{tm,0};
	attachInterrupt(17, intRoutine, CHANGE);
}

bool ledOn = true;
int cnt = 0;
void loop()
{
	digitalWrite(16,ledOn);
	ledOn = !ledOn;
	delay(delayValue);
	cnt++;
	if (cnt == 10)
	{
		delete cm;
	}
}
``` 

On this PR. Code looks OK, but also for this not yet fully tested. 
Need to setup quite a FreeRTOS program with taskPriorities at different levels to make it complete.
Won't be able to finish that this week but will do.

Related :
In the current implementation you acquiring a mutex and do a std:map lookup for every gpio interrupt.
Does add to interrupt latency. Might be better to have a std::array, indexed by gpio number for the callbacks.
That could be accessed without the need to lock.
If you want I can take a look and do some latency measurements, on that and come with a proposal.

earlephilhower:
Thanks, would appreciate your testing.  I think I'll merge this and do a release in a day or two to get even more testing in the field.  If it really kills things users can move back a release always,

Re std::map, I'm not sure it's worth the effort but if you really care, be my guest.  The std::map uses a binary search  O(log2(#-of-interrupts) while the std::array will be an O(1) since it's a direct lookup.  But for the very small # of IRQs I think they're pretty close in absolute time.  Plus, the std::map only allocates space as IRQs are assigned.  So memory is ~12 bytes overhead + #-irqs * 8.  std::array will allocate all pointers so will be 30 * 4 = 120 bytes, always, which is non-trivial.

GUVWAF:
Just for your information: we did some "testing in the field" with Meshtastic with good results. Using version 3.2.1 of this core, we observed that the device went into `rtosFatalError` every now and then (see https://github.com/meshtastic/firmware/issues/2546), which doesn't happen when using 3.2.2. So, thanks for the great work!

earlephilhower:
Thanks for the feedback!  If you run into something else like this, please don't hesitate to open an issue here even if you don't have a PR to solve it.  Can't fix what we don't know is broken, etc.

PR #1487, opened 2023-05-30
make dma_claim_unused_channel() not panic in AudioBufferManager::begin()

LinusHeu:
If `dma_claim_unused_channel(true)` panics when none are available, then why have the check if it returned -1? 
I think you might mean `dma_claim_unused_channel(false)`, so that begin() can just return false.

Warning:
I'm sorry, but I haven't tested it. And to be honest, I'm not even completely sure what exactly these functions do. Please review carefully. I hope this is still useful.


earlephilhower (approved):
Thanks, looks good!

PR #1513, opened 2023-06-08
Add "Reboot to bootloader" function

madias123:
Easy way to reboot to build in bootloader via code.
Usage:
`rp2040.rebootToBootloader();`


madias123:
...version mismatch - ignore

PR #1519, opened 2023-06-10
fix for invalid __channelCount in ~AudioBufferManager()

lyusupov:
**AudioOutputPWM** method from [ESP8266Audio](https://github.com/earlephilhower/ESP8266Audio) library plays very first audio file just fine,
but gets stuck on a second one.

This is the gdb batcktrace:

```
(gdb) bt
#0  dma_channel_get_irq0_status (channel=3)
    at /home/user/Arduino/hardware/pico/rp2040//pico-sdk/src/rp2_common/hardware_dma/include/hardware/dma.h:633
#1  AudioBufferManager::_irq ()
    at /home/user/Arduino/hardware/pico/rp2040/libraries/AudioBufferManager/src/AudioBufferManager.cpp:281
#2  0x20000cb8 in irq_handler_chain_slots ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
(gdb)
```

It was found out that the issue is caused by ~AudioBufferManager() destructor that have not freed the irq resource properly after the first run.

This commit fixes the issue. 
**AudioOutputPWM** (ESP8266Audio) can play multiple of WAV files now.

### DIFF
--- a/libraries/AudioBufferManager/src/AudioBufferManager.cpp
+++ b/libraries/AudioBufferManager/src/AudioBufferManager.cpp
@@ -78,8 +78,8 @@ AudioBufferManager::~AudioBufferManager() {
             dma_channel_abort(_channelDMA[i]);
             dma_channel_unclaim(_channelDMA[i]);
             dma_channel_acknowledge_irq0(_channelDMA[i]);
+            __channelCount--;
         }
-        __channelCount--;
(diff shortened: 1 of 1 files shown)


earlephilhower (approved):
Thanks, appreciate the PR!  Looks good and logically makes sense.

PR #1524, opened 2023-06-12
I2S: reuse current PIO & SM settings on a second and further runs

lyusupov:
Since PIOProgram class destructor is not implemented - 'delete _i2s' operation has no any noticeable effect.
_i2s->prepare(&_pio, &_sm, &off) is trying to allocate new PIO & SM on every begin() call which causes troubles when multiple of WAV files being played.

### DIFF
--- a/libraries/I2S/src/I2S.cpp
+++ b/libraries/I2S/src/I2S.cpp
@@ -46,6 +46,7 @@ I2S::I2S(PinMode direction) {
     }
 #endif
     _freq = 48000;
+    _pio = nullptr;
     _arb = nullptr;
     _cb = nullptr;
     _buffers = 6;
@@ -145,7 +146,10 @@ bool I2S::begin() {
     } else {
         _i2s = new PIOProgram(_isOutput ? (_isLSBJ ? &pio_lsbj_out_swap_program : &pio_i2s_out_swap_program) : &pio_i2s_in_swap_program);
     }
-    _i2s->prepare(&_pio, &_sm, &off);
+    /* reuse current PIO & SM settings on a second run because PIOProgram class has no destructor implemented */
+    if (_pio == nullptr) {
+      _i2s->prepare(&_pio, &_sm, &off);
+    }
     if (_isOutput) {
         if (_isLSBJ) {
             pio_lsbj_out_program_init(_pio, _sm, off, _pinDOUT, _pinBCLK, _bps, _swapClocks);
--- a/libraries/I2S/src/I2S.cpp
+++ b/libraries/I2S/src/I2S.cpp
@@ -148,7 +148,7 @@ bool I2S::begin() {
     }
     /* reuse current PIO & SM settings on a second run because PIOProgram class has no destructor implemented */
     if (_pio == nullptr) {
-      _i2s->prepare(&_pio, &_sm, &off);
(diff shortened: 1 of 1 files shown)


LinusHeu:
Wouldn't it be better to properly de-allocate the PIO & SM instead? (Isn't that already done in I2S::end() with         pio_sm_set_enabled(_pio, _sm, false); ?)

earlephilhower:
As @LinusHeu says, I think this is a workaround for a bug in `PIOProgram`.  I should release the SM in its destructor and then things should "just work."  The PIOProgram won't allocate new PIO program space, since it keys off of the program address in flash (a constant here), and will re-use the existing code in the PIO instruction RAM if you create another object with the same insns.

lyusupov:
Sure, the proper `PIOProgram`. class destructor is the best solution.
Feel yourself free to close this PR when you will complete implementation of  the destructor.

earlephilhower:
Can you try #1526 and report back?  What I *thought* I wrote and what I *actually* wrote for the PIOProgram were wildly different things so I redid it.  It should now properly re-use INSN memory and free SMs.

lyusupov:
@earlephilhower 

Please, look that `I2S::begin()` does not inspect return value of ` _i2s->prepare(&_pio, &_sm, &off) `
This may cause issues _(specifically with Pico W)_ when all the PIOs and SMs are already exhausted.

```
bool I2S::begin() {
    _running = true;
    _hasPeeked = false;
    _isHolding = 0;
    int off = 0;
    if (!_swapClocks) {
        _i2s = new PIOProgram(_isOutput ? (_isLSBJ ? &pio_lsbj_out_program : &pio_i2s_out_program) : &pio_i2s_in_program);
    } else {
        _i2s = new PIOProgram(_isOutput ? (_isLSBJ ? &pio_lsbj_out_swap_program : &pio_i2s_out_swap_program) : &pio_i2s_in_swap_program);
    }
    _i2s->prepare(&_pio, &_sm, &off);
    if (_isOutput) {
```

earlephilhower:
@lyusupov done, #1528, thanks.  Also checks the DMA setup succeeds, too.  Probably need to scour the rest of the code for similar issues outside I2S.

lyusupov:
Thank you!

One more question:

Is it Ok that `_holdWord `  defined in I2S.h as follows

```
    int32_t _holdWord = 0;
    int _isHolding = 0;

```

is not cleared in `I2S::begin()` ?

```
bool I2S::begin() {
    _running = true;
    _hasPeeked = false;
    _isHolding = 0;
    int off = 0;
    if (!_swapClocks) {
```
  
This may have a negative effect that the variable may keep old value after I2S.end() -> I2S.begin() cycle.

earlephilhower:
> Is it Ok that _holdWord ... is not cleared in I2S::begin() ?

Yes, because the `isHolding` is the important bit (how may bits we have in the temp buffer.  By setting to 0 we will overwrite `_holdWord` in `I2s::read()` at https://github.com/earlephilhower/arduino-pico/blob/d18f8dce2f9bbd0e272ace123ae5c27cd1926395/libraries/I2S/src/I2S.cpp#L229-L232

Same logic on the ::write side (but in multiple fcns).

PR #1526, opened 2023-06-12
Enable proper reuse of PIO programs

earlephilhower:
Rewrite the PIOProgram helper class to properly re-use loaded programs and to try to re-use loaded instructions before allocating a new PIO program.

Supersedes #1524

### DIFF
--- a/cores/rp2040/PIOProgram.cpp
+++ b/cores/rp2040/PIOProgram.cpp
@@ -22,8 +22,7 @@
 #include "PIOProgram.h"
 #include <map>
 
-std::map<const pio_program_t *, int> __pio0Map;
-std::map<const pio_program_t *, int> __pio1Map;
+static std::map<const pio_program_t *, int> __pioMap[2];
 auto_init_mutex(_pioMutex);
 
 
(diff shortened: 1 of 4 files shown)


earlephilhower:
I've tested with the following 2 sketches as I2S/PIO torture tests and was able to restart I2S dozens of times (i.e. many more times than there are SMs or INSN space if there was a leak):

````
#include <I2S.h>
void setup() {
}

int cnt = 0;
void loop() {
  I2S i2s = I2S(OUTPUT);
  i2s.setBCLK(19);  // WS = 27
  i2s.setDATA(21);
  i2s.setBitsPerSample(16);
  i2s.setBuffers(4, 64);
  while (1) {
    delay(5000);
    Serial.printf("Run %d\n", ++cnt);
    if (!i2s.begin(22040)) {
      Serial.printf("fail init I2S\n");
    } else {
      delay(1000);
      i2s.end();
    }
  }
}
````


````
#include <I2S.h>
void setup() {
}

int cnt = 0;
void loop() {
  while (1) {
    delay(5000);
    Serial.printf("Run %d\n", ++cnt);
    I2S i2s = I2S(OUTPUT);
    i2s.setBCLK(19);  // WS = 27
    i2s.setDATA(21);
    i2s.setBitsPerSample(16);
    i2s.setBuffers(4, 64);
    if (!i2s.begin(22040)) {
      Serial.printf("fail init I2S\n");
    } else {
      delay(1000);
      i2s.end();
    }
  }
}
```

earlephilhower:
Also tested with #1528 applied to this patch and the examples above still pass (and they fail as expected w/o this PR).

earlephilhower:
Thanks for the feedback.  Looks like this is a net positive, so I'm going to merge now as-is.  If you can get an MCVE or stack trace for the other failure, please open another issue so we can get that one, too.

nils-trubkin:
Hello,
Is this issue related to the issue described here? https://github.com/earlephilhower/ESP8266Audio/issues/601

I'm experiencing an issue playing a WAV file on my Qt Py RP2040 through an I2S DAC (98357A) from an SD card. The problem is that every second time I try to play the file, the `begin()` function returns `true`, but no sound is produced. According to the logs, the sound should have stopped playing once completed but continues indefinitely (with no actual sound coming). The `loop()` of the WAV generator will continue returning `true` indefinitely.

I suspect the issue is related to the i2s initialization/deinitialization in either the sound library or this library. I've tried different methods to resolve the issue, including reinitializing the I2S, sd source, and WAV generator, but the only solution that works is restarting the entire board. Do you have any suggestions on how I can fix this problem? I'd appreciate any advice on playing multiple sounds separately without repeatedly restarting the entire board. Thank you.

Issue #1522, opened 2023-06-11
Pico W, Wifi HTTP request delays.


hreintke:
Hi,

Running non FreeRTOS, basic webserver example I see some late responses (5 seconds delay, but not always) on HTTP requests.
This only happens when I use a webbroswer, using curl all works as expected (keep tcp connection open ?)

Would like to do some IP traffic analyses for that, using the Netdump library I made for esp8266.

I am not sure, but think there needs to be a patch/addition in lwip netif (phy_capture) for that.
Was not (yet?) be able to track that down in pico-w and/or esp8266.

Do you know where/how to add that ? 
If not, was @d-a-v involved in the wifi/lwip implementation of pico-w ? I can ask him for some help too.

earlephilhower:
@d-a-v was not involved here, sadly.  I'm not sure he'd be able to help w/the PicoW side of things.

My guess would be that if you're using Chrome it's trying to send a bunch of connections in parallel, and since the processing is very much serial in nature on the Pico, things get stuck waiting.  FWIW under Firefox I haven't noticed any issue like this on the AdvancedWebServer.

We are using the unpatched LWIP, straight from the Pico-SDK, but I think there are a couple things low-effort to try.

First, you can grab/dumb all received **Ethernet** level packets by hacking in to https://github.com/earlephilhower/arduino-pico/blob/cc800713bd7b30b19ca4b5f02e88f3e915802527/cores/rp2040/cyw43_wrappers.cpp#L48.   You can get all **Ethernet** sent packets here: https://github.com/earlephilhower/arduino-pico/blob/cc800713bd7b30b19ca4b5f02e88f3e915802527/libraries/lwIP_CYW43/src/utility/CYW43shim.cpp#L108

It is also possible to enable very verbose LWIP debugging by editing https://github.com/earlephilhower/arduino-pico/blob/master/tools/libpico/lwipopts.h (and enabling Tools->Debug Port->Serial to get ::printf output to USB).

earlephilhower:
Another possibility could be a bug in the webserver code, which just happens to have been noted a couple days ago in the upstream sources...  https://github.com/esp8266/Arduino/issues/8941

I have not looked at the code, yet, so can't really comment if it's a real issue or not, but if it is then it's one heck of a coincidence. :four_leaf_clover:

hreintke:
Checked into it.

You are using another webserver implementation  (webserver vs esp8266webserver) so the issue cannot come from there. 
[...]

PR #1537, opened 2023-06-15
Solve http request delay

hreintke:
Serve next wificlient (http_request) if current client does not have data

Solves https://github.com/earlephilhower/arduino-pico/issues/1522
I included also the the proposed change to https://github.com/esp8266/Arduino/issues/8941, just to be sure.  

Disclaimer : I did not test with HTTPS, don't have that running in my environment

Edit : This solution is identical to the current implementation of esp8266

Edit 2 : I think the remaining CI failure is not a result of this PR.


earlephilhower:
Thanks!  LGTM after read-thru.  I wonder if they updated the `aspell` version to cause the weird document failure.  Let me try fixing that in another PR, merge that change here, then I'll merge this one...

PR #1541, opened 2023-06-15
Linker tweaks to place vital PicoDVI library funcs in RAM

PaintYourDragon:
As discussed in previous issue (which I’ll go close in a moment), adjustments to linker script to place certain code in RAM, required for PicoDVI library when used in combination with flash-writing code (e.g. filesystem).

If you’d prefer, I could go through PicoDVI.cpp and dvi.c and extensively decorate all functions there as not-in-flash (so they could then be removed here), but regardless this is _still_ going to require *interp.c.obj *divider.S.obj in the linker script.

For non-PicoDVI projects, interp and divider are obscure and/or small such that they’ll have minimal or no impact on RAM.


earlephilhower (approved):
LGTM, thanks!

PR #1551, opened 2023-06-20
fix typo in fs.rst regarding info64

mef51:
Example code for [`info64`](https://arduino-pico.readthedocs.io/en/latest/fs.html#info64) has the incorrect function calls


earlephilhower (approved):
Thanks!

PR #1555, opened 2023-06-22
Palmerr23 i2 s mclk

palmerr23:
Sorry about the messy PR, I updated the arduino-pico board files after developing this and had to update some of your I2S.cpp code by hand.

Note also the change to setFrequency(long) - simply to match the documentation.

I'll add a PR for the doco later.

Main changes:
Add MCLK support - enabled by setting the MCLK pin. 
Sample rate to MCLK multiplier can be independently set to multiples of 64.
Changing the setting of the PIO dividers from float mode to 16:8 mode and restricting the clock dividers to 16:0 only, so that MCLK can be reliably synchronised.
Add SYSCLK to sample rate optimisation - by changing the processor frequency.


palmerr23:
Hold off on this Earle - the update to the board files has broken a few things.

palmerr23:
I think everything is OK now. Multiple copy and pastes to align my PR with the 3.3.0 release that occurred midway in development of this code.

libraries/I2S/src/I2S.cpp
```diff
@@ -199,13 +250,18 @@ void I2S::end() {
...
         avail *= 4; // 4 samples per 32-bits
-        if (_bps < 24 && !_isOutput) {
-            avail += _isHolding / 8;
+        if (_bps < 24) {
+            if (_isOutput) {
+                // 16- and 8-bit can have holding bytes available
+                avail += (32 - _isHolding) / 8;
+            } else {
+                avail += _isHolding / 8;
+            }
```
LinusHeu, line 264: I think you made a merge mistake here. You're undoing my #1548

palmerr23: Reinserted code into pull request as per #1548

palmerr23:
Sorry Linus, 

I'll fix it.

palmerr23:
Reinserted #1548 code from Linus.

earlephilhower:
Can you run `tools/restyle.sh` to fix the formatting issues?

Also, it looks like that maybe there are some old/missing files:
````
/home/runner/arduino_ide/hardware/pico/rp2040/libraries/I2S/src/I2S.cpp: In member function 'void I2S::MCLKbegin()':
/home/runner/arduino_ide/hardware/pico/rp2040/libraries/I2S/src/I2S.cpp:180:32: error: 'pio_i2s_mclk_program' was not declared in this scope; did you mean 'pio_i2s_in_program'?
  180 |     _i2sMCLK = new PIOProgram(&pio_i2s_mclk_program);
      |                                ^~~~~~~~~~~~~~~~~~~~
      |                                pio_i2s_in_program
/home/runner/arduino_ide/hardware/pico/rp2040/libraries/I2S/src/I2S.cpp:182:5: error: 'pio_i2s_MCLK_program_init' was not declared in this scope; did you mean 'pio_i2s_in_program_init'?
  182 |     pio_i2s_MCLK_program_init(_pioMCLK, _smMCLK, off, _pinMCLK);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~
      |     pio_i2s_in_program_init
Using library I2S at version 2.0 in folder: /home/runner/arduino_ide/hardware/pico/rp2040/libraries/I2S 
Using library AudioBufferManager at version 1.0.0 in folder: /home/runner/arduino_ide/hardware/pico/rp2040/libraries/AudioBufferManager 
exit status 1
````

**Comment by palmerr23** [comment]
Will do, probably tomorrow.

**Comment by jfrey-xx** [comment]
Also, maybe I missed it, but there is no default value for `_pinMCLK`?

**Comment by palmerr23** [comment]
Jeremy,

That is correct. As this is an independent PIO program and not using multiple pins, any legitimate pin can be assigned as for DIN/DOUT.

**Comment by palmerr23** [comment]
Jfrey,

Yes. I'll add code to terminate the MCLK PIO program on I2S exit.

**Comment by jfrey-xx** [comment]
> Jeremy,
> 
> That is correct. As this is an independent PIO program and not using multiple pins, any legitimate pin can be assigned as for DIN/DOUT.

`_pinDOUT` is set to 28 in the constructor, I am just wondering what would happen here with `_pinMCLK`, if `setMCLK()` is not called upon start -- it defaults to 0 and not a random value?

**Comment by palmerr23** [comment]
Good point. I think PIN 27 makes a more sensible default than PIN 0.

**Comment by earlephilhower** [comment]
> Good point. I think PIN 27 makes a more sensible default than PIN 0.

I would suggest that, by default, no pin is used since most codecs don't require it.  Set the default pin to `255` or something and check and only if the app tries to set a pin should we start driving an MCLK..

**Comment by palmerr23** [comment]
Oops! make that pin 25, pin 27 is LRCLK.

**Comment by palmerr23** [comment]
Earle,

Yes that's the default behaviour: MCLK is not driven unless the MCLK pin is set.

There is no harm, however in Jeremy's suggestion that a default be set (and not assigned to anything active) in the constructor.

I've also added a default of 256 for _multMCLK

I'll run the tidy-up routine and update the pull, once I've tested the additions.

**Comment by palmerr23** [comment]
Updates now restyled with restyle.sh

Additions of default pins and stopping MCKL in i2s.end() now tested.

**Comment by palmerr23** [comment]
I wrote some code to test all possible sysclk and PIO divider permutations. There may be more at higher sysclks, but I stopped checking at 200MHz.

The two I have suggested in the doco are:
44.1kHz (etc) 135.6 MHz. Sample rate error = 0.092%
48kHz (etc) 147.6 MHz. Sample rate error = 0.098%

I found  two further 'good' sysclks:
44.1kHz (etc) 158 MHz. Sample rate error = 0.034%
48kHz (etc) 172 MHz. Sample rate error = 0.019%

While these are even closer to the desired sample rates, they are more overclocked. Should I include them in the doco for completeness?

**Comment by lyusupov** [comment]
For the Pico PIO USB function to work properly it requires
CPU clock must be multiple of 120 Mhz -

https://github.com/sekigon-gonnoc/Pico-PIO-USB/blob/main/examples/arduino/device_info/device_info.ino#L50

Could this MCLK PR co-exist with PIO USB ?

**Comment by palmerr23** [comment]
I've just tested at 11025 and 22050 and they work OK.

Also, the nominated frequency for 48k has been tested OK at 8k, 16k, 32k 48k, 96k and 192k in both 16 and 24 bit words.

Note, these tests have been made on a PCM5102A and confirmed on a logic analyser and the audio output analysed for distortion. Other DACs and CODECs may differ! Happy to hear how others go. I have several different CODECs which I may test in due course, probably when I've written the I2S slave code - to save pins.

No, the MCLK option won't work at sysclk = 120MHz. I'd advise using the LRCLK/BCLK only method at other than the specified frequencies, as MCLK generation needs to be strictly frequency and phase controlled to BCLK.

**Comment by lyusupov** [comment]
> > setSysClk() has support for 44100 and 8000 sample rates only. Is there any chance to support 11025 rate as well ?

> I've just tested at 11025 and 22050 and they work OK.

The code below is deprecated at this time, right ?

```
bool I2S::setSysClk(int samplerate) { // RP - optimise sys_clk for desired samplerate 
    if(samplerate % 44100 == 0){
        set_sys_clock_khz(I2SSYSCLK_44_1, false); // 147.6 unsuccessful - no I2S no USB
        return true;
    }
    if(samplerate % 8000 == 0) {
          set_sys_clock_khz(I2SSYSCLK_8, false); 
          return true;
    }
    return false; 
}
```

**Comment by palmerr23** [comment]
You are correct, I'll fix the code - and no the setSysClk() code  was not supposed to be deprecated. Unfortunately, I had to switch from Windows to Linux to run tests/restyle.sh and it reformatted every file in the repo, so I had to reload the entire arduino-repo. Many of my proposed changes appear to have been been lost in the process.

The workflow should stall until I can re-make them from local copies.

**Comment by palmerr23** [comment]
I think all the changes have been re-entered, and setSysck() now has 11025 as the 44.1kHz base.

**Comment by earlephilhower** [comment]
There's still something missing or old files included as the PR isn't building.  Requires a `tools/restyle.sh` run, too.

**Comment by palmerr23** [comment]
Earle,

Yes, some things got lost in translation when I blew up my fork yesterday. Hopefully, everything is back correctly, now.

I also added an example for using the MCLK option.

restyle.sh has been executed, just on the libraries/I2S folder.

**Comment by earlephilhower** [comment]
Unfortunately still some compilation errors.  You'll need to re-run the styler, too, and include the examples directory...

**Comment by palmerr23** [comment]
Fixed styling and added missing #includes to i2s.h

**Comment by earlephilhower** [comment]
Took some time, but looks like it's building OK finally!  Because this is a little big, I'll need a few days to look it over.

If anyone watching the PR could give it a try in their application that would be great, too.

File: libraries/I2S/src/pio_i2s.pio.h [hunk 2]
```diff
@@ -217,6 +245,16 @@ static inline pio_sm_config pio_i2s_in_swap_program_get_default_config(uint offs
     return c;
 }
 
+// doesn't yet force the same PIO as the IN/OUT program
```
earlephilhower, line 248: Same as prior comment

libraries/I2S/src/pio_i2s.pio
```diff
@@ -157,6 +166,18 @@ right:
...
 
+// doesn't yet force the same PIO as the IN/OUT program
+static inline void pio_i2s_MCLK_program_init(PIO pio, uint sm, uint offset, uint MCLK_pin) {
+    pio_gpio_init(pio, MCLK_pin);
+		pio_sm_set_consecutive_pindirs(pio, sm, MCLK_pin, 1, true);
+    pio_sm_config sm_config = pio_i2s_mclk_program_get_default_config(offset);   
+    sm_config_set_set_pins(&sm_config, MCLK_pin, 1);
+    pio_sm_init(pio, sm, offset, &sm_config);
+		
+   // pio_sm_set_pins(pio, sm, 0); // clear pins
```
earlephilhower, line 177: For clarity could we remove and not just comment out un-needed code like this and the next line, please?

libraries/I2S/src/pio_i2s.pio.h
```diff
@@ -217,6 +245,16 @@ static inline pio_sm_config pio_i2s_in_swap_program_get_default_config(uint offs
...
+static inline void pio_i2s_MCLK_program_init(PIO pio, uint sm, uint offset, uint MCLK_pin) {
+    pio_gpio_init(pio, MCLK_pin);
+    pio_sm_set_consecutive_pindirs(pio, sm, MCLK_pin, 1, true);
+    pio_sm_config sm_config = pio_i2s_mclk_program_get_default_config(offset);
+    sm_config_set_set_pins(&sm_config, MCLK_pin, 1);
+    pio_sm_init(pio, sm, offset, &sm_config);
+    // pio_sm_set_pins(pio, sm, 0); // clear pins
```
earlephilhower, line 255: Same as prior (but should be re-gen'd automatically)

libraries/I2S/src/pio_i2s.pio
```diff
@@ -157,6 +166,18 @@ right:
 
 % c-sdk {
 
+// doesn't yet force the same PIO as the IN/OUT program
+static inline void pio_i2s_MCLK_program_init(PIO pio, uint sm, uint offset, uint MCLK_pin) {
+    pio_gpio_init(pio, MCLK_pin);
+		pio_sm_set_consecutive_pindirs(pio, sm, MCLK_pin, 1, true);
```
earlephilhower, line 172: Weird indent??

docs/i2s.rst
```diff
@@ -58,6 +68,13 @@ Sets the word clock frequency, but does not start the I2S device if not
 already running.  May be called after ``I2S::begin()`` to change the
 sample rate on-the-fly.
 
+bool setSysClk(int samplerate) 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Changes the PICO system clock to optimise for the desired samplerate. 
+The clock changes to 147.6 MHz for samplerates that are a multiple of 8kHz, and 135.6 MHz for multiples of 44.1kHz.
```
earlephilhower, line 74: The setSysClock actually seems to be for multiples of 8k and 11.025k, not 8k and 44.1k.

libraries/I2S/keywords.txt
```diff
@@ -16,11 +16,14 @@ end	KEYWORD2
 
 setBCLK	KEYWORD2
 setDATA	KEYWORD2
+setMCLK	KEYWORD2 
 setBitsPerSample	KEYWORD2
 setFrequency	KEYWORD2
 setBuffers	KEYWORD2
 setLSBJFormat	KEYWORD2
 swapClocks	KEYWORD2
+setMCLKmult  KEYWORD2
```
earlephilhower, line 25: `keywords.txt` is very fussy and needs to have actual `tab` characters, not spaces, between the word and the `KEYWORD/xxx` type or it will ignore it.  I've been bitten by this myself more than once!

libraries/I2S/src/I2S.cpp
```diff
@@ -188,6 +240,11 @@ bool I2S::begin() {
 }
 
 void I2S::end() {
+    if (_MCLKenabled && _running) {
+        pio_sm_set_enabled(_pioMCLK, _smMCLK, false);
+        _MCLKenabled = false;
```
earlephilhower, line 245: Why disable the mclk on ::end?  We don't change any other settings on ::end so I would say we should let the value stick as-it-was.

libraries/I2S/src/I2S.h
```diff
@@ -21,7 +21,11 @@
 
 #pragma once
 #include <Arduino.h>
+#include <hardware/pio.h>
+#include <pico/stdlib.h>
```
earlephilhower, line 25: Why are we pulling in the SDK's `pico/stdlib`?  Is this just left over from some debugging?  Can it be removed?

libraries/I2S/src/I2S.h
```diff
@@ -30,11 +34,14 @@ class I2S : public Stream {
 
     bool setBCLK(pin_size_t pin);
     bool setDATA(pin_size_t pin);
+    bool setMCLK(pin_size_t pin); // RP - also enables MCLK output
```
earlephilhower, line 37: No need to comment with your initials, the `git blame` will let everyone know any bugs here are your fault, not mine.  :laughing:

libraries/I2S/src/I2S.h
```diff
@@ -110,14 +117,17 @@ class I2S : public Stream {
...
     int _bps;
     int _freq;
+    int _multMCLK;  // RP
     size_t _buffers;
     size_t _bufferWords;
     int32_t _silenceSample;
     bool _isLSBJ;
     bool _isOutput;
     bool _swapClocks;
+    bool _MCLKenabled = false; // RP
```
earlephilhower, line 130: Could we set this value in the constructor, not here?  We are setting the mclk multipler/pin/etc. there already.

libraries/I2S/examples/SquareWaveMCLK/SquareWaveMCLK.ino
```diff
@@ -0,0 +1,83 @@
+/*
+  This example demonstrates I2S output with the optional MCLK  signal.
+  Note: System clock sppeds used here may not be compatible with all other libraries
+  requiring specific sys_clks, particularly Pico PIO USB.
+
+  Richard Palmer June 2023
+
+  Original  17 November 2016
+  by Sandeep Mistry
+  modified for RP2040 by Earle F. Philhower, III <earlephilhower@yahoo.com>
```
earlephilhower, line 10: Want to add your name to the comment history since you've added the MCLK stuff?

libraries/I2S/src/I2S.cpp
```diff
@@ -19,7 +19,7 @@
     Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 #include <Arduino.h>
-#include "I2S.h"
+#include <I2S.h>
```
earlephilhower, line 22: Why the change from `"I2S.h"` to `<I2S.h>` ?  We want to only include the header in the same directory as this source file, so `""`s are appropriate here.

earlephilhower (changes requested):
Very nice work and thanks for the effort in packaging it up.  Some really minor niggles and I think we're all set!

earlephilhower:
Oh, and double thanks for actually adding documentation for your changes!!!

docs/i2s.rst
LinusHeu: Imho it would be nice if the documentation mentioned, that this is only necessary when using MCLK. Or does this also have a positive effect when using I2S without MCLK?

palmerr23:
Thanks Earle and Linus,

I'll get to all the minor issues - in which you are both correct in all cases. 

It might take a few days, as my fork seems to be out of date with the base branch. I rarely use git in community contribution mode and I'm not sure how to bring my fork into line without overwriting all my changes, as it did last time when I hit the "update branch" button. I'd rather abandon the pull than go through that pain again!

For the one below, I was intending to release the pin, rather than set it to false. My bad, not having written a lot of PIO code. Clearly, once the PIO code abort, it will no longer try to change the pin value, and PIO to pin mappings are not exclusive.

```
void I2S::end() {
    if (_MCLKenabled && _running) {
        pio_sm_set_enabled(_pioMCLK, _smMCLK, false);
        _MCLKenabled = false;
```

Richard

earlephilhower:
Don't worry about the update, just do your changes and push.  The online `update branch` button has never given me any trouble (and if there are any conflicts they can be edited online by me if needed).

FWIW I normally just do the `update branch` from the web interface and then just `git pull` while in my editing branch and things generally "just work."

palmerr23:
:-)

Thanks

   _________________________
   Richard Palmer

 0417 514 473

palmerr23:
I've made all the changes suggested, with the exception of removing #include <pico/stdlib.h> which is required for the call to set_sys_clock_khz().

The doco file is full of CR/LF sequences, however you don't seem to care about these. All other changes have been restyled.

earlephilhower (approved):

earlephilhower:
Thanks, looks good now!

I fixed some merge weirdness (had `<<<old ..... >>> new .... ====` section) and the `I2S::end()` lost the MCLK stop bit completely.  Re-added that and `delete`d the PIOProgram to avoid a memory leak.  Other very minor things scattered around hence the flurry of small change emails you probably just received.

earlephilhower:
Thanks again!  Know it was a long slog, but we do really appreciate the efforts!

LinusHeu:
I think #1548 is still/again accidentally overwritten and should be restored

earlephilhower:
Huh, did not notice that.  Good catch!  See #1582 .

PR #1623, opened 2023-08-01
Add new board Silicognition RP2040-Shim

xorbit:
Board documentation can be found here:
https://silicognition.com/Products/rp2040-shim/


earlephilhower (approved):
LGTM, thanks!

Issue #1620, opened 2023-07-31
USB Serial seems not to send data in some weird cases


marklinmax:
I am running a rather simple code on my raspberry pi pico (RP2040), and testing the communication with the Arduino IDE serial monitor is working great.
Problem is, when I run the communication in the actual system, I can receive data sent to the pico, but the data sent from the pico seems lost somewhere in its entirety.

The target for communication is an application on which I have zero control over, so I did debugging by sniffing the COM port it is connected to with another software.

Funny thing is, the exact same code runs just fine on an Arduino UNO, so I guess this is an issue related to either how the Serial is implemented using the USB CDC library, or the USB CDC library itself (but I find it hard to believe).

maxgerhardt:
Does the program in your actual system set DTR = 1? This virtual data-terminal-ready line needs to be asserted, the TinyUSB CDC implementation checks for this.

marklinmax:
> Does the program in your actual system set DTR = 1 in the COM settings? This virtual data-terminal-ready line needs to be asserted, the TinyUSB CDC implementation checks for this.
> 
> Specifically, I'm talking about
> 
> https://github.com/adafruit/Adafruit_TinyUSB_Arduino/blob/4a93c7d869852c03b198d74cefe87de4ba071eca/src/class/cdc/cdc_device.h#L168-L171
> 
> https://github.com/adafruit/Adafruit_TinyUSB_Arduino/blob/4a93c7d869852c03b198d74cefe87de4ba071eca/src/class/cdc/cdc_device.c#L113-L117
> 
> https://github.com/earlephilhower/arduino-pico/blob/3cc5ac14ff5ab9fd2d0399c8d4a99c3451c5e818/cores/rp2040/SerialUSB.cpp#L125-L133

Seems like that was the case... I simply bypassed the _tud_cdc_connected()_ test in SerialUSB.cpp and now it works like a charm.

Maybe there should be an option to ignore it, as it seems some weird implementations on the application side can be found and will mess with writing on the com port.

Also, this confirms that Arduino implementation just does not care about it either.

Thanks a lot, I would never have found this one alone.

maxgerhardt:
I see, then my guess was right. (Same issue as in https://github.com/earlephilhower/arduino-pico/issues/718)

[...]

PR #1624, opened 2023-08-02
Add SerialUSB::ignoreFlowControl() method

marklinmax:
In some cases, the target application will not assert the DTR virtual line, thus preventing writing operations to succeed. 

For this reason, the SerialUSB::ignoreFlowControl() method disables the connection's state verification, enabling the program to write on the port, even though the data might be lost.

This is in response to #1620


earlephilhower (approved):
Thanks, seems fine.  We'll probably want to update the docs too, but that can be done in the editor online after a merge.

PR #1641, opened 2023-08-09
SPI debugging: tiny fix & log actual baudrate

LinusHeu:
1. I think in the first log line, you want to log the parameters with which beginTransaction() got called, and not the old settings. (_spis are the previous settings)
2. I wanted to see which baudrate was actually set. The compiler optimizes `actualBauds` away if logging is disabled, right?

### DIFF
--- a/libraries/SPI/src/SPI.cpp
+++ b/libraries/SPI/src/SPI.cpp
@@ -187,8 +187,8 @@ void SPIClassRP2040::beginTransaction(SPISettings settings) {
             spi_deinit(_spi);
         }
         DEBUGSPI("SPI: initting SPI\n");
-        uint actualBauds = spi_init(_spi, _spis.getClockFreq());
-        DEBUGSPI("SPI: actual baudrate=%u\n", actualBauds);
+        spi_init(_spi, _spis.getClockFreq());
+        DEBUGSPI("SPI: actual baudrate=%u\n", spi_get_baudrate(_spi));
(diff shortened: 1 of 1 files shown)


LinusHeu:
Oh, of course it gives an unused variable warning when logging is disabled. So much for "the compiler will optimize that away" 😆 
Should I do it like this then?
```cpp
#ifdef DEBUG_RP2040_SPI
uint actualBauds = spi_init(_spi, _spis.getClockFreq());
DEBUGSPI("SPI: actual baudrate=%u\n", actualBauds);
#else
spi_init(_spi, _spis.getClockFreq());
#endif 
```

And I really need to get WSL working so I can run run_CI_locally.sh

earlephilhower:
Thanks.  You're right, this debug stuff was from before some optimizations I did which rendered the values stale.

> And I really need to get WSL working so I can run run_CI_locally.sh

Please, don't even try.  I don't think I ever got it running since GH has been so fast for its runs.  It's something I stole from our ESP8266 repo.

Why not:
````
DEBUGSPI("SPI: actual baudrate=%u\n", spi_get_baudrate(_spi));
````
?

LinusHeu:
Hehe, then I will just use your GitHub actions. :D

Thanks, that definitely looks way nicer. I didn't know that function existed and didn't check the sdk docs last night.

earlephilhower (approved):

PR #1651, opened 2023-08-20
Make it compile with IPv6 enabled

HeMan:
ip_addr_t has different structure with LWIP_IPV6 enabled. This fix makes it compile, but doesn't enable MDNS over IPv6.


earlephilhower (approved):
Thanks!

PR #1657, opened 2023-08-21
Update to new GCC 10.5, GDB 10, OpenOC 0.12

earlephilhower:
New toolchain build


earlephilhower:
Since it seems MPFR and 32-bit code is having a problem and taking up a lot of time, I think I'll bite the bullet and get to the latest GCC 12 instead of spending the effort on a 10.5 release.  Closing.

PR #1659, opened 2023-08-22
Hopefully fix I2S #1656

### LINKED ISSUE #1656: Sometimes I2S doesn't properly start
I'm really pulling my hair out here.
I2S seemed to work reliably when I used just this core and played audio from an array in flash. But I want to use Arduino-Audio-Tools to play wav files from an SD-card (and ideally add more features later) but that doesn't seem to work.

### Symptom:
According to the Logs of Arduino-Audio-Tools it starts I2S successfully, then it writes audio data to the output buffer. But it gets stuck after [Size of I2S output buffer] bytes are written.
No sound is audible.
The MCU doesn't crash. (In another test, I had the onboard LED blink on core1, and it never stopped.)

With the Debugger, I can see that it gets stuck here forever (Arduino-Audio-Tools uses blocking write to I2S):
https://github.com/earlephilhower/arduino-pico/blob/f11cc4db69b91410a7904adf9eefa4a5650ff115/libraries/AudioBufferManager/src/AudioBufferManager.cpp#L174

It looks to me like the I2S output buffer is not played. It just gets filled and never gets emptied.

I added `digitalWrite(LED_BUILTIN, pio_sm_is_tx_fifo_empty(_pio,_sm));` to the end of I2S::begin()
https://github.com/earlephilhower/arduino-pico/blob/9352d2d14bbbae1849234207ee98065a35199ac6/libraries/I2S/src/I2S.cpp#L238-L243
The LED turns on every time the bug occurs, and stays off every time it works. So I think there is some issue with this core/I2S. The PIO state machines don't get fed any data? Even directly after I2S::begin() they should play silence, right? I assume this is "proof" that the issue occurs somewhere in the setup, before any data is actually written to the I2S output buffer.

### MCVE

The MCVE below reproduces it 100% of the time in my testing today. While changing logging changes the bug, I don't think it's actually directly related to logging. In another sketch (where I first encountered the bug) it was pretty random but *felt* like it occurred *maybe* more often when logging was off and less often when logging was enabled.

Very sorry, the MCVE is a bit large for my taste. I really hope you can still reproduce it and help me. The last few days I only had sketches where the bug occurred randomly, so I'm really happy to be able to reproduce it at all. 
It uses Arduino-Audio-Tools to play a wav file from the SD-card. The driver between Arduino-Audio-Tools and this core's I2S is in [AudioI2S/I2SRP2040.h ](https://github.com/LinusHeu/arduino-audio-tools/blob/i2sbug/src/AudioI2S/I2SRP2040.h)

main.cpp
```ino
#include <Arduino.h>
#include <AudioTools.h>
#include <SDFS.h>


I2SStream i2s;
EncodedAudioStream decoder(&i2s, new WAVDecoder());
StreamCopy copier;
File audioFile;

void setup() {
  pinMode(LED_BUILTIN, OUTPUT);
  Serial.begin();
  while (!Serial)
    ;
  delay(1000);
  Serial.println("Hallo");

  SDFSConfig sdconf;
  sdconf.setCSPin(17);
  sdconf.setSPISpeed(25000000);
  SPI.setRX(16);
  SPI.setTX(19);
  SPI.setSCK(18);
  SDFS.setConfig(sdconf);
  Serial.println(SDFS.begin());

  Serial.print("SDFS.exists(\"CExplorerK.wav\"): ");
  Serial.println(SDFS.exists("CExplorerK.wav"));
  Serial.println("Let's go");

  audioFile = SDFS.open("CExplorerK.wav", "r");

  //AudioLogger::instance().begin(Serial, AudioLogger::Debug);  //Bug doesn't occur. But there is some underflow because of the logging
  AudioLogger::instance().begin(Serial, AudioLogger::Info); //Bug occurs
  //AudioLogger::instance().begin(Serial, AudioLogger::Error); //Bug doesn't occur

  I2SConfig i2sconfig = i2s.defaultConfig(TX_MODE);
  i2sconfig.buffer_count = 5;
  i2sconfig.buffer_size = 512;
  i2sconfig.pin_bck = 27;
  i2sconfig.pin_data = 26;
  i2sconfig.pin_ws = 28;
  i2s.begin(i2sconfig);

  decoder.begin();

  copier.begin(decoder, audioFile);
}

void loop() {
  if (!copier.copy()) {
    Serial.println("ENDED SUCCESSFULLY");
    i2s.end();
	  stop();
  }
}
```
platformio.ini:
```ini
[env:rpipico]
board = rpipico
framework = arduino
board_build.core = earlephilhower
platform_packages = framework-arduinopico@symlink://C:\Users\linus\Projekte\arduino-pico

debug_build_flags = -ggdb3 -g3

debug_tool = cmsis-dap
;upload_protocol = cmsis-dap
debug_speed = 30000 #default debug speed is only 1000kHz

monitor_echo = true
monitor_filters = log2file
;monitor_speed = 115200
```

Version of Arduino-Pico: https://github.com/LinusHeu/arduino-pico/tree/i2sBug (only 2 commits old and the debug LED added that I mentioned above)
Version of Arduino-Audio-Tools: https://github.com/LinusHeu/arduino-audio-tools/tree/i2sbug (git cloned to /lib/) (should be the same as the latest main version of philschatzman)
Adafruit I2S Amp: https://www.adafruit.com/product/3006
Adafruit Micro-SD breakout: https://www.adafruit.com/product/4682
Micro-SD: Sandisk Extreme Pro 32GB
Only file on the card: [sdcardsong.zip](https://github.com/earlephilhower/arduino-pico/files/12398636/sdcardsong.zip)
Raspberry Pi Pico

LinusHeu:
Closes #1656 (maybe)
At least my tests, that sometimes or always triggered the bug, seem to work reliably with this change.

**But I don't understand why.**
So I'm not sure if this actually fixes the bug or just changed something, so I can't reproduce it any more. I would really appreciate it if you could look at this closely.

It looks like [dma_channel_cleanup()](https://github.com/raspberrypi/pico-sdk/blob/6a7db34ff63345a7badec79ebea3aaef1712f374/src/rp2_common/hardware_dma/dma.c#L73-L83) does (almost) the same as what was done in ~AudioBufferManager() except it also disables chain_to:
```cpp
    // Disable CHAIN_TO, and disable channel, so that it ignores any further triggers 
    hw_write_masked( &dma_hw->ch[channel].al1_ctrl, (channel << DMA_CH0_CTRL_TRIG_CHAIN_TO_LSB) | (0u << DMA_CH0_CTRL_TRIG_EN_LSB), DMA_CH0_CTRL_TRIG_CHAIN_TO_BITS | DMA_CH0_CTRL_TRIG_EN_BITS );
```
And this looks like it does exactly the same as `dma_channel_acknowledge_irq0()` except for both irq0 and irq1, but I don't see why that would matter.
```cpp
    // finally clear the IRQ status, which may have been set during abort
    dma_hw->intr = 1u << channel;
```

### DIFF
--- a/libraries/I2S/src/I2S.cpp
+++ b/libraries/I2S/src/I2S.cpp
@@ -237,8 +237,6 @@ bool I2S::begin() {
     }
     _arb->setCallback(_cb);
     pio_sm_set_enabled(_pio, _sm, true);
-    digitalWrite(LED_BUILTIN, pio_sm_is_tx_fifo_empty(_pio,_sm));
-    //delay(1000);
     return true;
 }
 
--- a/libraries/I2S/src/I2S.cpp
+++ b/libraries/I2S/src/I2S.cpp
@@ -363,6 +363,8 @@ size_t I2S::write(int32_t val, bool sync) {
     if (!_running || !_isOutput) {
         return 0;
     }
+    //Serial.print("pio_sm_is_tx_fifo_empty(_pio,_sm): ");
+    Serial.println(pio_sm_is_tx_fifo_empty(_pio,_sm));
     return _arb->write(val, sync);
 }
 
--- a/libraries/I2S/src/I2S.cpp
+++ b/libraries/I2S/src/I2S.cpp
@@ -237,7 +237,8 @@ bool I2S::begin() {
     }
     _arb->setCallback(_cb);
     pio_sm_set_enabled(_pio, _sm, true);
-
+    digitalWrite(LED_BUILTIN, pio_sm_is_tx_fifo_empty(_pio,_sm));
+    //delay(1000);
     return true;
 }
 
@@ -364,7 +365,8 @@ size_t I2S::write(int32_t val, bool sync) {
         return 0;
     }
     //Serial.print("pio_sm_is_tx_fifo_empty(_pio,_sm): ");
-    Serial.println(pio_sm_is_tx_fifo_empty(_pio,_sm));
+    //Serial.println(pio_sm_is_tx_fifo_empty(_pio,_sm));
+    //digitalWrite(LED_BUILTIN, pio_sm_is_tx_fifo_empty(_pio,_sm));
     return _arb->write(val, sync);
 }
 
--- a/libraries/AudioBufferManager/src/AudioBufferManager.cpp
+++ b/libraries/AudioBufferManager/src/AudioBufferManager.cpp
@@ -73,11 +73,9 @@ AudioBufferManager::~AudioBufferManager() {
     if (_running) {
         _running = false;
         for (auto i = 0; i < 2; i++) {
-            dma_channel_set_irq0_enabled(_channelDMA[i], false);
+            dma_channel_cleanup(_channelDMA[i]);
             __channelMap[_channelDMA[i]] = nullptr;
-            dma_channel_abort(_channelDMA[i]);
             dma_channel_unclaim(_channelDMA[i]);
-            dma_channel_acknowledge_irq0(_channelDMA[i]);
             __channelCount--;
         }
         if (!__channelCount) {


libraries/I2S/src/I2S.cpp
```diff
@@ -363,6 +362,9 @@ size_t I2S::write(int32_t val, bool sync) {
     if (!_running || !_isOutput) {
         return 0;
     }
+    //Serial.print("pio_sm_is_tx_fifo_empty(_pio,_sm): ");
```
earlephilhower, line 365: Can we blow this leftover debug away, please?

earlephilhower (changes requested):

earlephilhower:
I think your change is fine and maybe it's catching a race condition the original code missed.  At the worst, it removes several lines of my own code which means several lines which can't have any bugs in them. :)

libraries/I2S/src/I2S.cpp
LinusHeu: oh yes of course. Sorry, I missed that

PR #1673, opened 2023-08-25
Add TDM support to I2S

### LINKED ISSUE #1066: Feature request: Add TDM mode for I2S
Codecs (with more than 2 channels use TDM (Time Domain Multiplex) to provide up to 8 Channels of Inputs or Outputs per I2S port. For instance, the CS42448 Codec has 6 Inputs and 8 Inputs which can be used using one single I2S interface.

Adding TDM support would add a big range of new possibilities.

earlephilhower:
Fixes #1066

Implements a simple TDM mode for the I2S output object.


earlephilhower:
@DatanoiseTV give this a try.  On my crappy logic analyzer it seems to be working fine, but I don't have any actual codecs that can eat TDM mode.

DatanoiseTV:
> @DatanoiseTV give this a try. On my crappy logic analyzer it seems to be working fine, but I don't have any actual codecs that can eat TDM mode.

Does it also support TDM512 input? I just happen to have a board with a TDM512 ADC here to test, if yes.

PR #1685, opened 2023-08-31
Ensure HID reports aren't dropped

### LINKED ISSUE #1682: USB - HID - loosing reports
dear @earlephilhower ,

we experienced lost USB - HID reports when sending at a high rate or multiple devices at once (e.g. sending a keyboard & a mouse report at the same time).

I think it is related to 
https://github.com/earlephilhower/arduino-pico/blob/d0ac7f06b1fa75a4c098be3da543912e42a66d17/libraries/Mouse/src/Mouse.cpp#L48C34-L48C34

where the report is not sent when HID is is not ready.

Would it be feasible to replace the `if()` with a `while(!)`?
(maybe with an additional counter / us delay for a timeout)

Here is an example sketch, which triggers this error (warning: key "A" will be sticky)
[hid_test.txt](https://github.com/earlephilhower/arduino-pico/files/12484221/hid_test.txt)

earlephilhower:
Fixes #1682

Make the HID report wait up to 500ms for an existing one to go out before giving up sending a report.

### DIFF
--- a/cores/rp2040/RP2040USB.cpp
+++ b/cores/rp2040/RP2040USB.cpp
@@ -394,6 +394,18 @@ void __USBStart() {
 }
 
 
+bool __USBHIDReady() {
+    uint32_t start = millis();
+    const uint32_t timeout = 500;
+
+    while (((millis() - start) < timeout) && tud_ready() && !tud_hid_ready()) {
+        tud_task();
+        delay(1);
+    }
+    return tud_hid_ready();
+}
+
+
 // Invoked when received GET_REPORT control request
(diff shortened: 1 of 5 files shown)


meoinhan:
Dear Dev. I was a new person I tested it and it seems to work properly. But there are few errors that I see when I use mouse.cpp It's possible that mouse.move() doesn't work well and has a small bug (perhaps because the delay is 500 so it gives an error. I tried lowering it to 10 and mouse.move doesn't have any bugs anymore). Although I haven't fixed this before, I got errors in mouse.press and mouse.release. (it's a bit slow)

PR #1688, opened 2023-09-03
Fix gcc search path in make-ota.sh

oddstr13:
It seems it was using the system wide install I had in the past.

This change makes the paths consistent with `make-libpico.sh`
https://github.com/earlephilhower/arduino-pico/blob/0b56452c353e4fc5c8c0bd8e4383c2dc2d0a49c6/tools/libpico/make-libpico.sh#L5-L6


earlephilhower (approved):
Thanks, good catch!  I should probably rebuild OTA w/the new GCC 12.3...

PR #1706, opened 2023-09-12
GDB/Binutils release-gdb-13.2, GCC still 12.3

### LINKED ISSUE #1681: Update P.IO toolchain references to GitHub release
>  Absolutely, we can have reference abitrary direct download links to a `.zip` or `.tar.gz` instead (if it already has a `package.json` in), same for referencing repos (or even different branches of repos). I already do that in stuff like [platform-ch32v](https://github.com/Community-PIO-CH32V/platform-ch32v/blob/60ac59ab7c37cff8d73ae3d3bdd9f288d9372ba3/platform.py#L56-L63). That's definitely a thing we can pull in at the next platform update, but it should not be causing this specific issue.

_Originally posted by @maxgerhardt in https://github.com/earlephilhower/arduino-pico/discussions/1647#discussioncomment-6855328_

earlephilhower:
May fix #1681 due to breakage of GDB 12.3 in Platform.IO


earlephilhower:
GDB debugging and building some USB examples worked fine w/the updated binutils and GDB.  Note that a new linker option is required because the binutils adds additional warning messages which really aren't applicable to an embedded system.

earlephilhower:
PIO build failed because it's still pointing to 10.2 binaries which do not have this new flag.  Should fix when the PIO integration moves to 12.x+GDB13.2

PR #1714, opened 2023-09-16
Support IRQ disabled SPI transactions

### LINKED ISSUE #1147: SPI transactions
Are there any plans to implement SPI transactions in the future ?

It is very useful when there are multiple devices on an SPI bus and SPI comms will take place in an interrupt handler. Without this functionality, SPI comms in an ISR will corrupt in-progress SPI comms with another device when an interrupt fires.

For example, using the MCP2515 CAN bus controller, it is helpful to destage incoming messages from the chip as soon as it asserts the interrupt. Waiting until later (e.g. setting a flag) may result in missed messages at high traffic rates. Continually polling the chip for new messages seems wasteful. It is correctly implemented in the standard AVR core.

As understand it, in other cores, SPI.beginTransaction defers the chosen interrupt until SPI.endTransaction is called. Provided the interrupt remains asserted, the ISR will then run.

For my current projects, I isolate this device onto one SPI bus, with other non-interrupting SPI devices on the other bus.

Thanks.

earlephilhower:
Fixes #1147

When SPI.beginTrqansaction() is called, disable all GPIO IRQs that were registered using SPI.usingInterrupt().  On SPI.endTransaction(), restore all the IRQs to their prior state.

### DIFF
--- a/libraries/SPI/keywords.txt
+++ b/libraries/SPI/keywords.txt
@@ -14,6 +14,9 @@ SPI1	KEYWORD1
 #######################################
 begin	KEYWORD2
 end	KEYWORD2
+beginTransaction	KEYWORD2
+endTransaction	KEYWORD2
+SPISettings	KEYWORD2
 transfer	KEYWORD2
 setBitOrder	KEYWORD2
 setDataMode	KEYWORD2
--- a/libraries/SPI/src/SPI.cpp
+++ b/libraries/SPI/src/SPI.cpp
@@ -21,6 +21,8 @@
 #include "SPI.h"
 #include <hardware/spi.h>
 #include <hardware/gpio.h>
+#include <hardware/structs/iobank0.h>
+#include <hardware/irq.h>
 
 #ifdef USE_TINYUSB
 // For Serial when selecting TinyUSB.  Can't include in the core because Arduino IDE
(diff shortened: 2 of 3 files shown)


earlephilhower:
@obdevel, know it's been a while but if you still have a need for SPI transactions, could you try this PR out.   I don't have a good test case since it needs specific HW and timing to hit the problems you had with re-entrancy.

earlephilhower:
Simple testing w/DEBUGSPI enabled shows the latest push disabling IRQs on the proper pins and then re-enabling them later.  Also adds proper keyword hilighting.
````
#include <SPI.h>

static int irqs = 0;
void irq() {
  irqs++;
}


void setup() {
  delay(5000);
  Serial.printf("start\n");
  pinMode(0, INPUT_PULLUP);
  pinMode(1, INPUT_PULLUP);
  attachInterrupt(0, irq, LOW);
  attachInterrupt(1, irq, RISING);
  SPI.usingInterrupt(1);
  SPISettings s;
  SPI.begin();
  Serial.printf("done, pull down\n");
  delay(5000);
  Serial.printf("%d\n", irqs);
  Serial.printf("spi beginx, pull down\n");
  SPI.beginTransaction(s);
  delay(5000);
  Serial.printf("%d\n", irqs);
  SPI.endTransaction();
  Serial.printf("spi end, pull down\n");
  delay(5000);
  Serial.printf("%d\n", irqs);

}

void loop() {
Serial.printf("%d\n", irqs);
delay(1000);
}
````

**Comment by obdevel** [comment]
> @obdevel, know it's been a while but if you still have a need for SPI transactions, could you try this PR out. I don't have a good test case since it needs specific HW and timing to hit the problems you had with re-entrancy.

Yes, thank you, I will. It'll take a couple of days to setup a test environment. I can test at up to 1000 interrupts per second.

The original requirement has somewhat gone away. We previously needed to talk to an MCP2515 SPI CAN controller chip. However, there is now an excellent PIO implementation of a CAN controller (https://github.com/KevinOConnor/can2040) which works very well. I created an Arduino wrapper library for it which is now our default arrangement.

However, this addition will be very useful for other use-cases.

**Comment by earlephilhower** [comment]
Thanks in advance if you can do it, and no rush of course.

BTW, please use `git master` or include both #1714 and #1715 (just thought of one final race condition I couldn't test for w/my hand-shorting pins).

**Comment by LinusHeu** [comment]
Just curious, why do you use map? Wouldn't it be better to just use uint32_t[4] mask and

**Comment by LinusHeu** [comment]
I'm just curious. Why are you using map? I always hear that dynamic memory allocation and stl should be avoided. Wouldn't it be better to use a mask and int(s) to save the state like the AVR core? 
I bet it should be possible to use 4 uint32_t each for the mask and storing the state.

**Comment by earlephilhower** [comment]
The STL is  simple to use and tested by lots of other people are my main reasons.

Memory allocations for the map do happen, yes, but in all of the cases it's max 2 (we only have 2 SPI devices) and they're all done at the beginning of the app, not in an ISR, and only ~12(16?) bytes a piece so it's not a big deal with 256K of RAM.  They're also forever and at the start of the heap and not somewhere in the middle so no impact on allocations later on.

On an 8-bit AVR with 2KB of RAM the choice would probably be different of course!

**Comment by LinusHeu** [comment]
I see. Thank you :D

**Comment by obdevel** [comment]
I'm having a problem with python not being found for the uf2conv step. Is there are quick fix or shall I hack platform.txt ?
I'm running on macOS. I cloned the repo and placed a symlink in the appropriate place.

```
java.io.IOException: Cannot run program "/Users/xxx/Library/Arduino15/packages/rp2040/hardware/rp2040/3.5.0/system/python3/python3": error=2, No such file or directory
```

earlephilhower:
I don't use a Mac, but if you have Python installed please run tools/get.py to download  the whole tool chain.  Also, as you're probably aware, this core need Rosetta because I can't build for the M1 yet.

Issue #1711, opened 2023-09-15
Platformio: Debugging doesn't start if no port is specified & zadig still required & debugging doesn't start at all


LinusHeu:
I suspect this is since 3.5.0 but I can't be sure because I also switched PCs.
There are three issues with debugging. I'm putting them into one issue because I think they might be related.
### 1. IIRC previously, I didn't have to specify a port for the picoprobe, now I have to. If no debug_port is specified debugging fails to start without any error message.

If in platformio.ini, only the debug_tool is defined:
```ini
debug_tool = cmsis-dap
```
1. Click on Start debugging
3. A loading animation appears in the debug window, compilation starts in the terminal
4. Compilation is successful
5. Shortly after, the loading animation stops and nothing happens. No error message anywhere. Debug console is empty.
This is the same no matter if I have nothing, only the picoprobe, or picoprobe and Pico connected.

If in platformio.ini debug_port is defined:
```ini
debug_tool = cmsis-dap
debug_port = COM3
```
OpenOCD starts.
It doesn't matter what debug_port I define. `debug_port = COM3` does the same as `debug_port = COM99` or even `debug_port = YoThisSeemsLikeABug` or `debug_port = 😂`.

### 2. Why do I still have to fiddle with Zadig to install the WinUSB driver for the picoprobe?
By default, it's usbser, but with that driver I get the `Error: unable to find a matching CMSIS-DAP device` error message in the Debug console. 
The changelog for Picoprobe 1.0.1 claims "Obviates the need to use Zadig to load WinUSB for the debug interface" and I'm using the latest 1.0.3 uf2. 
With Zadig and WinUSB I get one step further to problem 3.

### 3. With 1 and 2 worked around, debugging still doesn't start.
It compiles successfully, outputs the following in the debug console, the loading animation in the debug window ends and nothing more happens.
```
undefinedOpen On-Chip Debugger 0.11.0-g228ede43d-dirty (2022-12-22-20:10)
Licensed under GNU GPL v2
For bug reports, read
	http://openocd.org/doc/doxygen/bugs.html
Info : auto-selecting first available session transport "swd". To override use 'transport select <transport>'.
2
Info : Hardware thread awareness created
Info : RP2040 Flash Bank Command
adapter speed: 1000 kHz
Info : Listening on port 6666 for tcl connections
Info : Listening on port 4444 for telnet connections
[...]

PR #1727, opened 2023-09-22
Check add'l binaries (GDB, OpenOCD, Picotool) run

earlephilhower:
For Mac and Windows builds, try to run the tools packaged in the pico-quick-toolchain release.  This should catch cases like seen in #1711


earlephilhower:
Closing for now.  Need to investigate Powershell executable invocation and paths

PR #1731, opened 2023-09-24
FAT16 library, expose internal flash as USB drive

maxgerhardt:
Built on the great library http://elm-chan.org/fsw/ff/, I just had to add the raw "read/write from flash" and some miscellenous info routines to get this raw library working.

I then used the existing Adafruit USB MSC library to expose read/write access via USB, and voilá, we get a nicely read/writable FAT16 filesystem on the Pico that is 1:1 accessible via USB :) 

grafik

Additionally now working on making this a standard "FS.h" compatible library with all the goodies (open, read, write, close for files and directories, timestamps etc.)

While working on this I also noticed that at least for PlatformIO the `include/tusb_config.h` file seemed not actually used (you can place a `#error` in there and it will still compile), which is due to the include order in PlatformIO being that the original Adafruit TinyUSB library's `tusb_config.h` is first, and then the `include/` for Arduino-Pico follows. Resolved by making use of the `CFG_TUSB_CONFIG_FILE` macro and pointing it to the uniquely named config file.

This is still in progress for reading / iterating through directories and some other stuff, so not yet merge ready, just showing that I'm working on it and that this would replace [SingleDriveFile](https://github.com/earlephilhower/arduino-pico/tree/master/libraries/SingleFileDrive) that does some LittleFS -> FAT16 translation for a single, read-only file, much better, by just formatting it in FAT16 natively.


lyusupov:
@maxgerhardt 
this feature is already included in the **Adafruit** variants of:
* [Adafruit_SPIFlash](https://github.com/adafruit/Adafruit_SPIFlash)  and
* Adafruit [**SdFat**](https://github.com/adafruit/SdFat) libraries

This is an example of usage:
https://github.com/lyusupov/SoftRF/blob/master/software/firmware/source/SoftRF/src/platform/RP2040.cpp#L159

When the Adafruit **SdFat** library is deployed into the _\<HOME>/Arduino/libraries_ folder it gets priority over Earle's Arduino Core built-in one.

maxgerhardt:
Hmmm I see. So that one also seems to have all the file system ops that LittleFS has (.begin(), .open(), ..) and it does it on 512 byte sector instead of 4096. 

So the problem is that `ESP8266SdFat` doesn't have stuff that Adafruit SdFat has, too?

lyusupov:
Both Earle's and Adafruit SdFat libraries are derivatives of [Greiman's](https://github.com/greiman/SdFat) one.
IMHO, it would be best if someone could transfer sensitive Adafruit  changes into the Earle's port.
In this case the [Adafruit_SPIFlash](https://github.com/adafruit/Adafruit_SPIFlash) and TinyUSB Mass Storage  example could run on top of the RP2040 built-in SdFat library with no need to install an external one.

earlephilhower:
Interesting work!

FWIW, the ESP8266SdFat lib here already has FAT16/32 support:

https://github.com/earlephilhower/ESP8266SdFat/blob/255180be140d252605f28667078201fba588b968/src/SdFatConfig.h#L36

https://github.com/earlephilhower/ESP8266SdFat/blob/255180be140d252605f28667078201fba588b968/src/SdFatConfig.h#L64-L71

But AFAIK the upstream SdFat tightly integrates the SDIO part of the library with the FAT portion.  I suppose if there's a way to remove the transfer backend and read/write flash directly instead of the SD protocol that would be all that's needed.

The Adafruit mass storage example, AIUI, simply exports the flash (or DRAM) as a small USB stick.  There's no filesystem support whatsoever.  Your PC could format it as FAT16, XFS, or even NTFS, since the controller doesn't actually ever parse the flash into something that an app could use to get a `File`.

On the 8266 we found lots of knockoff flash devices with microscopic write lifetimes (but OK for a normal write-once and then run forever use case).  FAT rewrites the 2 FAT copies at fixed sectors on every file append (or will do similar to the directory entry fixed sector).  So you could wear out the flash super-fast.  I believe on the ESP32 they include a simple wear-leveling layer between the flash and the FS and hence can run FAT on flash w/o this problem.

lyusupov:
Most typical use case that Adafruit have designated the SPIFlash "FAT-in-flash" feature is to store CircuitPython libraries and user program/data files.
The subsystem is widely used on Adafruit built SAMD/nRF52/ESP32/RP2040 products..
We would be able see numerous complains on the Adafruit CircuitPython  forums if the technology would not work.

earlephilhower:
Sure, it can work.  That's also a good very low write use case (vs, say, a data logger that needs to persist data after each sample). And I have no worries about Adafruit boards' flash.  But when you start getting to the XYZZYHOHO random letter branded boards from AX or AMZN who knows what you're going to get!

That said, a simple wear-leveling FTL could extend the life of even junk flash so you could run FAT updates for many multiples of flash PE cycles.  That's the reason for the LittleFS choice here.  The designers started from scratch to build a power fail tolerant, atomic, self wear-leveling filesystem.

In any case, PRs always welcome. :laughing:

traxanos:
ff15. warum wurde ff15 verwendet? wir haben das ganze im prinzip mit board mitteln gemacht.

ok, unser konezept ist etwas anders. wie haben einen eigenen bereich im flash als schreibcache abgezwackt um auch größere mengen an daten zu übertragen. dazu kommt eine virtuell allocation tabelle um die mindest größer von 6MB zu erreichen ohne 6MB auf dem Flash zu reservieren.

PR #1776, opened 2023-10-18
Clear dirty flag after commiting EEPROM to flash

coledeck:
Committing to flash doesn't clear the `_dirty` variable, so if `EEPROM.commit()` is called a second time, it will write the data again to flash even with no changes in the data since the previous commit.


earlephilhower (approved):
Thanks!

coledeck:
What a picky style enforcement.... 4 spaces is not ok!

PR #1777, opened 2023-10-19
Use __attribute((weak)) for _write() and other stubs called by newlib

rlcamp:
Use __attribute((weak)) for _write() and other stubs called by newlib, so that sketch and library code can provide non-stub implementations, as they can do in other Arduino cores


earlephilhower (approved):
I see what you're trying to do.  Would you like to try your hand at a real VFS"?

PR #1779, opened 2023-10-21
add Serial.dtr() and Serial.rts() methods

rlcamp:
Adds some more functionality present in other Arduino cores, allowing more fine-grained control over what to do when the USB host closes the connection or wants to indicate an EOF condition.

(C programmer disclaimer: any resemblance to an actual C++ programmer living or dead is purely coincidental, I have no idea whether the methods being added to the .h file should be virtual and/or override or any other such combination, feedback and mods to the PR welcome)

### DIFF
--- a/cores/rp2040/SerialUSB.cpp
+++ b/cores/rp2040/SerialUSB.cpp
@@ -197,6 +197,14 @@ static void CheckSerialReset() {
     }
 }
 
+bool SerialUSB::dtr() {
+    return _dtr;
+}
+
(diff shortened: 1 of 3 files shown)


earlephilhower (changes requested):
I'm in the middle of a multi week travel, so sorry for the delay 

Could you please add these new method to the Serial.rst documentation since they're rp2040 specific.

earlephilhower (approved):
Thanks!

PR #1780, opened 2023-10-21
RP2040 example for multicore fifo (blocking) with pointers

cegea:
This example shows how to pass pointers between cores using the software fifo blocking functions.

### DIFF
--- a/libraries/rp2040/examples/MulticoreFIFO/MulticoreFIFO.ino
+++ b/libraries/rp2040/examples/MulticoreFIFO/MulticoreFIFO.ino
@@ -14,7 +14,7 @@ void setup() {
   delay(5000);
 }
 
-const char *c0_char= "C0 commands C1"; // Can be declared locally but globally is more reliable.
+const char *c0_char= "C0 commands C1"; 
 void loop() {
   delay(1000);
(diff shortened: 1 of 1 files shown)


earlephilhower (changes requested):
I'm traveling for two weeks so sorry for the delay.  I'm fine with the idea of this example, but you pushed a`const char *` and are C style casting it to a`char *` which isn't really legal.  Can you fix that, please?

cegea:
Sure, no problem. Sorry about the casting (I'm used to C). 
I had to use `reinterpret_cast` for parsing a pointer into its decimal value stored as `uint32_t`. Is this safe enough? Is there any other way to push a pointer that I'm missing?

```
rp2040.fifo.push(reinterpret_cast<uint32_t>(c0_char));
```

I extended the example to push a `const char *`firstly and later any incoming bytes from serial that are stored into a `char *`.  Hope this is ok.

Thank you for this amazing work! (I forgot earlier).

earlephilhower (changes requested):
I'm still traveling so can only look at this via my phone, but isn't this the same as before?  You're casting a const to a non-const.   I was thinking you'd change the receiver to expect a `const char *` and that would take care of things...

PR #1784, opened 2023-10-23
Add Sea-Picro board

ayan4m1:
https://github.com/joshajohnson/sea-picro


ayan4m1:
@maxgerhardt Thank you for the information! I added `board_url` as a parameter to MakeBoard/MakeBoardJson which defaults to the pico URL that is currently default.

variants/sea_picro/pins_arduino.h
```diff
@@ -0,0 +1,41 @@
...
+
+// Not pinned out
+#define PIN_SERIAL2_TX (31u)
+#define PIN_SERIAL2_RX (31u)
+
+// SPI
+#define PIN_SPI0_MISO (20u)
+#define PIN_SPI0_MOSI (23u)
+#define PIN_SPI0_SCK (22u)
+#define PIN_SPI0_SS (31u) // not pinned out
```
maxgerhardt, line 21: But per [pinout diagram](https://github.com/joshajohnson/sea-picro/blob/master/documentation/pinout/sea-picro-top-pinout-rory-render.png), GPIO21 is nCS (=SS)? Would also match SPI0 CSn for GPIO21 per [datasheet](https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf).

variants/sea_picro/pins_arduino.h
```diff
@@ -0,0 +1,41 @@
...
+
+// Not pinned out
+#define PIN_SPI1_MISO (31u)
+#define PIN_SPI1_MOSI (31u)
+#define PIN_SPI1_SCK (31u)
+#define PIN_SPI1_SS (31u)
```
maxgerhardt, line 27: Per [pinout](https://github.com/joshajohnson/sea-picro/blob/master/documentation/pinout/sea-picro-top-pinout-rory-render.png), GPIO12 up to 15 are available, which are SPI1 capable pins (RX/MISO, CSn, SCK, TX/MOSI) per [datasheet](https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf).

variants/sea_picro/pins_arduino.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+// Pin definitions taken from:
+//    https://github.com/joshajohnson/sea-picro/blob/master/documentation/pinout/sea-picro-top-pinout-rory-render.png
+
+// LEDs not pinned out
+#define PIN_LED (31u)
```
maxgerhardt, line 7: While per schematics the board doesn't have a simple on/off LED, it does have a Neopixel on GPIO25. Hence you should add a `PIN_NEOPIXEL` definition like other boards do (like a lot from Adafruit)

https://github.com/earlephilhower/arduino-pico/blob/f5f7267f4459d83652ffdec8d3acf4fb9fe743b1/variants/adafruit_metro/pins_arduino.h#L6-L7

variants/sea_picro/pins_arduino.h
```diff
@@ -0,0 +1,41 @@
...
+
+// Not pinned out
+#define PIN_SPI1_MISO (31u)
+#define PIN_SPI1_MOSI (31u)
+#define PIN_SPI1_SCK (31u)
+#define PIN_SPI1_SS (31u)
+
+// Wire
+#define PIN_WIRE0_SDA (2u)
+#define PIN_WIRE0_SCL (3u)
```
maxgerhardt, line 31: GPIO2 and 3 are **I2C1** pins per [datasheet](https://datasheets.raspberrypi.com/rp2040/rp2040-datasheet.pdf), not I2C0. (Without [special macros](https://github.com/earlephilhower/arduino-pico/blob/f5f7267f4459d83652ffdec8d3acf4fb9fe743b1/variants/adafruit_feather/pins_arduino.h#L32-L39), Wire0 maps to i2c0).

Actual I2C0 pins would e.g. be GPIO 4 and 5, among many others that are pinned out.

maxgerhardt (commented):
Just some questions regarding pinmmappings.

variants/sea_picro/pins_arduino.h
ayan4m1: Even if pin 25 is not actually connected to a NeoPixel? It's meant for driving them, but there is no actual LED on the board.

maxgerhardt: Ah on the "1.0-EXT" variant that's just connected to the `74LVC1T45` driver / level-shifter chip, but not to an actual LED.

But on the "1.0-RST" variant it does show an actual WS2812C-2020 as D3. But this PR is for the "1.0-EXT" variant?

ayan4m1: Correct, for the EXT. Should I indicate that somehow? Preferred formatting?

tools/json/sea_picro.json
```diff
@@ -0,0 +1,56 @@
...
+  },
+  "debug": {
+    "jlink_device": "RP2040_M0_0",
+    "openocd_target": "rp2040.cfg",
+    "svd_path": "rp2040.svd"
+  },
+  "frameworks": [
+    "arduino"
+  ],
+  "name": "RP2040",
```
maxgerhardt, line 35: Some how this ended up as `RP2040` as the board name although it should be `Sea-Picro`? Not sure what's happening in the pyhton script right there

variants/sea_picro/pins_arduino.h
maxgerhardt: If it's a really different board variant I would have created `variants/sea_picro_1_0_ext` and `variants/sea_picro_1_0_rst`, but if it's about the Neopixel driver vs actual Neopixel and they're even on the same pin, I think this is fine too.

ayan4m1: Yeah, the only other difference is in the reset circuit for the RP2040, so I don't think it makes much sense to maintain two copies of everything unless you feel differently.

tools/json/sea_picro.json
ayan4m1: I got vendor and board name confused, I'm using "Generic" for the vendor since this is OSHW and the creator does not seem to have a company.

maxgerhardt (approved):
Looks all good to me now!

earlephilhower (approved):
Thanks for the assist, @maxgerhardt .

PR #1800, opened 2023-11-02
RP2040-Eins board

red-scorp:
Added support of RP2040-Eins, similarly to RP2040-ProMini.
Renamed board files to fit better into project's naming convention.


earlephilhower:
Can you please pull the latest master and rerun makeboards to get things back in sync?  We had two new boards coming in at the same time do the boards.txt file needs a full regeneration.

earlephilhower (changes requested):
Please sync with latest master.

Issue #406, opened 2022-01-02
Support for RAK11300 Board


fabltd:
Hi

Can support be added for this board:

(https://github.com/RAKWireless/WisBlock/tree/master/PlatformIO/RAK11300)

earlephilhower:
That repo looks a little bit of a mess, seems to duplicate a lot of files that are already in both the Arduino.cc MBED core and this one, and even includes a blob for libmbed.a (which isn't used or usable here).  There also doesn't seem to be a public internal schematic for the module, either.  So, it's not something I'd be able to do w/o much hands-on with with an actual module and a lot of time.

However, if you'd like to give it a go I'd be happy to include it in the core support!  Good luck!

PR #1802, opened 2023-11-02
Add RAKwireless RAK11300

nmaas87:
Adding a PR for getting the RAKwireless RAK11300 board into the Arudino Pico environment.
The RAK11300 is a combined board, having a RP2040 and an SX1262 LoRa RF chip included.

Information was scare, but I could find what I needed using:
- https://docs.rakwireless.com/Product-Categories/WisDuo/RAK11300-Module/Datasheet/#description
- https://forum.rakwireless.com/t/rak11300-pinout-rp2040-to-sx1262/8414
- https://github.com/RAKWireless/RAK-RP-Arduino/tree/main
- https://github.com/RAKWireless/RAK11300-AT-Command-Firmware/tree/main/RAK11300-AT-Arduino

The original BSP is only available for Arduino mbedOS - and the repo is quite mangled.

The RAK11300 is basically just an RP2040 with some pins for the SPI1 rearranged to better fit the directly connected SX1262 module - and that were basically the changes I included.

If accepted, this should close: https://github.com/earlephilhower/arduino-pico/issues/406

This repo does not contain any examples to use with the LoRa module included and also shall not contain something, as this is part of library for LoRa modules. I am currently trying to get it working with https://github.com/jgromes/RadioLib/issues/864

Best regards and thanks for the awesome work and support of Arduino Pico

Nico

Some additional infos about the whole project in my blog post: https://www.nico-maas.de/?p=2607


variants/rakwireless_rak11300/common.h
```diff
@@ -0,0 +1,76 @@
+#pragma once
```
maxgerhardt, line 1: This is the same as https://github.com/earlephilhower/arduino-pico/blob/master/variants/generic/common.h. New variants should not copy this file into their variant board folder, but just reference it with

https://github.com/earlephilhower/arduino-pico/blob/87f6646e959646ef96f6d7ceb71bdd9ac973395d/variants/adafruit_feather/pins_arduino.h#L44

in their `pins_arduino.h` file.

variants/rakwireless_rak11300/pins_arduino.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+// Pin definitions taken from:
+//    RAK definition: https://github.com/RAKWireless/RAK-RP-Arduino/
+//    RAK datasheet: https://docs.rakwireless.com/Product-Categories/WisDuo/RAK11300-Module/Datasheet/#overview
+//    Internal wiring of SX1262 module: https://forum.rakwireless.com/t/rak11300-pinout-rp2040-to-sx1262/8414/
```
maxgerhardt, line 6: Maybe we can add the pin definitions for SX1262 NRESET, BUSY, DIO1 and ANT_PWR (GPIO25) directly in this file, together with a define for SX1262_SPI = SP1?

tools/makeboards.py
```diff
@@ -414,6 +414,9 @@ def MakeBoardJSON(name, vendor_name, product_name, vid, pid, pwr, boarddefine, f
 MakeBoard("pimoroni_plasma2040", "Pimoroni", "Plasma2040", "0x2e8a", "0x100a", 500, "PIMORONI_PLASMA2040", 2, "boot2_w25q080_2_padded_checksum")
 MakeBoard("pimoroni_tiny2040", "Pimoroni", "Tiny2040", "0x2e8a", "0x100a", 500, "PIMORONI_TINY2040", 2, "boot2_w25q64jv_4_padded_checksum")
 
+# RAKwireless
+MakeBoard("rakwireless_rak11300", "RAKwireless", "RAK11300", "0x2e8a", "0x00c0", 500, "RAKWIRELESS_RAK11300", 2, "boot2_w25q080_2_padded_checksum")
```
maxgerhardt, line 418: If you add `, None, "https://store.rakwireless.com/products/wisduo-lpwan-module-rak11300` in the arugment list, then that URL can show up in the board definition too, instead of just the generic Pico one.

nmaas87: Thanks, I did that, also more closely specified the used flash chip as I took one of this modules apart under a microscope :)

variants/rakwireless_rak11300/pins_arduino.h
nmaas87: Thanks, I did most of that and added some more comments as some of these pins have multiple names and this might be otherwise confusing. I kept SPI1 on SPI1 as it neatly integrates with the RadioLib library.

variants/rakwireless_rak11300/common.h
nmaas87: Thanks, done :)

maxgerhardt (approved):
Looking great!

nmaas87:
Thanks so much 💯

earlephilhower:
Addin again, thanks @maxgerhardt for the assist!

nmaas87:
Thank you so much for merging @earlephilhower and @maxgerhardt for the guidelines and support 👍🏻

PR #1805, opened 2023-11-10
Protect W5500/ENC28J60 isLinked from periodic interrupt

### LINKED ISSUE #1786: Repeated call of W5500's isLinked() crashes (?) program
repeated calls of [isLinked() ](https://github.com/earlephilhower/arduino-pico/blob/master/libraries/lwIP_w5500/src/utility/w5500.h#L85-L87) crashes my programm. (no serials outputs anymore, no led flashing...).

It occures also when I increase the delays betweens the isLinked() calls to 2000ms.

Reproduction steps: Use this simple sketch. Link is not necessary.

```
#include <W5500lwIP.h>

Wiznet5500lwIP eth(1, SPI, -1);
bool toggle = true;
uint32_t cnter = 0;

void setup() {

  pinMode(24, OUTPUT);
  digitalWrite(24, toggle);

  Serial.begin(115200);
  delay(5000);

  SPI.setRX(0);
  SPI.setTX(3);
  SPI.setSCK(2);
  SPI.setCS(1);

  Serial.println("Starting Ethernet port");

  bool r = eth.begin();
  if (!r)
  {
    Serial.println("No wired Ethernet hardware detected. Check pinouts, wiring.");
    while (1) {
      delay(1000);
    }
  }

  while (!eth.connected()) {
    eth.isLinked();
    Serial.println(cnter++);

    toggle = !toggle;
    digitalWrite(24, toggle);
    delay(50);
  }
}

void loop() {
  Serial.print("Link: ");
  Serial.println(eth.isLinked());
  delay(1000);
}
```

earlephilhower:
Fixes #1786

### DIFF
--- a/libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
+++ b/libraries/lwIP_enc28j60/src/utility/enc28j60.cpp
@@ -39,6 +39,7 @@
 #include <stdio.h>
 #include <stdarg.h>
 #include <string.h>
+#include <LwipEthernet.h>
 
 #include "enc28j60.h"
 
@@ -696,6 +697,8 @@ uint16_t ENC28J60::phyread(uint8_t reg) {
 
 bool ENC28J60::isLinked() {
     // ( https://github.com/JAndrassy/EthernetENC/tree/master/src/utility/enc28j60.h )
-
-    return !!(phyread(MACSTAT2) & 0x400);
+    ethernet_arch_lwip_begin();
+    auto ret = !!(phyread(MACSTAT2) & 0x400);
+    ethernet_arch_lwip_end();
+    return ret;
 }
--- a/libraries/lwIP_w5500/src/utility/w5500.h
+++ b/libraries/lwIP_w5500/src/utility/w5500.h
@@ -38,6 +38,7 @@
 #include <stdint.h>
 #include <Arduino.h>
 #include <SPI.h>
+#include <LwipEthernet.h>
 
 class Wiznet5500 {
 public:
@@ -83,7 +84,10 @@ public:
         @return true when physical link is up
     */
     bool isLinked() {
-        return wizphy_getphylink() == PHY_LINK_ON;
+        ethernet_arch_lwip_begin();
+        auto ret = wizphy_getphylink() == PHY_LINK_ON;
+        ethernet_arch_lwip_end();
+        return ret;
     }
 
     /**


earlephilhower:
@Ing-Dom, Can you please try this little change and report back?  I've got a board w/hardwired W5100, which doesn't seem to support link state reporting so wouldn't suffer this issue.  Basically it's just protecting the only unprotected SPI access in the 5500 driver that I can see using the standard calls.

Ing-Dom:
I'll test it today, thanks

Ing-Dom:
so, I can report back:
Now I can call isLinked() without issues.

But:
I call netif_set_link_up and down if I detect a link state change. Main reason for this is DHCP.

But when loading you ko branch, there are no more DHCP request wenn I call netif_set_link_up...

With 3.6.0 it works. But "hangs" after some while.
With "ko" no DHCP requests.

Also: with "ko" branch I don't see any more "prolonged" or "delayed" loops.
I monitor my loop timing and every time distance is >= 7ms, I report this.
with 3.6.0 it looks like this: (about one message every 30s)
```
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (8 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (9 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (36 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (9 >= 7)
Common:                  Warning: The loop took longer than usual (16 >= 7)
Common:                  Warning: The loop took longer than usual (16 >= 7)
Common:                  Warning: The loop took longer than usual (8 >= 7)
Common:                  Warning: The loop took longer than usual (8 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (8 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (9 >= 7)
Common:                  Warning: The loop took longer than usual (8 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
Common:                  Warning: The loop took longer than usual (7 >= 7)
```

Did you change more on the network coding?

earlephilhower:
Thanks for reporting back.

> Now I can call isLinked() without issues.
> Also: with "ko" branch I don't see any more "prolonged" or "delayed" loops. I monitor my loop timing and every time distance is >= 7ms, I report this. with 3.6.0 it looks like this: (about one message every 30s)

Yeah, that sounds good.  Re-entrancy in LWIP/SPI will cause memory corruption eventually, so that would explain the issues you were seeing.  By locking around the chip access it avoids that potential.

> But: I call netif_set_link_up and down if I detect a link state change. Main reason for this is DHCP.
> But when loading you ko branch, there are no more DHCP request wenn I call netif_set_link_up...
> With 3.6.0 it works. But "hangs" after some while. With "ko" no DHCP requests.
> Did you change more on the network coding?

The only thing added here was a small lock/unlock around the SPI chip access to stop the re-entrancy.

In any case, the corruption before might have lead to the DHCP lease info being bonked and the stack auto-resending requests.  Would the more direct `dhcp_start` be a more direct way of getting what you want after link up?  And how do you know there's no DHCP after link up (i.e. are you setting the IP to 0.0.0.0  or something and then link down/up and waiting for a change)?

Also, grabbing a `wireshark` trace would be helpful to see if it's, say, sending a DHCP request but ignoring the response or not sending anything at all.

And maybe more directly, should the object itself be resending a DHCP request on a link state change, as opposed to the app?  (I'm not 100% sure that a link change should cause a DHCP lease to expire...need to see what the norm is here).

In any case, this PR does seem to fix the corruption issue so I'm going to merge it.  If you can get an MCVE then we can open another issue around this to track it's solution.

jonspieg:
Looks like this will solve my issues with isLinked() halting the chip!
How do I get this change in PIO to test?

Thanks

earlephilhower:
I think adding/updating a line in the `ini` is all you need:  ` platform_packages = framework-arduinopico @ https://github.com/earlephilhower/arduino-pico.git`

Issue #1814, opened 2023-11-13
wrong PICO_STACK_SIZE 


traxanos:
hi,

the PICO_STACK_SIZE ist specified with 2048. but in map you used 0x1000 (4096).
This means that the stack addresses cannot be determined correctly with:
> 0x20042000 - PICO_STACK_SIZE 

there is another problem with the dynamic memory allocation for core1. i have no way of determining the current memory address so that i can determine the consumption.
https://github.com/earlephilhower/arduino-pico/blob/master/cores/rp2040/main.cpp#L140-L144

Thats my current helper:
```
uint32_t stackPointer()
{
    uint32_t *sp;
    asm volatile("mov %0, sp"
                 : "=r"(sp));
    return (uint32_t)sp;
}

int freeStackSize()
{
    const uint16_t stackSize = 0x1000; // Bug: PICO_STACK_SIZE deliver wrong size
    const unsigned int sp = stackPointer();
    uint32_t ref = 0x20042000 - stackSize;
    if (rp2040.cpuid() == 1) ref -= stackSize;
    return sp - ref;
}
```

Why don't you solve the problem as follows:

```
multicore_launch_core1_with_stack(main1, (uint32_t *)(0x20042000 - PICO_STACK0_SIZE - PICO_STACK1_SIZE), PICO_STACK1_SIZE);
multicore_launch_core1_with_stack(main1, (uint32_t *)(0x20042000 - PICO_STACK_SIZE - PICO_STACK_SIZE), PICO_STACK_SIZE);
```

then i could continue to determine the stack usage cleanly. which would also not be bad if the functions were to migrate to the RP2040 class

earlephilhower:
The allocation is optional for core1 and it was only to allow some folks who asked for a full 8K for both cores.

The `PICO_STACK_SIZE` macro is correct at 2048 (it's defined as the # of 4-byte words, not a byte count) in single core mode (probably 95%+ of users) or dual-core with allocated stack for core1.

So I'm not really seeing any issue here.  If you want to make a PR that assigns the allocated stack for core1 (only in the case it's enabled) to a global or something, I'd be happy to look at it.

Thx!

traxanos:
well we use core1 and we always have the problem that the stack memory is too small. that is also the reason why we monitor the free size. now i want to enlarge the stack, but i can’t until i know the addresses.

hence my question, why this happens dynamically at runtime and why the size is fixed.
that’s why my suggestion was to find a different or better solution.

[...]

PR #1816, opened 2023-11-14
optimizes core1 stack handling

traxanos:
as discussed in ticket #1814 , a suggestion for optimizing batch processing for core1 incl matching helper functions.
i tried to adopt your naming scheme. but feel free to customize it if you don't like it.
i hope this is as you envisioned it.

### DIFF
--- a/cores/rp2040/RP2040Support.h
+++ b/cores/rp2040/RP2040Support.h
@@ -157,6 +157,10 @@ extern RP2040 rp2040;
 extern "C" void main1();
 extern "C" char __StackLimit;
 extern "C" char __bss_end__;
+extern "C" void setup1() __attribute__((weak));
+extern "C" void loop1() __attribute__((weak));
+extern "C" bool core1_separate_stack;
+extern "C" uint32_t* core1_separate_stack_address;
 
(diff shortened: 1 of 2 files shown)


traxanos:
Is it possible that the global macros do not access main.c in the framwork?

traxanos:
I have also made another variant.
https://github.com/earlephilhower/arduino-pico/compare/master...OpenKNX:arduino-pico:optimizes-core1-stack-alternative

earlephilhower (changes requested):
Thanks, but the problem with the PR here is that it's not usable in the Arduino IDE, only from the P.IO config.  `#define`s can't really be set by the app.

I did like the 2nd version you linked, though.  The formatting will need to match the core's (tools/restyle.sh), but OTW it looked OK to me.

traxanos:
i have now adopted suggestion 2 as mr and hope that now fits. tools/restyle.sh does not exist. and tests/restyle.sh does not run on windows.

earlephilhower (approved):
Thanks!

traxanos:
i have removed the 4 whitespaces for style check

PR #1862, opened 2023-12-01
WiFi and Ethernet - config static IP auto gw,mask,dns as in Arduino libs

JAndrassy:
The 'config' functions with not all parameters specified should use default values for gw IP, subnet mask and DNS server IP. The first Arduino Ethernet library has this (with begin), I guess the first Arduino WiFi library  had it too, but the default values were implemented in the firmware. Then WiFi101 and WiFiNINA are written base on WiFi lib's source code, but without the defaults implemented in firmware. (My PR are waiting to fix that.) The new [WiFiS3](https://github.com/arduino/ArduinoCore-renesas/blob/main/libraries/WiFiS3/src/WiFi.cpp#L151) and WiFiC3 libraries by Arduino have it right.

This PR adds the default values for gateway, subnet mask and DNS. The solution in LwipIntfDev is the same as in ESP8266 core. And the WiFi library now redirects all versions of `config` to LwipIntfDev.

Tested with Pico W and W5500

### DIFF
--- a/libraries/lwIP_Ethernet/src/LwipIntfDev.h
+++ b/libraries/lwIP_Ethernet/src/LwipIntfDev.h
@@ -61,8 +61,13 @@ public:
         memset(&_netif, 0, sizeof(_netif));
     }
 
+    //The argument order for ESP is not the same as for Arduino. However, there is compatibility code under the hood
+    //to detect Arduino arg order, and handle it correctly.
     bool config(const IPAddress& local_ip, const IPAddress& arg1, const IPAddress& arg2,
                 const IPAddress& arg3 = IPADDR_NONE, const IPAddress& dns2 = IPADDR_NONE);
+
+    // two and one parameter version. 2nd parameter is DNS like in Arduino. IPv4 only
+    boolean config(IPAddress local_ip, IPAddress dns = IPADDR_NONE);
 
     // default mac-address is inferred from esp8266's STA interface
     bool begin(const uint8_t* macAddress = nullptr, const uint16_t mtu = DEFAULT_MTU);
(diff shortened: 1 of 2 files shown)


libraries/lwIP_Ethernet/src/LwipIntfDev.h
```diff
@@ -264,6 +269,24 @@ bool LwipIntfDev<RawDev>::config(const IPAddress& localIP, const IPAddress& gate
...
+boolean LwipIntfDev<RawDev>::config(IPAddress local_ip, IPAddress dns) {
+
+    if (!local_ip.isSet()) {
+        return config(INADDR_ANY, INADDR_ANY, INADDR_ANY);
+    }
+    if (!local_ip.isV4()) {
+        return false;
+    }
+    IPAddress gw(local_ip);
+    gw[3] = 1;
```
earlephilhower, line 283: Notionally we support IPv6 as well as IPv4.  I'm not sure this assignment is valid in the IPv6 case.  Wouldn't we need to set the last octet of the 128 instead in that case?  (Asking as someone whose internet provider doesn't have IPv6 support so has never actually used it so apologies it it's a silly point!)

JAndrassy:
there are no good defaults for IPv6 so the function has a comment with "IPv4 only" and
```
   if (!local_ip.isV4()) {
        return false;
```

libraries/lwIP_Ethernet/src/LwipIntfDev.h
JAndrassy: there are no good defaults for IPv6 so the function has a comment with "IPv4 only" and
```
   if (!local_ip.isV4()) {
        return false;
```

PR #1877, opened 2023-12-04
Updated WiFiMulti.cpp to work with FreeRTOS

richteel:
Modified WiFiMulti.cpp to work with FreeRTOS by removing delay and using a while loop.

### DIFF
--- a/libraries/WiFi/src/WiFiMulti.cpp
+++ b/libraries/WiFi/src/WiFiMulti.cpp
@@ -124,13 +124,13 @@ uint8_t WiFiMulti::run(uint32_t to) {
             // failure when using RTOS
             unsigned long tWifiDelayForRtos_start = millis();
             while (millis()-tWifiDelayForRtos_start <= 5) {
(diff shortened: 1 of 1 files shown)


richteel:
Still failing astyle checks.

richteel:
Earl,

 

I had started a discussion at https://github.com/earlephilhower/arduino-pico/discussions/1873 and wrote quite a bit about the issue on my blog at https://teelsys.com/raspberry-pi-pico-w-rtos-and-wi-fi/. In both the discussion and the blog, I included a zip file with sample code with the failure.

 

If needed, I could pare down the sample code even more as there really is no need to read a config from an SD Card to show the issue. Let me know if you want a bit less code to show the issue and I will pull something together.

 

BTW: I used debug in Arduino and found that the code was not actually hanging but FreeRTOS appears to have been in an infinite loop, which in effect hung code execution.

 

Thanks,

Rich

 

Mobile: 703-472-1049

PR #1878, opened 2023-12-04
Update WiFiMulti.cpp to support FreeRTOS

richteel:
When used with FreeRTOS, the delay statement caused the code to stop responding as FreeRTOS was stuck. Replaced delay with another loop to do nothing for 5ms.

Attempting to fix astyle failed test.

### DIFF
--- a/libraries/WiFi/src/WiFiMulti.cpp
+++ b/libraries/WiFi/src/WiFiMulti.cpp
@@ -124,13 +124,13 @@ uint8_t WiFiMulti::run(uint32_t to) {
             // failure when using RTOS
             unsigned long tWifiDelayForRtos_start = millis();
             while (millis()-tWifiDelayForRtos_start <= 5) {
-		    if(millis()<tWifiDelayForRtos_start) {
-			    // millis wrapped around to zero. Rare to hit this
-			    // issue, but it will do this every 49 days.
-			    tWifiDelayForRtos_start = millis();
-		    }
-	    }
-	}
+                if(millis()<tWifiDelayForRtos_start) {
+                    // millis wrapped around to zero. Rare to hit this
+                    // issue, but it will do this every 49 days.
+                    tWifiDelayForRtos_start = millis();
+                }
+            }
+        }
         if (WiFi.status() == WL_CONNECTED) {
             return WL_CONNECTED;
         }
--- a/libraries/WiFi/src/WiFiMulti.cpp
+++ b/libraries/WiFi/src/WiFiMulti.cpp
@@ -121,16 +121,16 @@ uint8_t WiFiMulti::run(uint32_t to) {
         }
         while (!WiFi.connected() && (millis() - start < to)) {
             // Replaced delay(5); with the following to prevent
-			// failure when using RTOS
-			unsigned long tWifiDelayForRtos_start = millis();
+            // failure when using RTOS
+            unsigned long tWifiDelayForRtos_start = millis();
             while (millis()-tWifiDelayForRtos_start <= 5) {
-               if(millis()<tWifiDelayForRtos_start) {
-                // millis wrapped around to zero. Rare to hit this
-                // issue, but it will do this every 49 days.
-                tWifiDelayForRtos_start = millis();
-               }
-            }
-        }
+		    if(millis()<tWifiDelayForRtos_start) {
+			    // millis wrapped around to zero. Rare to hit this
+			    // issue, but it will do this every 49 days.
+			    tWifiDelayForRtos_start = millis();
+		    }
+	    }
+	}
         if (WiFi.status() == WL_CONNECTED) {
             return WL_CONNECTED;
         }
--- a/libraries/WiFi/src/WiFiMulti.cpp
+++ b/libraries/WiFi/src/WiFiMulti.cpp
@@ -123,8 +123,8 @@ uint8_t WiFiMulti::run(uint32_t to) {
             // Replaced delay(5); with the following to prevent
             // failure when using RTOS
             unsigned long tWifiDelayForRtos_start = millis();
-            while (millis()-tWifiDelayForRtos_start <= 5) {
-                if(millis()<tWifiDelayForRtos_start) {
+            while (millis() - tWifiDelayForRtos_start <= 5) {
+                if (millis() < tWifiDelayForRtos_start) {
                     // millis wrapped around to zero. Rare to hit this
                     // issue, but it will do this every 49 days.
                     tWifiDelayForRtos_start = millis();
--- a/libraries/WiFi/src/WiFiMulti.cpp
+++ b/libraries/WiFi/src/WiFiMulti.cpp
@@ -120,7 +120,16 @@ uint8_t WiFiMulti::run(uint32_t to) {
             WiFi.beginBSSID(j->ssid, j->bssid);
         }
         while (!WiFi.connected() && (millis() - start < to)) {
-            delay(5);
+            // Replaced delay(5); with the following to prevent
+			// failure when using RTOS
+			unsigned long tWifiDelayForRtos_start = millis();
+            while (millis()-tWifiDelayForRtos_start <= 5) {
+               if(millis()<tWifiDelayForRtos_start) {
+                // millis wrapped around to zero. Rare to hit this
+                // issue, but it will do this every 49 days.
+                tWifiDelayForRtos_start = millis();
+               }
+            }
         }
         if (WiFi.status() == WL_CONNECTED) {
             return WL_CONNECTED;
--- a/libraries/WiFi/src/WiFiMulti.cpp
+++ b/libraries/WiFi/src/WiFiMulti.cpp
@@ -58,6 +58,15 @@ bool WiFiMulti::addAP(const char *ssid, const char *pass) {
     return true;
 }
 
+void WiFiMulti::clearAPList() {
+    while (!_list.empty()) {
+        struct _AP ap = _list.front();
+        _list.pop_front();
+        free(ap.ssid);
+        free(ap.pass);
+    }
+}
+
 uint8_t WiFiMulti::run(uint32_t to) {
     struct _scanAP {
         char *ssid;
--- a/libraries/WiFi/src/WiFiMulti.h
+++ b/libraries/WiFi/src/WiFiMulti.h
@@ -31,6 +31,8 @@ public:
     ~WiFiMulti();
 
     bool addAP(const char *ssid, const char *pass = nullptr);
+	
+	void clearAPList();
 
     uint8_t run(uint32_t to = 10000);
 
--- a/libraries/WiFi/src/WiFiMulti.h
+++ b/libraries/WiFi/src/WiFiMulti.h
@@ -31,7 +31,7 @@ public:
     ~WiFiMulti();
 
     bool addAP(const char *ssid, const char *pass = nullptr);
-	
+
     void clearAPList();
 
     uint8_t run(uint32_t to = 10000);
--- a/libraries/WiFi/src/WiFiMulti.h
+++ b/libraries/WiFi/src/WiFiMulti.h
@@ -32,7 +32,7 @@ public:
 
     bool addAP(const char *ssid, const char *pass = nullptr);
 	
-	void clearAPList();
+    void clearAPList();
 
     uint8_t run(uint32_t to = 10000);
 


richteel:
@earlephilhower, I've made a few attempts to get this to pass the astyle check but still fails there. If you could let me know what may be wrong here, I will try to correct formatting in the future. In the meantime, I would appreciate it if you could pull this into the code base.

I also had a question regarding having Arduino IDE pick up these changes. I assume a new version needs to be created, which is done on a schedule or major update of the library.

libraries/WiFi/src/WiFiMulti.cpp
```diff
@@ -120,7 +120,16 @@ uint8_t WiFiMulti::run(uint32_t to) {
             WiFi.beginBSSID(j->ssid, j->bssid);
         }
         while (!WiFi.connected() && (millis() - start < to)) {
-            delay(5);
+            // Replaced delay(5); with the following to prevent
+            // failure when using RTOS
+            unsigned long tWifiDelayForRtos_start = millis();
+            while (millis()-tWifiDelayForRtos_start <= 5) {
+                if(millis()<tWifiDelayForRtos_start) {
```
earlephilhower, line 127: FYI, with unsigned math there is no need for this check at all.  Wraparound deltas just work.

libraries/WiFi/src/WiFiMulti.cpp
```diff
@@ -120,7 +120,16 @@ uint8_t WiFiMulti::run(uint32_t to) {
             WiFi.beginBSSID(j->ssid, j->bssid);
         }
         while (!WiFi.connected() && (millis() - start < to)) {
-            delay(5);
+            // Replaced delay(5); with the following to prevent
+            // failure when using RTOS
+            unsigned long tWifiDelayForRtos_start = millis();
```
earlephilhower, line 125: `delayMicroseconds(5000); // Avoid task switch in FreeRTOS` does the same thing, more simply.

earlephilhower (changes requested):
Please give an MCVE showing the failure you're correcting here.  WiFi is under use by many folks w/FreeRTOS already so I'm not sure exactly what you're seeing or trying to accomplish here.

Please also explain your reasoning for swapping out a task-switching `delay` for a busy wait.  I'm not grokking any difference except this ties up a complete core while waiting for WiFi vs. letting other tasks run...

maxgerhardt:
I think this PR is a direct result of https://github.com/earlephilhower/arduino-pico/discussions/1873. Let me just test their example code that they say hangs..

maxgerhardt:
I've just tested this with the `code.zip` in the discussion but cut out the SD card by commenting out all calls to it and instead hardcoding the config object in `setup()`
```cpp
  // Overwrite config object
  clockSdCard.sdCardConfig = {
    .hostname = "My Pico W",
    .timezone = "GMT",
    .networks = {
      {
        .ssid = "Vodafone-1E90",
        .pass = "XXXXXXXXXXXXXXX"
      }
    }
  };
  configLoadMillis = millis();
```
And ran the sketch with `delay(5);` and `delayMicroseconds(5000);`. In the discussion it seems like this patch helps reduce the connection tries a ton ("it took 12 resets before it connected and kept working"). I did my own testing and found that, indeed, with the microsecond delay, it never had to retry to connect to WiFi. In the millisecond delay, it sometimes had one connection retry.

```
Connecting WiFi...

Failed to connect to Wi-Fi. Trying again...
------^^------ WIFI: Added AP Vodafone-1E90 ------^^------
Connecting WiFi...
```

earlephilhower:
I can't reproduce the failure here using a simplest MCVE with only WiFiMulti.
````
12:19:59.501 -> Connecting to NOBABIES
12:19:59.501 -> [WIFIMULTI] Adding: 'NOBABIES' ElephantsAreGreat' to list
12:19:59.501 -> [WIFIMULTI] Rescanning to build new list of APs
12:20:00.992 -> [WIFIMULTI] scanList: SSID: 'NOBABIES' -- BSSID: '382C4A90FA80' -- RSSI: -30
12:20:00.992 -> [WIFIMULTI] scanList: SSID: 'NOBABIES' -- BSSID: 'AC84C66CEFF3' -- RSSI: -56
12:20:00.992 -> [WIFIMULTI] scanList: SSID: 'NOBABIES' -- BSSID: '50D4F7FEA34D' -- RSSI: -74
12:20:00.992 -> [WIFIMULTI] Connecting to: SSID: 'NOBABIES' -- BSSID: '382C4A90FA80' -- RSSI: -30
12:20:07.458 -> 
````

Would you be able to try adding `#include <FreeRTOS.h>` to the `File->Examples->WiFi->WiFiClient.ino` example, enable `Tools->Debug Port->Serial` and `Tools->Debug Level->All` and capture the output as it runs?

Random guessing here, but what may be happening is noise on the power supply for some PicoWs affecting WiFi performance.  When you `delay()` the core goes to sleep, and the power supply's going to have to adjust down it's supplied current quickly.  When you `delayMicroseconds()` the core runs at 100% meaning the power supply itself doesn't see any transients and should theoretically be less noisy, meaning better WiFi reception.  Especially if you're at the limits of reception already.

The debug logs w/RSSI might help see if it's a weak signal problem.

**Comment by maxgerhardt** [comment]
In a minimal sketch using WiFiMulti **and** FreeRTOS I can also not determine a difference between `delay()` and `delayMicroseconds()`. Each connection is done without retries. My MVCE

```cpp
#include <Arduino.h>
#include <FreeRTOS.h>
#include <task.h>
#include <semphr.h>
#include <WiFi.h>
#include <WiFiMulti.h>

#define WIFI_SSID "Vodafone-1E90"
#define WIFI_PASS "XXXXXXXXXXXXXX"

WiFiMulti multi;
int _beginLoopCount;
bool skip = false;

void setup() {
    while(!Serial);
    multi.addAP(WIFI_SSID, WIFI_PASS);
    delay(1000);    
    Serial.println("Connecting to " + String(WIFI_SSID));
    Serial.flush();
}

void loop() {
    if(skip) return;
    Serial.println("Running on core " + String(get_core_num()) + " task ID " + String(uxTaskGetTaskNumber(xTaskGetCurrentTaskHandle())));
    unsigned long startConnect = millis();
    if (multi.run() != WL_CONNECTED) {
      if (_beginLoopCount < 5) {
        Serial.println("\nFailed to connect to Wi-Fi. Trying again...");
        _beginLoopCount++;
        delay(1000);
      } else {
        Serial.println("\nFailed to connect to Wi-Fi. Quiting");
        _beginLoopCount = 0;
      }
    } else {
      _beginLoopCount = 0;
      Serial.print("\nWiFi connected, SSID: ");
      Serial.print(WiFi.SSID());
      Serial.print(", IP address: ");
      Serial.println(WiFi.localIP());
      Serial.printf("Connected in %d ms\n", millis() - startConnect);
      skip = true;
    }
}
```

earlephilhower:
I also tried the full app example (commenting out and hacking in the config like you did @maxgerhardt and making all debug go to USB). I did *not* use the edited WiFiMulti.cpp/.h in the `lib` directory, only the core version.

````
earle@amd:~/Arduino$ diff -w --recursive sketch_dec2a1 sketch_dec2a
diff -w --recursive sketch_dec2a1/DbgPrint.h sketch_dec2a/DbgPrint.h
6,16d5
< 
< #if DEBUG
< #if USE_PICOPROBE
< #define Dbg_begin(...) Serial1.begin(__VA_ARGS__);
< #define Dbg_end(...) Serial1.end(__VA_ARGS__);
< #define Dbg_print(...) Serial1.print(__VA_ARGS__)
< #define Dbg_println(...) Serial1.println(__VA_ARGS__)
< #define Dbg_printf(...) Serial1.printf(__VA_ARGS__)
< #define Dbg_printlnf(...) Serial1.printlnf(__VA_ARGS__)
< #define Dbg_write(...) Serial1.write(__VA_ARGS__)
< #else
24,33d12
< #endif
< #else
< #define Dbg_begin(...)
< #define Dbg_end(...)
< #define Dbg_print(...)
< #define Dbg_println(...)
< #define Dbg_printf(...)
< #define Dbg_printlnf(...)
< #define Dbg_write(...)
< #endif
diff -w --recursive sketch_dec2a1/sketch_dec2a.ino sketch_dec2a/sketch_dec2a.ino
111c111
< 
---
> delay(5000);
122c122,134
<   clockSdCard.begin();
---
>   // Overwrite config object
>   clockSdCard.sdCardConfig = {
>     .hostname = "My Pico W",
>     .timezone = "GMT",
>     .networks = {
>       {
>         .ssid = "NOBABIES",
>         .pass = "xxxx"
>       }
>     }
>   };
>   configLoadMillis = millis();
> //  clockSdCard.begin();
126,128c138,140
<   if (clockSdCard.isCardPresent()) {
<     configLoadMillis = millis();
<   }
---
> //  if (clockSdCard.isCardPresent()) {
> //    configLoadMillis = millis();
> //  }
````

I've power-cycled the PicoW about ten times so far and it always connects on the 1st try for me.  That said, I am ~3 meters from the AP it's connecting to w/a RSSI of~-32.  So if this is noise related there's plenty of headroom in the signal it's seeing which might not be the case at the @richteel's location...

**Comment by richteel** [comment]
Phil,

 

I will give those options a try and see if I still have an issue. It could be a power supply issue although the hub that it is connected to has its own power supply and nothing else is connected to that hub. It should not be an issue with signal as my AP is about 4 foot and my second testing one is about 2 feet away. If anything, I would think it is a power issue but really doubt that. I will try it on a Pico W that is not in circuit too just to be certain.

 

The reason I suspected it was the delay as the I read somewhere in the FreeRTOS documentation that delay should not be used but vTaskDelay should be used instead if a delay is required. When I was stepping through the code when it was stuck, I noticed that it was bouncing around in the FreeRTOS code. Don’t recall exactly but remember seeing it look at get minimum delay or something similar.

 

Thanks,
Rich

**Comment by richteel** [comment]
Oh, I should have stated that when I was having the issue, it would take about 30 seconds before it seemed to lock up. Once I resolved the issue, it took 8 seconds to connect. Another reason I’m sure it is no the signal or noise. If it is anything it would be power but seriously doubt that.

 

From: ***@***.*** ***@***.***> 
Sent: Monday, December 4, 2023 4:11 PM
To: 'earlephilhower/arduino-pico' ***@***.***>; 'earlephilhower/arduino-pico' ***@***.***>
Cc: 'Mention' ***@***.***>
Subject: RE: [earlephilhower/arduino-pico] Update WiFiMulti.cpp to support FreeRTOS (PR #1878)

 

Phil,

 

I will give those options a try and see if I still have an issue. It could be a power supply issue although the hub that it is connected to has its own power supply and nothing else is connected to that hub. It should not be an issue with signal as my AP is about 4 foot and my second testing one is about 2 feet away. If anything, I would think it is a power issue but really doubt that. I will try it on a Pico W that is not in circuit too just to be certain.

 

The reason I suspected it was the delay as the I read somewhere in the FreeRTOS documentation that delay should not be used but vTaskDelay should be used instead if a delay is required. When I was stepping through the code when it was stuck, I noticed that it was bouncing around in the FreeRTOS code. Don’t recall exactly but remember seeing it look at get minimum delay or something similar.

 

Thanks,
Rich

**Comment by earlephilhower** [comment]
Re: `delay()` vs `vTaskDelay()`...in this core, `delay()` calls it behind the scenes, no need to change Arduino code:
https://github.com/earlephilhower/arduino-pico/blob/280fc437315278291c1d70d946493aa0de2015d5/libraries/FreeRTOS/src/variantHooks.cpp#L155-L157

**Comment by earlephilhower** [comment]
Also, by "power" I meant the onboard power supply.  There's an LDO on the PCB which can operate in different modes to bring the 5V USB down to the 3.3V (or lower) required by the chip.  I was suggesting that the transients could come from that and not the USB cable or hub itself (although on the ESP8266 there are many, many instances where a poor quality USB cable caused WiFi problems there with Vdroop because it was so power hungry while transmitting...)

Checking the seen RSSI would be a useful thing just to see if it's at the margin or if it's loud and clear to the Pico.

**Comment by richteel** [comment]
Here are the results of running the ScanNetworks example code to get the RSSI for each network.

```
Beginning scan at 5003
Found 8 networks

                            SSID   ENC     BSSID         CH RSSI
                        tl6cd020  WPA2 00:25:F0:6C:D0:20 10  -62
                      bananabros  WPA2 08:BF:B8:22:DC:30  2  -81
                      Fios-212BS  WPA2 20:C0:47:C4:3F:04  1  -78
                      Fios-X0BEY  WPA2 20:C0:47:EC:33:2E  1  -89
                        T_Travel  WPA2 28:87:BA:88:10:5A 10  -46
                                  AUTO 56:12:65:53:20:5F  6  -87
                         TeelSys  WPA2 C4:41:1E:A9:B8:57 10  -34
                                  WPA2 CA:41:1E:A9:B8:57 10  -33

--- Sleeping ---
```

Running the minimal example code that @maxgerhardt, posted above, I get the same behavior as I saw in my example code. The Raspberry Pico W seems to lock up as it fails to execute the remaining code. Also, when I attempt to upload a new sketch, I need to unplug it and press the BOOTSEL button while I plug it back in. I tried this on three Pico W devices, with one of them being brand new. Here is the output of the sample code above. It has been sitting for about 10 minutes now without advancing any further.

```
Connecting to TeelSys
Running on core 0 task ID 3330250393
```

I tried the code without any references to FreeRTOS and it connects right away. Here is the results of that code.

```
Connecting to TeelSys
Running on core 0

WiFi connected, SSID: TeelSys, IP address: 10.140.1.58
Connected in 8001 ms
```
It appears having debugging turned on is changing the behavior of the code. It connects successfully with debugging but takes a long time to connect as it fails with the first attempt.
Here is the debug output with the minimal code example:

```
[WIFIMULTI] Adding: 'TeelSys' xxxxxxxxx' to list
Connecting to TeelSys
Running on core 0 task ID 3330250457
[WIFIMULTI] Rescanning to build new list of APs
[WIFIMULTI] scanList: SSID: 'TeelSys' -- BSSID: 'C4411EA9B857' -- RSSI: -47
[WIFIMULTI] Connecting to: SSID: 'TeelSys' -- BSSID: 'C4411EA9B857' -- RSSI: -47

Failed to connect to Wi-Fi. Trying again...
Running on core 0 task ID 3330250457
[WIFIMULTI] Rescanning to build new list of APs
[WIFIMULTI] scanList: SSID: 'TeelSys' -- BSSID: 'C4411EA9B857' -- RSSI: -41
[WIFIMULTI] Connecting to: SSID: 'TeelSys' -- BSSID: 'C4411EA9B857' -- RSSI: -41

WiFi connected, SSID: TeelSys, IP address: 10.140.1.58
Connected in 7702 ms
```

After unplugging the Pico W, it seemed to lock up again but it did try to make a second attempt.

```
[WIFIMULTI] Adding: 'TeelSys' xxxxxxxxx' to list
Connecting to TeelSys
Running on core 0 task ID 3598685913
[WIFIMULTI] Rescanning to build new list of APs
[WIFIMULTI] scanList: SSID: 'TeelSys' -- BSSID: 'C4411EA9B857' -- RSSI: -44
[WIFIMULTI] Connecting to: SSID: 'TeelSys' -- BSSID: 'C4411EA9B857' -- RSSI: -44

Failed to connect to Wi-Fi. Trying again...
Running on core 0 task ID 3598685913
[WIFIMULTI] Rescanning to build new list of APs
[WIFIMULTI] scanList: SSID: 'TeelSys' -- BSSID: 'C4411EA9B857' -- RSSI: -43
[WIFIMULTI] Connecting to: SSID: 'TeelSys' -- BSSID: 'C4411EA9B857' -- RSSI: -43
```

**Comment by earlephilhower** [comment]
Lockups and slow connections are very different symptoms with very different causes.  I suggest debugging them separately.

For lock ups, please make sure you are using the latest and greatest `git master` and not the 3.6.1 release.  #1872 has a fix for an IRQ race condition.  Also, make sure you are not using MDNS because it is incompatible w/FreeRTOS #1880.

When you get a lock up, please grab the stack trace for both cores.  If that's all scrambled, or PicoProbe can't connect to the SWD debugger, then it's a physical/voltage problem and not SW...

**Comment by richteel** [comment]
Okay, the plot thickens. Delay is not the issue but having only delay in the while statement causes the issue.

I had removed the if block from the while loop and I saw the same behavior as the delay.

```
        while (!WiFi.connected() && (millis() - start < to)) {
            // Replaced delay(5); with the following to prevent failure when using RTOS
            unsigned long tWifiDelayForRtos_start = millis();
            while (millis() - tWifiDelayForRtos_start <= 5) {
                ;
            }
        }
``` 

I then added the if block back in and commented out the while statement and closing bracket and added the delay statement back in and it would connect on the first or second connect.

```
        while (!WiFi.connected() && (millis() - start < to)) {
            // Replaced delay(5); with the following to prevent failure when using RTOS
            unsigned long tWifiDelayForRtos_start = millis();
            // while (millis() - tWifiDelayForRtos_start <= 5) {
                if (millis() < tWifiDelayForRtos_start) {
                    // millis wrapped around to zero. Rare to hit this
                    // issue, but it will do this every 49 days.
                    tWifiDelayForRtos_start = millis();
                }
            //}
			delay(5);
        }
```

Any idea why there needs to be something in addition to delay in the loop?

Calling it a night so will pick this back up tomorrow afternoon.

maxgerhardt:
Take not I tested with PlatformIO and the commit d2461a14ad5aa920e44508d236c2f459e3befbf8. That is exactly the commit that has the IRQ fix. If *any* delay somewhere causes a task switch which causes the incorrect enabling of IRQs, that could be the problem. Can you manually apply the above commit and retry?

PR #1882, opened 2023-12-04
Fix SPI 16-bit transfers

### LINKED ISSUE #1874: SPI beginTransaction does not check if bits are set to 8
If SPI beginTransaction is called for the first time, then the format is correctly set to 8 bits since this is required for the SPI library functions to work correctly.

With the latest 3.6.1 release, if the user calls endTransaction and reconfigures the SPI port for 16 bits, then a subsequent call to beginTransaction with the same settings will not result in the port being configured for 8 bits and the library SPI transfer calls then behave incorrectly. Earlier board package versions handled this situation because the SPI bit count was set in the transfer functions.

There is no "get" call in the SDK to check the SPI bits setting and the ["settings" comparison](https://github.com/earlephilhower/arduino-pico/blob/master/libraries/SPI/src/SPI.cpp#L184) will pass if the settings are the same since the bit count is not part of these settings.

One solution is to always set the bit count to 8 in beginTransaction using:
```
hw_write_masked(&spi_get_hw(_spi)->cr0, (8 - 1) << SPI_SSPCR0_DSS_LSB, SPI_SSPCR0_DSS_BITS);
```
This line can be added [here](https://github.com/earlephilhower/arduino-pico/blob/master/libraries/SPI/src/SPI.cpp#L185). This is a register write solely for setting the SPI bits to 8, so it is faster than using:
```
spi_set_format(_spi, 8, cpol(), cpha(), SPI_MSB_FIRST);
```

Another option is to leave "as is" and rely on the user to always reset the SPI bit count to 8 before calling beginTransaction.

### LINKED ISSUE #1879: SPI transfer16 operation changed in v3.6.1
Some SPI devices require 16 bit command data transfers:
LSM330 SPI read

These devices worked fine prior to v3.6.1 since a true 16 bit transfer occured and the chip select was raised at the end of the 16 bit transfer (clock (blue) and CS(red) signals shown):
transfer16 v3_6_0

If CS is raised in the middle of the 16bit transfer then the device aborts the command and fails to return data.

In v3.6.1 the 16 bit transfer is split into two 8 bit transfers, this causes problems with some devices:
transfer16

It is not clear why this change as been made since there is a performance penalty and it now precudes the use of transfer16() in some SPI peripheral device cases.  Note: it is safe to change the format from 8 to 16 bits, providing checks are made that the SPI FIFO buffer is clear and transfer is complete. In the case where DMA is used then it must be checked that the DMA is complete AND the SPI port has completed transfer of the last bytes/words. If these checks are not made then glitches/faults can occur in the transfer/data signals.

The short term work-around is to change user code to manage the chip select under software control. but this is inconvenient...

earlephilhower:
Fixes #1879
Fixes #1874

### DIFF
--- a/libraries/SPI/keywords.txt
+++ b/libraries/SPI/keywords.txt
@@ -18,6 +18,7 @@ beginTransaction	KEYWORD2
 endTransaction	KEYWORD2
 SPISettings	KEYWORD2
 transfer	KEYWORD2
+transfer16	KEYWORD2
 setBitOrder	KEYWORD2
 setDataMode	KEYWORD2
 setClockDivider	KEYWORD2
--- a/libraries/SPI/src/SPI.cpp
+++ b/libraries/SPI/src/SPI.cpp
@@ -111,6 +111,7 @@ byte SPIClassRP2040::transfer(uint8_t data) {
     }
     data = (_spis.getBitOrder() == MSBFIRST) ? data : reverseByte(data);
     DEBUGSPI("SPI::transfer(%02x), cpol=%d, cpha=%d\n", data, cpol(), cpha());
+    hw_write_masked(&spi_get_hw(_spi)->cr0, (8 - 1) << SPI_SSPCR0_DSS_LSB, SPI_SSPCR0_DSS_BITS); // Fast set to 8-bits
     spi_write_read_blocking(_spi, &data, &ret, 1);
     ret = (_spis.getBitOrder() == MSBFIRST) ? ret : reverseByte(ret);
     DEBUGSPI("SPI: read back %02x\n", ret);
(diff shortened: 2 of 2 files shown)


earlephilhower:
@Bodmer can you give this a try and report back, please?

````
#include <SPI.h>

void setup() {
  SPI.setRX(0);
  SPI.setCS(1);
  SPI.setSCK(2);
  SPI.setTX(3);
  SPI.begin(true);
}
int x = 0;
void loop() {
  delay(3000);
  Serial.println(x++);
  SPI.beginTransaction(SPISettings(1000000, MSBFIRST, SPI_MODE0));
  SPI.transfer16(0xaaaa);
  delayMicroseconds(1);
  SPI.transfer(0x55);
  delayMicroseconds(1);
  uint8_t out[2] = {0xaa, 0x55};
  uint8_t in[2];
  SPI.transfer16(0x0000);
  SPI.transfer(out, in, 2);
  delayMicroseconds(1);
  SPI.endTransaction();
}
````
Gives the expected CS values:
image

PR #1883, opened 2023-12-05
Disable interrupts first when idling core

GUVWAF:
I encountered an issue similar to the one in #1868, but this one is way more unlikely to happen. I believe what happens is similar to what you suspected in https://github.com/earlephilhower/arduino-pico/issues/1868#issuecomment-1836761063. When `core0` wants to idle `core1`, it waits forever in `while (!__otherCoreIdled)` since at that time `core1` is in an ISR, acquiring a lock (see the callstacks of both cores below). Therefore, I’ve moved `portDISABLE_INTERRUPTS` up in `IdleThisCore` and with this it’s been running for 11 hours straight without occurring. 
I’m not sure if this guarantees that it cannot happen anymore, but it should at least lower the chance.

**Callstack `core0`:**
```
__freertos_idle_other_core@0x20000120 (~/.platformio/packages/framework-arduinopico/libraries/FreeRTOS/src/variantHooks.cpp:182)
_MFIFO::idleOtherCore@0x10039dc0 (~/.platformio/packages/framework-arduinopico/cores/rp2040/RP2040Support.h:97)
RP2040::idleOtherCore@0x1004cb88 (~/.platformio/packages/framework-arduinopico/cores/rp2040/RP2040Support.h:265)
__wrap_get_rand_64@0x1004cb88 (~/.platformio/packages/framework-arduinopico/cores/rp2040/_freertos.cpp:80)
__wrap_get_rand_32@0x1004cbb6 (~/.platformio/packages/framework-arduinopico/cores/rp2040/_freertos.cpp:92)
igmp_start_timer@0x1005e4ec (Unknown Source:695)
igmp_delaying_member@0x1005e4ec (Unknown Source:718)
igmp_delaying_member@0x1005e4ec (Unknown Source:713)
igmp_input@0x1005e6f2 (/igmp_input.dbgasm:103)
ip4_input@0x1005f45e (/ip4_input.dbgasm:305)
ethernet_input@0x1005f8ba (/ethernet_input.dbgasm:87)
__wrap_cyw43_cb_process_ethernet@0x1009b4de (~/.platformio/packages/framework-arduinopico/cores/rp2040/cyw43_wrappers.cpp:56)
cyw43_ll_process_packets@0x100603b0 (/cyw43_ll_process_packets.dbgasm:29)
cyw43_poll_func@0x100561fc (Unknown Source:238)
cyw43_poll_func@0x100561fc (Unknown Source:219)
cyw43_do_poll@0x100568ae (Unknown Source:90)
[...]

PR #1884, opened 2023-12-05
Avoid freezeing the core from LWIP under FreeRTOS

earlephilhower:
Avoid issues with interrupts and priority inversions and other deadlocks and use a SW based random generator for LWIP when under FreeRTOS.

This means removing any overrides for sleep_until and the two get_rand_xx calls from the SDK, making things much saner.

Related to #1883, #1872, and other random FreeRTOS lockups.


earlephilhower:
@GUVWAF would you be able to try this PR with your app?  It removes the whole need for the IRQ-level freeze/unfreeze and actually should be faster overall since we don't need to freeze the opposite core every time LWIP asks for a random number (i.e. for a port assignment or backoff timer).

GUVWAF:
Ah, this sounds like a better overall solution indeed. Thanks for looking so deeply into it. 
I'll try it by letting it run overnight with debugger.

GUVWAF:
All seems fine with this. I’ve not observed a hang during two 10-hour tests (had to replace my set-up once). 

Thanks a lot!

earlephilhower:
Great news.  Thanks for the update!

Issue #1893, opened 2023-12-11
SD.h wrong calculation of totalBlocks


erpebe:
In the following code should that not be a multiplication instead of a division?
There are more blocks than clusters.

```
size_t totalBlocks() {
  return (totalClusters() / blocksPerCluster());
}
```

earlephilhower:
Seems like a good catch.  Want to do the PR since you found and fixed it?

earlephilhower:
Fixed by #1899

PR #1899, opened 2023-12-13
correct calculation of totalBlocks() in SD.h and SDFS.h

erpebe:
#1893 SD.h wrong calculation of totalBlocks


earlephilhower (approved):
Thanks very much for fixing in both places!

PR #1911, opened 2023-12-23
Add variants_dir support to pio build, allowing local variants in project repos

josephduchesne:
This is adapted from the same feature implemented in https://github.com/espressif/arduino-esp32/blob/master/tools/platformio-build.py on line 178.

It seems to work correctly locally.


earlephilhower (approved):
LGTM, thanks.  @maxgerhardt this is really your file, any objections?

maxgerhardt:
Change was 100% good, somehow the Github reply-email didn't show up here (?), but it all worked out ):

Issue #1182, opened 2023-02-13
Not working with different i2c pins


lolren:
Tried to set the i2c pis to 26 and 27, to use it with a u8g2 library. It doesn’t work. Display get’s found on the right address. It does work with a different core (mbed), but not with this one. Does anyone else have this issue? Regards

metehoca:
Hi, adding "#define WIRE_INTERFACES_COUNT 2" in "packages\rp2040\hardware\rp2040\3.6.0\variants\rpipico\pins_arduino.h" works with this library.

@earlephilhower would you add this define in next release if won't create problem?

PR #1921, opened 2024-01-01
Adds definition for WIRE_INTERFACES_COUNT #1182

freespace:
so those libraries which rely on it can detect and use Wire1. e.g. u8g2 will not use Wire1 unless this is (a) defined and (b) >1. This constant is defined on other cores, e.g. for SAMD based boards.

This should resolve situations like that reported in #1182.

I am not entirely sure if `WIRE_INTERFACES_COUNT` should be defined in `common.h`, as in other cores it is defined in variant-specific headers, e.g.

https://github.com/search?q=repo%3Aarduino%2FArduinoCore-samd%20WIRE_INTERFACES_COUNT&type=code

However given that `generic/pins_arduino.h` defines pins for both I2C0 and I2C1, this doesn't seem unreasonable. i.e. the generic variant already assumes the presence of 2 I2C peripherals.

### DIFF
--- a/variants/generic/common.h
+++ b/variants/generic/common.h
@@ -7,6 +7,7 @@
 #define NUM_ANALOG_INPUTS   (4u)
 #define NUM_ANALOG_OUTPUTS  (0u)
 #define ADC_RESOLUTION      (12u)
+#define WIRE_INTERFACES_COUNT (2u)
 
 #ifdef PIN_LED
 #define LED_BUILTIN PIN_LED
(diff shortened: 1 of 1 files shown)


earlephilhower (changes requested):
Thanks for this, but we already have a per-device flag that should be used instead of a fixed 2, `WIRE_HOWMANY`.    Can you change the `(2)` to `(WIRE_HOWMANY)` please?

freespace:
@earlephilhower thanks for letting me know! I will make the change, test and update :)

earlephilhower (approved):
Great, thanks!

PR #1933, opened 2024-01-10
Adding Dude's Cab board

arnoz26:
Adding Dude's Cab board (rp2040 based board for virtual pinball use)


earlephilhower (approved):
Thanks, looks much better!

Issue #392, opened 2021-12-26
spi.beginTransaction() takes a long time to complete


Bodmer:
A line like this using this board package:
```
spi.beginTransaction(SPISettings(SPI_FREQUENCY, MSBFIRST, TFT_SPI_MODE));
```
takes ~50us to complete. Using the Arduino Mbed OS RP2040 package completes in <1us.

It appears that the function is slowed down for some reason, maybe by debug code?

Bodmer:
Here is a test sketch that demonstrates the problem. This slows down SPI based sketches significantly.
```
// Arduino RP2040 package   ~0.49us
// Community pacakge      ~ 69.10us
#include <SPI.h>
void setup() {
  Serial.begin(115200);
  SPI.begin();
}

void loop() {
  uint32_t dt = micros();
  for (int i = 0; i < 1000; i++) {
    SPI.beginTransaction(SPISettings(70000000, MSBFIRST, 0));
    SPI.endTransaction();
  }
  Serial.println(micros() - dt);
}
```

earlephilhower:
@Bodmer please try the PR #394 and report back, since you seem to have a setup ready to test.

It should make performance basically identical to the MBED one, but I need someone to verify that the change I did still results in working SPI comms...

Bodmer:
Yes, performance is now the same.

Thank you.

earlephilhower:
> Yes, performance is now the same.

And, just to be 100% sure, it still works, right? :)

Bodmer:
Yes, it works for me. I have two SPI devices on the same bus that operate at different SPI clock frequencies and the beginTransaction() and endTransaction() functions operate as expected, changing the SPI frequency when each device is accessed.

Bodmer:
Performance is now comparable to the Mbed package. Drawing 10,000 pixels to a TFT, first using beginTransaction() and endTransaction(), second without.
```
1.9.9 Philhower package!
drawPixel = 1486ms
Locked drawPixel = 75ms

Arduino MBED!
drawPixel = 113ms
Locked drawPixel = 77ms

FastSPI Philhower package!
drawPixel = 110ms
Locked drawPixel = 75ms
```

earlephilhower:
Thanks for the confirmation!  This will be in the next release, coming soon (serial stuff I want to get in, too).

PR #1934, opened 2024-01-11
Skip SPI re-init if clock frequency doesn't change

Nerradia:
Hello,

In link with #392, I have two SPI devices that use the same clock frequency but not the same clock polarity. From what I've seen, re-initing the whole SPI is slow and unecessary when clock is the same.

I haven't done much testing but it works in my case.


libraries/SPI/src/SPI.cpp
```diff
@@ -183,16 +183,20 @@ void SPIClassRP2040::beginTransaction(SPISettings settings) {
...
+    }
+    else {
+        /* Only de-init if the clock changes frequency */
+        if (settings.getClockFreq() != _spis.getClockFreq()) {
+          if (_initted) {
+              DEBUGSPI("SPI: deinitting currently active SPI\n");
+              spi_deinit(_spi);
+          }
+          DEBUGSPI("SPI: initting SPI\n");
+          spi_init(_spi, _spis.getClockFreq());
```
earlephilhower, line 195: This is backwards, actually.  `settings` is the new value (you've moved the copy `_spis = settings` after this block) so you want `spi_init(_spi, settings.getClockFreq());` I believe.

earlephilhower (changes requested):

Nerradia:
Sorry I've been wuite busy last week. I have fixed the code as you requested. I guess I can't update this PR if it is closed ?

earlephilhower:
Try it now, please.

Nerradia:
Alright, I think the formatting is good now.
Sorry for struggling a bit, that's my first PR !

earlephilhower (approved):
Thanks for getting this cleaned up!

PR #1935, opened 2024-01-11
WiFi - prepare for alternative drivers

JAndrassy:
Before integrating an alternative WiFi driver (lwIP_ESPHost) I moved CYW43 WiFiClass function implementations to lwIP_CYW43.
For boards using the WiFi library with wired network I added lwIP_nodriver empty implementation.

Tested with [WiFiTest](https://github.com/JAndrassy/NetApiHelpers/blob/master/examples/WiFiTest/WiFiTest.ino), [WiFiScanNetworks](https://github.com/JAndrassy/NetApiHelpers/blob/master/examples/WiFiScanNetworks/WiFiScanNetworks.ino) and [ModernEthernetTest](https://github.com/JAndrassy/NetApiHelpers/blob/master/examples/ModernEthernetTest/ModernEthernetTest.ino)


earlephilhower:
Thanks, looks like a reasonable refactoring and I'll try and get on it over the long weekend.

However, do you think it would make sense to make a base `WiFiRadio` or `WiFiBase` class which virtually defines the common WiFi calls like `scanNetworks` or `beginAPMode` and then make the CYW43 class inherit from that in addition to LwipIntfDev?

I haven't thought through the implications of either way yet, but you might have some insight after spending so much time refactoring here...

JAndrassy:
and I was thinking about inheriting WiFiClass from specific LwipIntfDev implementation (with ifdefed lines after `class WiFiClass :`). 
as this in internal it can be refactored at any time. doesn't have to be ideal from the start. 
so for this PR I just did the simplest refactoring to separate the CYW43 code from WiFiClass. 
I think it goes in the right direction for a future nice OOP architecture between WiFiClass and LwipIntfDev

earlephilhower:
Looks okay to me.  Still a draft?

JAndrassy:
>  Still a draft?

as I write the lwIP_ESPHost library, I changes a few things, tested them and forgot to remove the draft state. 
it can be merged.

earlephilhower (approved):
Thanks again!

PR #1950, opened 2024-01-21
lwIP_ESPHost - new library

JAndrassy:
questions:
* how to do a semaphore? mutex is not good to wait
* why beginAP returns WL_CONNECTED and not WL_AP_LISTENING? I had a problem with AP example from WiFiNINA


earlephilhower:
>     * how to do a semaphore? mutex is not good to wait

The SDK semaphore routines would be best: https://www.raspberrypi.com/documentation/pico-sdk/high_level.html#sem

>     * why beginAP returns WL_CONNECTED and not WL_AP_LISTENING? I had a problem with AP example from WiFiNINA

No reason other than I think it was what we did on the 8266.  If it should be changed, please feel free to do so.

JAndrassy:
`WL_AP_LISTENING` can only exists if soft AP and STA can't be active together. so esp8266 WiFi library doesn't use the WL_AP_ states and uses `status()` only for STA. Here the Pico W WiFi can't run in both modes as I understand. The ESPHost has STA+AP mode, but the lwIP_ESPHost doesn't support it for now, just to match with the WiFi library. It would be too early to rewrite the WiFi library. lwIP_ESPHost can still be a failure technically or for no interest from users.

JAndrassy:
The CoreMutex class has `if (__isFreeRTOS) ...` . Will the SDK sem functions work with RTOS?

earlephilhower:
The SDK semaphore won't work properly in FreeRTOS, no.  It only tracks the core as the owner, not a task id.  What is the use case for the semaphore?  The same kind of wrapper can be imported to use FreeRTOS semaphores of needs be.

JAndrassy:
>  What is the use case for the semaphore?

The lwIP_ESPHost library uses polling with LwipIntfDev.handlePackets which runs with a timer. But It also calls the ESPHost library from WiFi control functions begin, end, scanNetworks. And access to ESPHost library must be synchronized.

earlephilhower:
I would suggest looking at using the `‎ethernet_arch_lwip_begin/_end` pre-existing setup.  That protects against LWIP re-entrancy and other badness for the periodic timer/IRQ-based interfaces including the CYW43 Wifi driver.

JAndrassy:
basic tests with WiFiTest and WiFiScanTest from my NetApiHelpers library and AP_SimpleWebServer modified from WiFiNINA library.
and I tested with the bundled ArduinoOTA and with FSBrowser. the performance look acceptable.

earlephilhower:
@JAndrassy really nice work, but I wonder how you're setting the `ESPSPI` define you're bracketing things with?  It's not defined in any of the variants here and you generally can't set a global `#define` in any of the Arduino IDEs?

I've just received my package of ESP32-(alphabet soup) from AliExpress this week and hope to wire things up and give it a go.

Where is `CEspControl` coming from?  Is this one of your other libraries?  I would like to `submodule` it and then add an example to ensure that this code would be built on every PR.  (That's why there's one example for each of the other lwip_xxx drivers...if I don't build everything on every check in then I can almost guarantee it not compiling in a couple releases...)

JAndrassy:
CEspControl is from the ESPHost library. Link is in the README. The README of the library has some instructions for the firmware. 
the ESPHOSTSPI should be in the future defined in boards.txt (or in boards.local.txt). I work in Eclipse so I just set it for the project.

PR #1959, opened 2024-01-25
Add native Apple ARM silicon M1/M2/M3 support

earlephilhower:


earlephilhower:
@rgov if you have a few minutes, would you please be able to check this PR out?

This *should* deploy Apple ARM native versions of the toolchain (separate version, not a fat binary).  You'll need to run the `tools/get.py` after checking out this PR to download the new `darwin_arm` versions of the toolchain from my 2.2.0 pre-release.  (Also for now the `picotool` and `openocd` tools are still always x86_64....if the main toolchain bins are good here I can see how to make arm64 versions using GH actions).

earlephilhower:
Arg, I think the `get.py` script isn't going to work properly because it doesn't know how to identify your Mac is ARM based.

Can someone run this Python script and give me the output on an M1/2/3 Mac?  Or if you have some other Pythonic way of ID'ing an ARM version?
````
import platform
print(platform.system())
````

**Comment by blurfl** [comment]
M1 Studio, python 3.11.6, the output is one line:

`Darwin`

**Comment by earlephilhower** [comment]
Yuck, that is crazy.  Is Python M1 native, @blurfl ?  Seems like it's reporting some crazy x86 Intel report string?!

Could you try one more thing suggested by SO (so not great confidence):
````
import platform
import os
print(platform.processor())
print (os.uname())
````

**Comment by Jason2866** [comment]
@earlephilhower  This:
```
import platform
print(platform.system())
print(platform.platform())
```
results in with a M1 in
```
Darwin
macOS-13.4-arm64-arm-64bit
```
this:
```
import platform
import os
print(platform.processor())
print (os.uname())
```
results in
```
arm
posix.uname_result(sysname='Darwin', nodename='MacBook-Air-von-Johann.local', release='22.5.0', version='Darwin Kernel Version 22.5.0: Mon Apr 24 20:53:44 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T8103', machine='arm64')
```

**Comment by Jason2866** [comment]
An example valid package.json for Platformio for native Apple Silicon toolchain
```
{
  "name": "toolchain-riscv32-esp",
  "version": "12.2.0+20230208",
  "description": "GCC Toolchain for Espressif 32-bit RISC-V based MCUs",
  "keywords": [
    "build tools",
    "compiler",
    "assembler",
    "linker",
    "preprocessor",
    "espressif",
    "esp32",
    "esp32c2",
    "esp32c3",
    "esp32c6",
    "esp32h2",
    "risc-v"
  ],
  "license": "GPL-2.0-or-later",
  "repository": {
    "type": "git",
    "url": "https://github.com/espressif/crosstool-NG"
  },
  "system": [
    "darwin_arm64"
  ]
}
```

earlephilhower:
Thanks @Jason2866 , I do have the Platform.IO JSON format good (I think, matches the system you show):
````
{
   "description": "arm-none-eabi-gcc",
   "name": "toolchain-rp2040-earlephilhower",
   "system": [ "darwin_arm64" ],
   "url": "https://github.com/earlephilhower/pico-quick-toolchain",
   "version": "5.120300.240125"
}
````

I'm trying to modify the regular IDE download `get.py` script to be M1 aware so it will d/l and install the ARM binaries when run on a Mac...

Jason2866:
@earlephilhower  Maybe a look in get.py from espressif Arduino32 gives an idea
https://github.com/espressif/arduino-esp32/blob/master/tools/get.py

earlephilhower:
>  Maybe a look in get.py from espressif Arduino32 gives an idea

Perfect, exactly what I needed.  Thanks!  I just pushed so anyone on the M1 can now try `get.py` to get the proper tools (AFAIK!).

> EDIT: Do you see a chance to build the native Apple ARM toolchain for xtensa esp8266 too?

If this gets worked out then it's not too big a step to build them, but the major issue is that GCC does **NOT** support `darwin-arm64` natively.  To make this build I had to borrow a patchset from the `brew` gang, and those patches are limited to very specific versions of GCC because they twiddle lots of intimate bits.  If we can move the ESP8266 to GCC 12.3 then I have no doubt it can be done.  Other versions, we'll need to see what the homebrew guys have done...

Jason2866:
I provided a PR to espressif fork of GCC (v8.4.0) to build the toolchains native for Apple ARM.
https://github.com/espressif/gcc/tree/esp_based_on_8_4_0
A later GCC builds are based on this patch set

Maybe something you can use?

blurfl:
M1 yes, surprised me too.  Looks like it's due to my outdated homebrew installation.
FWIW, os.uname() returns
 
`posix.uname_result(sysname='Darwin', nodename='xxxxx.local', release='23.3.0', version='Darwin Kernel Version 23.3.0: Wed Dec 20 21:30:44 PST 2023; root:xnu-10002.81.5~7/RELEASE_ARM64_T6000', machine='x86_64')`

Jason2866:
@blurfl Rosetta kicks in. Should not happen. "Something" is screwed up in the setup.

rgov:
For the record, the preferred way to deliver binaries on macOS is "universal" aka fat binaries that contain both architectures. The OS chooses the correct slice at load time. The binaries can be built separately and then merged together. (The main way it gets tricky is when you need to provide a non-executable resource file that varies by architecture.) Is this an option?

earlephilhower:
> @rgov said:  For the record, the preferred way to deliver binaries on macOS is "universal" aka fat binaries that contain executables for both architectures inside an archive. The OS chooses the correct slice at load time.

Right now I'd be happy with the individual binaries running properly.  Anything else is icing on the cake. 🍰 😆 

FWIW, the IDE, `get.py`, and Platform.IO all download just the specific ARM/x86_64 versions of the code so other than making everyone download a file about 2x the size with 1/2 unused I'm not sure there's a big reason to merge them.  I'm also not 100% sure that fat-binary-tool (name?) exists in a cross-compilation mode (i.e. it needs to run on my Linux Docker image, not on any real Intel or ARM Mac)...

> @Jason2866 said:  I provided a PR to espressif fork of GCC (v8.4.0) to build the toolchains native for Apple ARM....

Thanks, but it looks like they are keeping their own ~~hacked~~ patched fork as opposed to a patchset.  I did a quick `git diff` between plain GNU GCC and it's pretty ginormous.  The 8266 setup now is using plain upstream w/a set of ~15 patches (the Xtensa backend seems to have lots of basic bugs).  It'll probably be be easier to migrate forward than untangle things from their repo.  Since I think the 8266 is in quasi-EOL (not for new designs) anyway, less work is better. :)

Jason2866:
@rgov Why prefered to load a not needed and used toolchain? The user is not involved to load accidently the wrong one. The current tool `get.py` does not provide this.

Jason2866:
> Thanks, but it looks like they are keeping their own hacked patched fork as opposed to a patchset. I did a quick git diff between plain GNU GCC and it's pretty ginormous. The 8266 setup now is using plain upstream w/a set of ~15 patches (the Xtensa backend seems to have lots of basic bugs). It'll probably be be easier to migrate forward than untangle things from their repo. Since I think the 8266 is in quasi-EOL (not for new designs) anyway, less work is better. :)

Thx, for the explanation. Do understand and agree fully.

rgov:
> @rgov Why prefered to load a not needed and used toolchain? The user is not involved to load accidently the wrong one. The current tool `get.py` does not provide this either.

It's not necessarily a *problem* to ship a native, single architecture binary – Homebrew only ships these, for example. Especially when you have a tool making the decision for you, it's better than making the end-user pick from a list of downloads.

Advantages include: simplified deployment; files can be copied/synced/shared between Macs of different architectures; libraries can be loaded both by native programs and ones running under Rosetta.

Jason2866:
@earlephilhower Tried to use the toolchains with Platfromio https://github.com/maxgerhardt/platform-raspberrypi/issues/53

The toolchain does install and when calling it crashes. I had the same problem when building the espressif32 toolchains myself on Apple ARM. The issue was caused from the `strip` command. `strip`damaged my builded APPs
Leaving out this step the toolchains do works as expected.

earlephilhower:
Thanks @Jason2866 for the info.  Very weird issue you found and it just has to be a basic bug somewhere...

I removed any `strip` on the M1 and just rebuilt and uploaded new binaries to the same (pre)release 2.2.0.  If you (or anyone else) could give these a try I'd appreciate it!

Jason2866:
@earlephilhower It is working now :-) Great Job!

earlephilhower:
Great, thanks for the feedback!  I would have never thought to drop the `strip` w/o your suggestion.  I'll leave this open for a few days to see if anyone else using the plain IDE can report back and then promote tools@2.2.0 to release and merge this for the next release.

Jason2866:
Feel with you. A lot of work and time you have invested to get it done. Thx again!
But this is great news for all GH Action users!

PR #1977, opened 2024-02-01
Make OpenOCD and Picotool M1 native, too

earlephilhower:
All binaries should now be native on Apple silicon.

### DIFF
--- a/package/package_pico_index.template.json
+++ b/package/package_pico_index.template.json
@@ -286,10 +286,10 @@
                "systems": [
                   {
                      "host": "arm64-apple-darwin",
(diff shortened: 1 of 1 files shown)


earlephilhower:
@maxgerhardt you'll probably want to update the P.IO stuff with this.  The compilation and `otool` edits looks like they went fine but I don't have any way of testing it.  (Also the name will be back to normal naming in the next toolchain release...this was just a hand hack to get things primed.)

maxgerhardt:
Mhh I already have the ARM M1 OpenOCD in PlatformIO (picotool goes unused because it's already provided by PlatformIO-managed packages) ([source](https://github.com/maxgerhardt/platform-raspberrypi/blob/develop/platform.py#L26-L57)).

PR #1986, opened 2024-02-05
Add W5100, W5500, and ECN28J60 interrupt-driven mode

earlephilhower:
No polling needed and should reduce latency by using the GPIO interrupt to signal the Pico to read a received packet.

### DIFF
--- a/cores/rp2040/lwip_wrap.cpp
+++ b/cores/rp2040/lwip_wrap.cpp
@@ -33,12 +33,17 @@
 
 extern void ethernet_arch_lwip_begin() __attribute__((weak));
 extern void ethernet_arch_lwip_end() __attribute__((weak));
+extern void ethernet_arch_lwip_gpio_mask() __attribute__((weak));
+extern void ethernet_arch_lwip_gpio_unmask() __attribute__((weak));
 
 auto_init_recursive_mutex(__lwipMutex); // Only for case with no Ethernet or PicoW, but still doing LWIP (PPP?)
 
 class LWIPMutex {
 public:
     LWIPMutex() {
+        if (ethernet_arch_lwip_gpio_mask)  {
+            ethernet_arch_lwip_gpio_mask();
+        }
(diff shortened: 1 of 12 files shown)


earlephilhower:
FWIW this passes the simple WiFiClient tests and reduces delay on `ping` to ~1.5ms on W5500, ~3.5ms on W5100, but it does seem to have some LWIP re-entrancy issues I need to track down.

earlephilhower:
@JAndrassy this PR should enable you to use IRQ-based reception on your ESP32 WiFi driver as well.  The changes to the driver itself should be minimal and it can really reduce latency and CPU overhead.

JAndrassy:
the esp-hosted firmware maintainers messed up the handshake pin which was originally intended as interrupt signal. 
so Arduino implemented the ESPHost library as polling. I too tried to change it to work with interrupt but the  signal is wrong.

https://github.com/espressif/esp-hosted/issues/308

earlephilhower:
Ah, too bad.  This change may also heolp w/the starvation issue you found with LittleFS in your sketch from #1973.  If you still have the setup and run either WizNet chip, it might be worth re-adding the LittleFS bits and stting if you get any better performance.  With polling it's possible the LWIP lock is always taken when the poll happens...even if it's unlocked at other times.  With level-based IRQs, all you need is 1 clock cycle unlocked and it'll catch the interrupt and poll.

earlephilhower:
I've had this running for >1hr on various boards with `AdvancedWebServer` responding to 15 requests/second over 5 browsers plus an additional 2 `ping`s/second.  Seems as solid as can be and really does improve average latency a ton.

On the ENC I can get <1ms ping response.  On the W5100, if I bump up the SPI from 4MHz to 25MHz I get ping responses of 2ms. Both of which are very stable.  With polling the ping would range from 1-20ms (since the Pico would only look for packets every 20ms!), and the web server response under load felt much worse.

PR #1989, opened 2024-02-06
Update ide.rst

LinusHeu:
I think that info was outdated. 
There's the "Upload Method" menu option to switch between UF2, and Picoprobe. 

Actually, what are the differences between picotool and default(UF2)? Somehow I thought it always used picotool, at least to reset the board. And how can that be selected in platformio?

### DIFF
--- a/docs/ide.rst
+++ b/docs/ide.rst
@@ -1,15 +1,11 @@
 IDE Menus
 =========
 
-Model
+Board
 -----
-Use the boards menu to select your model of RP2040 board.  There will be two
-options:  `Boardname` and `Boardname (Picoprobe)`.  If you want to use a
-Picoprobe to upload your sketches and not the default automatic UF2 upload,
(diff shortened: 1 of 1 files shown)


earlephilhower (approved):

earlephilhower:
> Actually, what are the differences between picotool and default(UF2)? Somehow I thought it always used picotool, at least to reset the board. And how can that be selected in platformio?

UF2 uses serial 1200bps toggle to reset and then sends UF2 up to the new drive.  Picotool uses the special USB custom device code on the Pico (when enabled thru this menu)  and the RPI `picotool` to do the reset to UF2 upload, and then copies over. like the plain UF2  It's really just how the board is reset.

earlephilhower:
Err, actually Picotool does the reset and then the UF2 upload, again thru the custom USB interface.  Mybad!  But, essentially, there's not real reason to use one over the other unless you're having issues.

PR #1996, opened 2024-02-10
Low bitrate whine fix (DMA pacing timer)

keever50:
When using low bitrates like 8000hz or 11025hz as example, PWMAudio will produce a high pitched whine(same as sample rate hz), that sometimes can overwhelm the audio.
This is because the DMA transfers (sample rate) are synced to the PWM wave generation (pwm frequency).

In the RP2040 datasheet under DMA 2.5.5.1, there is a section about pacing timers.
This is a timer that triggers the DREQ every n clock_sys ticks.
Now this is replacing the PWM synced method, which allows the PWM wave generation to have a different frequency than the sample rate.

There is one downside;
The begin() function CAN take quite a few milliseconds worst case to find the perfect fraction for the pacing timer. As the pacer has to be configured with a numerator and a denominator.
It has to find the right fraction based on CPU frequency and bitrate target.

For backwards compatibility and to prevent confusion, I kept the old functions.
setFrequency is still bit rate and setPWMFrequency is to set the PWM.
Now begin has an override so you can select either the usual begin(bitrate) OR begin(bitrate, pwmfreq).

If there is anything to be changed, let me know. (This is my first contribution too!)

Thanks!

### DIFF
--- a/libraries/PWMAudio/src/PWMAudio.cpp
+++ b/libraries/PWMAudio/src/PWMAudio.cpp
@@ -32,8 +32,8 @@ PWMAudio::PWMAudio(pin_size_t pin, bool stereo) {
     _buffers = 8;
     _bufferWords = 0;
     _stereo = stereo;
-    _sampleRate=48000;
-    _pacer=-1;
+    _sampleRate = 48000;
+    _pacer = -1;
 }
 
 PWMAudio::~PWMAudio() {
@@ -94,13 +94,14 @@ bool PWMAudio::setPWMFrequency(int newFreq) {
     return true;
 }
 
-bool PWMAudio::setFrequency(int frequency)
-{
-    if(_pacer<0) return false;
+bool PWMAudio::setFrequency(int frequency) {
+    if (_pacer < 0) {
+        return false;
+    }
     uint16_t _pacer_D, _pacer_N;
     /*Flip fraction(N for D, D for N) because we are using it as sys_clk * fraction(mechanic of dma_timer_set_fraction) for smaller than sys_clk values*/
-    find_pacer_fraction(frequency, &_pacer_D, &_pacer_N); 
-    dma_timer_set_fraction(_pacer, _pacer_N, _pacer_D); 
+    find_pacer_fraction(frequency, &_pacer_D, &_pacer_N);
+    dma_timer_set_fraction(_pacer, _pacer_N, _pacer_D);
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


libraries/PWMAudio/src/PWMAudio.cpp
```diff
@@ -219,3 +245,33 @@ size_t PWMAudio::write(const uint8_t *buffer, size_t size) {
...
+void PWMAudio::find_pacer_fraction(int target, uint16_t *numerator, uint16_t *denominator) {
+    const uint16_t max = 0xFFFF;
+
+    float targetRatio = (float)F_CPU / target;
+    float lowestError = HUGE_VALF;
+    uint16_t bestNum = 1;
+    uint16_t bestDenom = 1;
+    
+    for (uint16_t denom = 1; denom < max; denom++) {
+        Serial.println(denom);
```
earlephilhower, line 258: Can we remove the print statement here, please?

libraries/PWMAudio/src/PWMAudio.cpp
```diff
@@ -219,3 +245,33 @@ size_t PWMAudio::write(const uint8_t *buffer, size_t size) {
     }
     return writtenSize;
 }
+
+void PWMAudio::find_pacer_fraction(int target, uint16_t *numerator, uint16_t *denominator) {
+    const uint16_t max = 0xFFFF;
+
```
earlephilhower, line 251: This does look like it could take a little bit to find the right numerator/denominator.  It's the same problem as the SPI clock change (and others), and I think your method is the cleanest way.

That said. can we cache the last result for numerator/denomiator?  Maybe make something like
````
void PWMAudio::find_pacer_fraction(int target, uint16_t *numerator, uint16_t *denominator) {
    const uint16_t max = 0xFFFF;
    static int  lastTarget;
    static  uint16_t lastNum, lastDen;

    if (target==lastTarget) {
        *numerator = lastNum;
        *denominator = lastDen;
        return;
    }
....
<before exiting the function...>
    lastTarget = target;
    lastNum = bestNum;
    lastDen = bestDenom;

    *numerator = bestNum;
    *denominator = bestDenom;
}
````

I'm thinking that most use cases will have the same frequency if they do multiple PWMAudio samples.  So no need to recalc.  And, worst case, we use 64 bits of RAM and a negligible amount of CPU cycles

earlephilhower (changes requested):
Thanks very much!  I have a couple suggestions to make it better.

I haven't had time to try it out yet.  Is there anything you could give as an example code/sample to show the difference in sound quality?

This will also fail the `astyle` formatting check.  If you're under Linux/Windows WSL you can run `tests/restyle.sh` to clean it up.

libraries/PWMAudio/src/PWMAudio.cpp
keever50: I put 4 samples together in a video(Because i cant upload audio here). First two are after fix and last two before fix.
https://github.com/earlephilhower/arduino-pico/assets/9679272/4b2914d6-5eee-4e04-abee-97fcbb8285c5
Here, two spectrum graphs if it weren't obvious already (Or some got bad hearing)
1 is 8000hz samplerate before fix and 2 is 8000hz after.
image
image

Also thank you for the feedback! I will fix those right away and see what i can do.

keever50:
I added caching with statics as requested, this will skip floating math when its not needed.
Might be useful whenever we want to use fraction calculation more often.

I removed the debugging print.

And i ran the restyle in WSL.

Hope these changes are passing the tests!

earlephilhower (approved):
Awesome work, thanks!

PR #1999, opened 2024-02-12
Update README.md

ardnew:
The `Picoprobe` example showing how to upload to Ubuntu is invalid because Ubuntu does not create or assign users to a `users` group (for many major releases, now). 

The reason the example worked is because the permissions were applied globally, rendering the `GROUP=users` assignment in the rule irrelevant. Hence, this assignment has been dropped.  

The `pico-debug` example has been updated similarly, but it uses a proper group-level rule, and Ubuntu does still use group `plugdev`. 

The reader thus has two good examples of creating `udev` rules.


earlephilhower (approved):
Thx!

PR #2001, opened 2024-02-12
Fix WiFiMulti and ESPhost STA connection w/BSSID

earlephilhower:
WiFiMulti specifies a specific BSSID, in addition to the AP name and password.  In the WiFi core the BSSID is stored as the raw 6-byte MAC address, but the ESPHostedFG firmware expects a formatted C-String (i.e. "ab:cd:ef:01:02:03" instead of {0xab, 0xcd, 0xef, 1, 2, 3})

Convert the raw bytes to the string format expected in the ESP FW.

### DIFF
--- a/libraries/lwIP_ESPHost/src/lwIP_ESPHost.cpp
+++ b/libraries/lwIP_ESPHost/src/lwIP_ESPHost.cpp
@@ -54,7 +54,7 @@ void ESPHostLwIP::setBSSID(const uint8_t *bssid) {
     if (bssid == nullptr) {
         ap.bssid[0] = 0;
     } else {
-        memcpy(ap.bssid, bssid, sizeof(ap.bssid));
+        snprintf((char *)ap.bssid, sizeof(ap.bssid), "%02x:%02x:%02x:%02x:%02x:%02x", bssid[0], bssid[1], bssid[2], bssid[3], bssid[4], bssid[5]);
     }
 }
 


JAndrassy:
it doesn't connect now without bssid specified, because it can't connect to 00:00:00:00:00:00

earlephilhower:
Ah, poop.  Can you try this diff?

My setup is literally two USB based dev boards with multiple USB cables and about 8 dupont wires of different lengths but identical colors between the two and I just cleaned my desk so don't trust the wiring stuck. :(

````
diff --git a/libraries/lwIP_ESPHost/src/lwIP_ESPHost.cpp b/libraries/lwIP_ESPHost/src/lwIP_ESPHost.cpp
index 43f505749..d3b13f46b 100644
--- a/libraries/lwIP_ESPHost/src/lwIP_ESPHost.cpp
+++ b/libraries/lwIP_ESPHost/src/lwIP_ESPHost.cpp
@@ -51,7 +51,7 @@ void ESPHostLwIP::setSSID(const char *ssid) {
 }
 
 void ESPHostLwIP::setBSSID(const uint8_t *bssid) {
-    if (bssid == nullptr) {
+    if (bssid == nullptr || !(bssid[0] | bssid[1] | bssid[2] | bssid[3] | bssid[4] | !bssid[5])) {
         ap.bssid[0] = 0;
     } else {
         snprintf((char *)ap.bssid, sizeof(ap.bssid), "%02x:%02x:%02x:%02x:%02x:%02x", bssid[0], bssid[1], bssid[2], bssid[3], bssid[4], bssid[5]);
````

JAndrassy:
yes it works. only remove the ! next to bssid[5]

PR #2002, opened 2024-02-13
Update the included Adafruit_TinyUSB_Arduino library to 2.4.1

zjwhitehead:
This updates the library to the latest 2.x release as of this PR.


earlephilhower:
Thanks, but it looks like there are some fatal warnings in the library examples and an actual failure in the Platform.IO TinyUSB example build with this version.

zjwhitehead:
You're right, thats strange. I didnt expect any breaking changes but it appears so

maxgerhardt:
Arduino-Pico regular CI build shows that the `-Werror` flag is in the mix ([explicitly set](https://github.com/earlephilhower/arduino-pico/blob/c670f66140e4188aab4a6f1a92644791a89cb153/tests/common.sh#L205), this will hard-abort with the first warning, which the new Adafruit TinyUSB version has due to not initializing a member in a struct.

PlatformIO fails possibly because of what I've already discovered in [this PR](https://github.com/earlephilhower/arduino-pico/pull/1731#issue-1910276037), regarding

>While working on this I also noticed that at least for PlatformIO the include/tusb_config.h file seemed not actually used (you can place a #error in there and it will still compile), which is due to the include order in PlatformIO being that the original Adafruit TinyUSB library's tusb_config.h is first, and then the include/ for Arduino-Pico follows. Resolved by making use of the CFG_TUSB_CONFIG_FILE macro and pointing it to the uniquely named config file.

At the time the configs were probably equivalent, now the standard config in Adafruit has changed, so not maybe anymore.

earlephilhower:
https://github.com/adafruit/Adafruit_TinyUSB_Arduino/pull/386 takes care of the GCC warning/error.  Is there something we can do here with P.IO for the header problem?

earlephilhower:
Closing as Adafruit is up to 3.1 now.

PR #2028, opened 2024-02-27
Add FatFS and FatFSUSB - Wear-Leveled FTL based FAT filesystem for onboard flash

earlephilhower:
Adds a `FatFS` filesystem using an embedded static wear-leveling based FTL to preserve flash life.  Presents the same interface as `LittleFS` and `SDFS`.

Adds a simple way to export onboard FTL to a host allowing drag-and-drop file read/update from the PC, with examples.


earlephilhower:
Differs from #1731 by supporting 512b sectors (the FTL makes this easy w/o rewrites required) and the base `Arduino File` class infrastructure, not the ChaN API.

PR #2036, opened 2024-03-03
lwIP_ESPHost - using interrupt

JAndrassy:
I attempt to make the ESPHost work with interrupt. There are two problems.

First problem is small. The DATA_READY pin is active HGH with esp_hosted. So we need a way to return from RawDev the state for interrupt.

Second problem is big. The semaphores don't work. It is a problem without the interrupt too, but with the interrupt version `handlePackets` doesn't run in async_context so there is no locking.
`ethernet_arch_lwip_begin/end` only guards code executed within async_context so using it elsewhere doesn't stop the code executing. It only blocks handlePackets if executed from the timer worker.

### DIFF
--- a/libraries/lwIP_ESPHost/src/ESPHost.cpp
+++ b/libraries/lwIP_ESPHost/src/ESPHost.cpp
@@ -19,7 +19,8 @@
 */
 
 #include "ESPHost.h"
-#include "CEspControl.h"
+#include <CEspControl.h>
+#include <LwipEthernet.h>
 
 ESPHost::ESPHost(int8_t cs, arduino::SPIClass &spi, int8_t intrpin) {
     (void) cs;
@@ -42,8 +43,10 @@ void ESPHost::end() {
 }
 
 uint16_t ESPHost::sendFrame(const uint8_t *data, uint16_t datalen) {
+    ethernet_arch_lwip_gpio_mask();
     int res = CEspControl::getInstance().sendBuffer(apMode ? ESP_AP_IF : ESP_STA_IF, 0, (uint8_t*) data, datalen);
     CEspControl::getInstance().communicateWithEsp();
+    ethernet_arch_lwip_gpio_unmask();
     return (res == ESP_CONTROL_OK) ? datalen : 0;
 }
 
--- a/libraries/lwIP_ESPHost/src/ESPHost.h
+++ b/libraries/lwIP_ESPHost/src/ESPHost.h
@@ -44,7 +44,7 @@ public:
     }
 
     bool interruptIsPossible() {
-        return false;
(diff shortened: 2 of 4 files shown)


earlephilhower:
For the 2nd problem, how is this different from the other Ethernet drivers?  I'm not sure I understand it.  When using IRQs the callback here does the processing which keeps the LWIP locked:
https://github.com/earlephilhower/arduino-pico/blob/77209a12ccf7c940f92591a0826cea2204c09f47/libraries/lwIP_Ethernet/src/LwipIntfDev.h#L431-L438

The only gotcha is the sendFrame call needs to mask off all Ethernet GPIO IRQs during processing, like here, to avoid Ethernet re-entrancy/deadlock:
https://github.com/earlephilhower/arduino-pico/blob/77209a12ccf7c940f92591a0826cea2204c09f47/libraries/lwIP_w5100/src/utility/w5100.cpp#L280-L281
...all the sendFrame procesing, finishing with...
https://github.com/earlephilhower/arduino-pico/blob/77209a12ccf7c940f92591a0826cea2204c09f47/libraries/lwIP_w5100/src/utility/w5100.cpp#L312-L314

JAndrassy:
I will try to use ` ethernet_arch_lwip_gpio_mask();` to block the IRQ while accessing the library from the main thread. I need to block access to ESPHost library not just from lwip, but from WiFi functions like MACAddress(mac) and RSSI() too.

JAndrassy:
ok. `ethernet_arch_lwip_gpio_mask()` works. thank you

libraries/lwIP_Ethernet/src/LwipIntfDev.h
```diff
@@ -399,7 +399,7 @@ bool LwipIntfDev<RawDev>::begin(const uint8_t* macAddress, const uint16_t mtu) {
         if (RawDev::interruptIsPossible()) {
             noInterrupts(); // Ensure this is atomically set up
             pinMode(_intrPin, INPUT);
-            attachInterruptParam(_intrPin, _irq, LOW, (void*)this);
+            attachInterruptParam(_intrPin, _irq, HIGH, (void*)this);
```
earlephilhower, line 402: Only comment I have would be to make this a function from the template specialization so you can be `high` here and the other Ethernet devices can be `low`.

Did you get any idea how much this helps with latency for, say, `ping`?

JAndrassy:
> Did you get any idea how much this helps with latency for, say, ping?

I didn't have much time. Today I continue the testing and I want to do a PR for the interrupt change state setting, which would be merged before this one.

earlephilhower (approved):
Thanks!

PR #2042, opened 2024-03-06
RawDev::interruptMode

JAndrassy:


libraries/WiFi/src/utility/nodriver.h
```diff
@@ -48,6 +48,10 @@ class NoDriver {
         return true;
     }
 
+    PinStatus interruptMode() {
```
earlephilhower, line 51: Could we standardize these, please?  Some are normal fcns, some are `constexpr`, some are `static constexpr` now.  Compiled code should probably not differ much(at all?) but it's weird to have them all over the map.

OTW, LGTM to do this than your #2036

earlephilhower (changes requested):

libraries/WiFi/src/utility/nodriver.h
JAndrassy: I did the new method in every class to match the style of interruptIsPossible method in that class. there would be more to cleanup. for example interruptIsPossible doesn't do anything useful

earlephilhower: OK, no worries!

earlephilhower (approved):

PR #2058, opened 2024-03-18
increase  CFG_TUD_MSC_EP_BUFSIZE to 4k

ZigZagJoe:
per discussion here https://github.com/ZuluSCSI/ZuluSCSI-firmware/pull/397#discussion_r1529172606

increase CFG_TUD_MSC_EP_BUFSIZE to 4k to allow working with devices with 512 byte sectors in a sane manner.


ZigZagJoe:
wrong repo, sorry

ZigZagJoe:
that said, it would be nice if CFG_TUD_MSC_EP_BUFSIZE is changed to be contained in an #ifndef block to allow overriding it.

earlephilhower:
Unfortunately we can't actually do that because TinyUSB comes from the Pico SDK which is CMake based and pre-built into `libpico.a` because the IDE doesn't understand `Makefile`s.  So one setting needs to be baked in.  Most users don't use the buffer so growing from 64->512 bytes would be somewhat wasteful IMHO.

(FWIW I did a quick test @ 512 bytes with the FatFS PR #2028 and got only ~25% improvement in write speed and ~2x speedup on flash FS reads...which sounds like a lot but with most folks only having 1MB of FS space on the 2MB Pico board it's kind of immaterial.)

What can be done, though, is define that in the Adafruit TinyUSB library that's included.  I think they support setting your own TUSB_CONFIG.h file.

ZigZagJoe:
Agreed, 512 would be wasteful in a majority of cases. The repo this was supposed to go in is concerned with interfacing to a SD card where 512 byte sectors are the norm, so anything smaller increases overhead and complicates code. 

We're using PlatformIO which recompiles the entire framework during our build process, so changing the bundled tusb_config.h works there. At any rate, definitely an outlier for the usual use case. We have a fork of the framework with the appropriate change made (which is where this was SUPPOSED to go) so we're covered. 

I appreciate the response!

PR #2063, opened 2024-03-19
wifi non blocking begin()

keever50:
A very small addition where the **WiFiClass begin()** functions have been _copied_ without the blocking part.
These new functions are **begin_noblock**.

The normal **begin** function blocks the microcontroller until wifi is connected. This can freeze code or trigger a watchdog.
Now using **begin_noblock**, the user can check for **.connected** themselfs.

I hope this helps!
Let me know if there are any feedback or problems


keever50:
Ill fix the checks!

keever50:
Passed the checks and changed the function name to beginNoBlock() like arduino style.

earlephilhower:
I am OK with this addition, but can we do it in a way w/o lots of cut-n-paste?  Maybe have the `::begin` and `::beginNoBlock` just call a private `::beginInternal` function which takes, as a parameter, whether or not it should block?  That way when changes come in there's only 1 spot to do the edits (i.e. we have done mods to `WiFi` for the ESP32-based wifi chips already).

keever50:
I agree! Will do that tomorrow. Good to know you are ok with this addition

keever50:
I didn't actually say it here, but it's so far I know ready. Anything left to change?

earlephilhower:
Let me run it locally first, but give me a day or two, please.

JAndrassy:
sorry, why not just use 0 timeout?

keever50:
> sorry, why not just use 0 timeout?

From my view, timeout==bad, so having a timeout out of the equation makes sense to me personally.

What if the timeout gets implemented somewhere else in the library where it is wanted? Maybe 0 timeout will conflict? A noblock function also leaves this question out of the equation.

I guess it depends on how one looks at it.

Let me know what you think

(Also, I didn't want to sound pushy on previous message. No worries, was making sure things are alright)

libraries/WiFi/src/WiFiClass.cpp
```diff
@@ -67,6 +67,44 @@ void WiFiClass::mode(WiFiMode_t m) {
     }
 }
 
+int WiFiClass::_beginInternal(const char* ssid, const char *passphrase, const uint8_t *bssid) {
```
earlephilhower, line 70: Seems OK, but could this return a `bool` since it's internal and only ever returns 0 or 1 (pass/fail)?

libraries/WiFi/src/WiFiClass.cpp
```diff
@@ -88,40 +135,27 @@ int WiFiClass::beginBSSID(const char* ssid, const uint8_t *bssid) {
...
-    _wifiHWInitted = true;
     uint32_t start = millis(); // The timeout starts from network init, not network link up
-    if (!_wifi.begin()) {
+
+    // Returns WL_IDLE_STATUS on error for compatibility.
+    if (_beginInternal(ssid, passphrase, bssid)) {
         return WL_IDLE_STATUS;
     }
-    noLowPowerMode();
-    // Enable CYW43 event debugging (make sure Debug Port is set)
```
earlephilhower, line 117: Could you please move this commented 2 lines (about CYW43 tracing) to the ::beginInternal method?  I use this very occasionally while debugging the PicoW, but not enough to remember how to do it w/o the prompt. :)

earlephilhower (changes requested):
Minor tweaks, but OTW LGTM.

libraries/WiFi/src/WiFiClass.cpp
```diff
@@ -67,30 +67,16 @@ void WiFiClass::mode(WiFiMode_t m) {
...
-*/
-int WiFiClass::begin(const char* ssid, const char *passphrase, const uint8_t *bssid) {
+bool WiFiClass::_beginInternal(const char* ssid, const char *passphrase, const uint8_t *bssid) {
     // Simple ESP8266 compatibility hack
     if (_modeESP == WIFI_AP) {
-        return beginAP(ssid, passphrase);
+        // When beginAP was a success, it returns WL_CONNECTED. Therefore we return false as no error when that happens.
+        int status = beginAP(ssid, passphrase);
+        if (status != WL_CONNECTED) {
+            return true;
```
earlephilhower, line 76: Err...are you a Linux or bash user?  0(false) for success, !=0(true) for fail is fine for the shell but backwards for most C I've ever seen, and opposite of normal in this code, too.  (i.e. `if !_wifi.begin()) return WL_IDLE_STATUS`).

To keep things consistent, could you please swap the logic so `::_beginInternal` returns `true` on success and `false` on failure?

earlephilhower (changes requested):

libraries/WiFi/src/WiFiClass.cpp
keever50: Woops good point, I will change that soon! 

I usually chose to return _above_ 0 as error, because you usually can have many different errors. So you can identify them if needed.
Of course theoretically you can have many different successes as well, but these don't matter too much 😄 

This is the reason for me returning an int, not a bool and above 0 as error. _But now I mixed them up_! It does not matter that much anyway.
However, for consistency I will change it. Inconsistent code is annoying :)

For the future, I will make sure I will double check consistency. Arduino style is not something I come from ^^

earlephilhower (dismissed):
Actually, now that I'm back home and checking this out, it seems this is changing all the return values from the `wl_definitions` into `bool` which is not appropriate and will break things.  I'll see if I can undo the changes in the GH editor...

keever50:
Thanks for the change! I got really confused after changing the error logic a couple times lol. I agree with you entirely.

keever50:
I replied before and exactly on the same time when you were posting your message. Yep, I got a little confused with the error logic as well. Seems like the SDK kinda tends to prefer my previous way of handling errors. Return int as error, because there are more errors than successes. But Arduino really likes bools... I am not entirely sure what is the right method at the moment.

earlephilhower (approved):
OK, on further re-read I see it does still return `wl_definition` values.  Mybad!

PR #2069, opened 2024-03-22
Add 128MHz frequency

jlbirccyn:
Hello,

I need to generate a 16MHz clock on a PWM output. So I added a 128MHz clock frequency. Generally speaking, I think that having a clock frequency that is a power of 2 is a good thing. 

Best regards


earlephilhower (approved):
LGTM!

jlbirccyn:
Thanks !

PR #2098, opened 2024-04-10
add native text to TinyUSB Host menu option

hathach:
Add `(native)` text to menu option -> "Adafruit TinyUSB Host (native)". 
The reason is the previous text is stil ambigous to some users. Some may think this enable pio-usb host option (which is more popular host option than native).

IDE menu change after the PR
image
 
@ladyada


earlephilhower (approved):
Thanks!

PR #2121, opened 2024-04-24
Shift I2S input data by 1 bit

### LINKED ISSUE #2037: PCM1808 Does Not Work Correctly Using I2S Unless I Shift the Result by Nine Bits.  I Expected Eight.
I believe the shift should be eight bits, but it does not work.  Nine bits does though.  Here are some screenshots. 
image
The screenshot above is with eight-bit shifts.   The one below is with nine and is the waveform I'm converting.
image
I just changed BIT_SHIFT in this code to obtain the two screenshots.
```
/*
 * I believe the 9-bit shift is incorrect and should be eight.  
*/
#include <I2S.h>

#define RATE 16000
#define MCLK_MULT 256  // 384 for 48 BCK per framme,  256 for 64 BCK per frame
#define BIT_SHIFT 8

I2S i2s(INPUT);


void setup() {
  Serial.begin();

  i2s.setDATA(2);
  i2s.setBCLK(0); 
  i2s.setMCLK(3);
  // Note: LRCK pin is BCK pin plus 1 (1 in this case).
  i2s.setBitsPerSample(24);
  i2s.setFrequency(RATE);
  i2s.setMCLKmult(MCLK_MULT);
  i2s.setSysClk(RATE);
  i2s.setBuffers(32,0,0);
  
  i2s.begin();
}

void loop() {
  static int32_t r, l;
  i2s.read32(&l, &r);
  Serial.printf("%d %d \r\n", r<<BIT_SHIFT, l<<BIT_SHIFT);
}
```
I can supply schematic diagrams and am willing to test things out.  It makes me nervous when things do not work as expected.

Thanks,

Rob

earlephilhower:
Fixes #2037

### DIFF
--- a/libraries/I2S/src/pio_i2s.pio
+++ b/libraries/I2S/src/pio_i2s.pio
@@ -294,6 +294,9 @@ static inline void pio_i2s_in_program_init(PIO pio, uint sm, uint offset, uint d
     pio_sm_set_pins(pio, sm, 0); // clear pins
 
     pio_sm_exec(pio, sm, pio_encode_set(pio_y, bits - 2));
+
+    // Offset the data by 1 bit
+    pio_sm_exec(pio, sm, pio_encode_in(pio_pins, bits - 1));
 }
 
 %}
--- a/libraries/I2S/src/pio_i2s.pio.h
+++ b/libraries/I2S/src/pio_i2s.pio.h
@@ -378,6 +378,8 @@ static inline void pio_i2s_in_program_init(PIO pio, uint sm, uint offset, uint d
     pio_sm_set_pindirs_with_mask(pio, sm, pin_mask, pin_mask);
     pio_sm_set_pins(pio, sm, 0); // clear pins
     pio_sm_exec(pio, sm, pio_encode_set(pio_y, bits - 2));
+    // Offset the data by 1 bit
+    pio_sm_exec(pio, sm, pio_encode_in(pio_pins, bits - 1));
 }
 
 #endif


frohro:
@earlephilhower I am happy to report you fixed the issue perfectly according to my test.  Just in case I made a mistake here is what I did.   There are parts of this I have never done before. 
1. I changed the platformio.ini to include this line:  platform_packages =
   framework-arduinopico@https://github.com/earlephilhower/arduino-pico.git#i2sin1
2. I connected my pico w and built and uploaded with the -> button in VScode PlatformIO.  It took a long time reconfiguring the PlatformIO framework but finally finished.
3. Then I changed my code so it was as we thought it should be, shifting 8 bits for 32 bit encoded packets or 0 bits if I am using 24 bit encoded packets.  That now looks like this:  
```  BufferFiller::BufferFiller(CircularBufferQueue &q) : queue(q) {}

void BufferFiller::fillBuffer()
{
    while (!mutex_try_enter(&my_mutex, &mutex_save))
    {
        // Mutex is locked, so wait here.
    }
    char *buffer = queue.getNextBufferAndUpdate(true);
    mutex_exit(&my_mutex);
    if (buffer != nullptr)
    {
        static int32_t r, l, packet_number = 0;
        uint32_t bufferIndex = 4;
        while (bufferIndex < BUFFER_SIZE - 4) // Leave space for the packet number
        {
            i2s.read32(&l, &r);
            if (BITS_PER_SAMPLE_SENT == 24)
            {
                // l = l << 1;
                // r = r << 1;
                l &= 0xFFFFFF;                       // Keep only the lower 24 bits
                r &= 0xFFFFFF;                       // Keep only the lower 24 bits
                memcpy(buffer + bufferIndex, &l, 3); // Copy only the lower 3 bytes
                bufferIndex += 3;
                memcpy(buffer + bufferIndex, &r, 3); // Copy only the lower 3 bytes
                bufferIndex += 3;
            }
            else if(BITS_PER_SAMPLE_SENT == 32)
            {
                // l = l << 9;
                // r = r << 9;
                l = l << 8; // Keep only the lower 24 bits
                r = r << 8; // Keep only the lower 24 bits
                memcpy(buffer + bufferIndex, &l, sizeof(int32_t)); // Copy only the lower 3 bytes
                bufferIndex += sizeof(int32_t);
                memcpy(buffer + bufferIndex, &r, sizeof(int32_t)); // Copy only the lower 3 bytes
                bufferIndex += sizeof(int32_t);
            }
        }
        memcpy(buffer, &packet_number, sizeof(int32_t));  
        packet_number++;
        Serial.printf("Filled %d\n", packet_number);
    }
}
```
4.  I sent it via UDP to my Linux box, received and plotted it there using a python script that was working just fine with the extra bit shift and old framework.  The plot looks like this:  
image
and this is what I was expecting, as I have a sinewave at 1 kHz and 2 V amplitude feeding both left and right channels.
That was the test.

I would appreciate your advice about when I should change frameworks, and what I should change to.

Thank you so much for fixing this bug for me!!

Rob

earlephilhower:
Thanks for the update!  With the waveform captured it was pretty obvious what was going on.

You should probably use the `git` version (I think you need to just drop the `#xxx`) until the next release to get this minor patch.

PR #2124, opened 2024-04-25
Avoid swapping L/R channels on I2S input

### LINKED ISSUE #2123: I2S Right Channel is One Sample Shifted in Time from Left
Hi @earlephilhower ,
My testing today seems to indicate that there is a one time sample shift between the right and left channels using the latest master branch.  It also seems to be true with the release branch using the Arduino IDE (using 3.7.2 shifting the data left 9 bits instead of 8).  Here is what a 1 kHz wave looks like sampled at 48 kHz. image  And here is what a24 kHz wave looks like sampled at 48 kHz.  Screenshot from 2024-04-24 22-02-52  Notice every other sample is 180 degrees out of phase with the previous.  The overall amplitude goes up and down sinusoidally because the sampling rate is not exactly 48 kHz due to the Pico clock frequency.  Here is the example program I used to test this.
The discontinuities are because the Serial.printf() cannot always keep up with the 48 kHz read of new data.

Here is the code for the latest master branch.
```
#include <I2S.h>
#include <Arduino.h>

#define RATE 48000
#define MCLK_MULT 256

I2S i2s(INPUT);

void setup() {
  Serial.begin();
  i2s.setDATA(2); // These are the pins for the data on the SDR-TRX
  i2s.setBCLK(0);
  i2s.setMCLK(3);
  // Note: LRCK pin is BCK pin plus 1 (1 in this case).
  i2s.setSysClk(RATE);
  i2s.setBitsPerSample(24);
  i2s.setFrequency(RATE);
  i2s.setMCLKmult(MCLK_MULT);
  i2s.setBuffers(32, 0, 0);
  i2s.begin();

  while (1) {
    int32_t l, r;
    i2s.read32(&l, &r);
    l = l << 8;
    r = r << 8;
    Serial.printf("%d %d\r\n", l, r);
  }
}

void loop() {}
```

earlephilhower:
Fixes #2123

### DIFF
--- a/libraries/I2S/src/pio_i2s.pio
+++ b/libraries/I2S/src/pio_i2s.pio
@@ -295,8 +295,8 @@ static inline void pio_i2s_in_program_init(PIO pio, uint sm, uint offset, uint d
 
     pio_sm_exec(pio, sm, pio_encode_set(pio_y, bits - 2));
 
-    // Offset the data by 1 bit
-    pio_sm_exec(pio, sm, pio_encode_in(pio_pins, bits - 1));
+    pio_sm_exec(pio, sm, pio_encode_in(pio_pins, bits)); // Shift in 1st L data
+    pio_sm_exec(pio, sm, pio_encode_in(pio_pins, bits - 1)); // Shift in 1st R data modulo one bit, avoiding bit shift from #2037
 }
 
 %}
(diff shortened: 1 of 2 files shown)


frohro:
I confirm this fixes the issue.  Thank you @earlephilhower !!

Issue #2127, opened 2024-04-26
Feature Request: Integration of class MacAddress from arduino-esp32?


matthias-bs:
Hi,

how about integrating https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/MacAddress.cpp?
As far as I can see, this should work right out of the box.

Best regards,
Matthias

earlephilhower:
Why not try your hand at a PR for it?  Looks like a pretty trivial porting effort, if any, would be needed.  It might be a good idea to update the `WiFi` and LWIP `Ethernet` classes to include support for this object, too.  OTW I'm not seeing how it'd be used...

PR #2130, opened 2024-04-27
Feature Request: Add MacAddress class from arduino-esp32

matthias-bs:
As discussed in https://github.com/earlephilhower/arduino-pico/issues/2127

### DIFF
--- a/cores/rp2040/MacAddress.cpp
+++ b/cores/rp2040/MacAddress.cpp
@@ -1,0 +1,228 @@
+#include <MacAddress.h>
+#include <stdio.h>
+#include <Print.h>
+
+//Default constructor, blank mac address.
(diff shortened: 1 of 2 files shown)


earlephilhower:
@matthias-bs thanks, but after looking at this it just seems kind of trivial.  I was expecting more meat (like validation, creation of unique virtual ones, hookup into the rest of the libraries which depend on MACs, etc.), but I see that it's unused even in the ESP32 core (outside of 1 ESPNOW example).

Given that, I think it would be best not to put this in the core here.  But you should be able to include the files in your own sketches just by copying them into the folder.  The IDE will auto-build them for you and your sketch should be able to just include the header and go.  Good luck, and thanks!

PR #2133, opened 2024-04-29
functions to set PDM pins

clhuang:
The current PDM class is not very user-friendly, as it requires you to modify some header file in order to set PIN_PDM_DIN/PIN_PDM_CLK (and there is no documentation on how to do so). Add some defaults for these defines, as well as a method to set them at runtime.


earlephilhower (approved):
Thanks!

PR #2134, opened 2024-04-29
BTstackLib: added function to set scan respone data.

djpearman:
Hi,

I added a function to set the BLE scan response data via BTstackManager - essentially just a wrapper around the underlying BTstack function. The passed data must be structured in the same manner as advertising data.

Best regards,
djpearman


earlephilhower (approved):
Seems fine, but man does the converting of a `const uint8_t *` to a non-const `uint8_t *` bother me.  But I get that's the way they wrote everything. :(

PR #2137, opened 2024-04-30
BTstack: remove superfluous call during setup.

djpearman:
Hi,

The call to `le_device_db_init()` seems to superfluous as it is called in the subsequent `sm_init()`. It is not present in BTstack examples I've seen, for example:

https://github.com/bluekitchen/btstack/blob/72ef1732c954d938091467961e41f4aa9b976b34/example/le_streamer_client.c

Removing has no apparent effects (good or bad) in my projects.

Best regards,
djpearman


earlephilhower (approved):
Looks good.  The BTStackLib is kind of old and not supported by the BTStack guys anymore, so it's not surprising that it's got a little cruft collected in it. :)

Issue #2129, opened 2024-04-27
Fix enum redeclaration issue when using WiFi and Ethernet


rpineau:
If we don't prevent the re-declaration of the enum this triggers
    a compiler  error as LwipIntfDev.h get included by both WiFi.h
    and the Ethernet modules (lwIP_w5500/src/W5500lwIP.h for example).
Without this, it's not possible to use the pico wifi and a W5500 

`diff --git a/libraries/lwIP_Ethernet/src/LwipIntfDev.h b/libraries/lwIP_Ethernet/src/LwipIntfDev.h
index 4bf7aab..3e548f7 100644
--- a/libraries/lwIP_Ethernet/src/LwipIntfDev.h
+++ b/libraries/lwIP_Ethernet/src/LwipIntfDev.h
@@ -50,11 +50,14 @@
 #define DEFAULT_MTU 1500
 #endif

+#ifndef __EthernetLinkStatus__
+#define __EthernetLinkStatus__
 enum EthernetLinkStatus {
     Unknown,
     LinkON,
     LinkOFF
 };
+#endif

 extern "C" void cyw43_hal_generate_laa_mac(__unused int idx, uint8_t buf[6]);`

earlephilhower:
Looks like you found a bug and a solution.  Want to make a PR?

rpineau:
Yes.
I guess I need to fork this first, make my change and make a PR from it ? (as I don't have commit rights).

JAndrassy:
sorry, but what error do you get? there is `#pragma once` to guard this

earlephilhower:
I just `grep`ped for that identifier and it's only defined in 1 spot in the entire tree.  In addition to @JAndrassy 's comments that the suggested diff doesn't do what you think it does, it also would break the `enum` typename.

````
earle@amd:~/Arduino/hardware/pico/rp2040$ grep -r EthernetLinkStatus
libraries/lwIP_Ethernet/src/LwipIntfDev.h:enum EthernetLinkStatus {
libraries/lwIP_Ethernet/src/LwipIntfDev.h:    EthernetLinkStatus linkStatus();
libraries/lwIP_Ethernet/src/LwipIntfDev.h:EthernetLinkStatus LwipIntfDev<RawDev>::linkStatus() {
````

Closing for now, but if you have an MCVE showing how it fails post it here and we'll look.

rpineau:
[...]

PR #2147, opened 2024-05-06
Ethernet legacy API compatibility layer

JAndrassy:
Ethernet legacy API compatibility layer EthernetCompat.h is from esp8266 core.

The example is from the Arduino Ethernet library with include changed and Ethernet object instance created.

For cases like this https://github.com/earlephilhower/arduino-pico/issues/2129#issuecomment-2081217703


earlephilhower (approved):
Thanks!

Issue #2132, opened 2024-04-28
WiFiServer and FreeRTOS - Repeatedly accepting connection eventually causes freeze


JeronimusII:
When using the WiFiServer while using FreeRTOS to repeatedly accept a new connection it eventually at a seemingly random time (usually 1-10 minutes with a new connection every 3 seconds) to hang up both cores right after a new incoming connection (even before calling **WiFiServer::accept()**).

By pausing using a debugger and checking the call stack I figured that it is stuck in an infinite loop by FreeRTOS due to a fatal error allocating memory in the **WiFiServer::_accept()** method at line 197:
https://github.com/earlephilhower/arduino-pico/blob/a49bcd4a9535bc9615ce46b67e91d50d501c07d0/libraries/WiFi/src/WiFiServer.cpp#L197

Which is caused by an assertion in the **xQueueSemaphoreTake()** method of FreeRTOS that expects the task scheduler not to be suspended:
https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/8e07366994f81354a2d4556ca1da9f73dab781e6/queue.c#L1672-L1677

The task scheduler seems to be suspended because the IRQ happened in the brief period the **vTaskDelay()** method called by some USB loop suspends the task scheduler.

I am running the program on a **Raspberry Pi Pico W**.
I tried it with a second board in case I might have fried something.

Here is the source of program that can reproduce the issue:
(Freeze happens during the delayMicroseconds() at the end of loop())
```c++
#include "FreeRTOS.h"
#include <WiFi.h>


#define WIFI_SSID "ssid"
#define WIFI_PASS "password"


WiFiServer server(5000);


void setup() {
    Serial.begin(115200);

    WiFi.mode(WIFI_STA);

    while (!Serial) {}
    delay(500);
    Serial.println("Start");

    uint8_t status = WiFi.begin(WIFI_SSID, WIFI_PASS);
    while (status == WL_DISCONNECTED) {
        delayMicroseconds(500000);
        status = WiFi.status();
    }
    if (status == WL_CONNECTED) {
        Serial.printf(
            "Connected\n"
            "   SSID: %s\n"
            "   IP Address: %s\n"
            "   BSSID: ",
            WiFi.SSID().c_str(),
            WiFi.localIP().toString().c_str()
        );
        uint8_t bssid[6]; 
        WiFi.BSSID(bssid);
        for (uint8_t i = 0; i < 6; i++) {
            if (i > 0) {
[...]

PR #2155, opened 2024-05-10
Disable IRQs when FreeRTOS task switching disabled

earlephilhower:
Avoids crash seen in #2132


earlephilhower:
@jeronimusii, can you give this a try and report back?  Note that you'll need to update the FreeRTOS library submodule after pulling to get the change.

PR #2259, opened 2024-07-12
Fix D pins mapping on Adafruit IB RP2040

AcThPaU:
Fix D pins mapping on Adafruit itsybitsy RP2040 and Adafruit Feather RP2040

### DIFF
--- a/variants/adafruit_feather/pins_arduino.h
+++ b/variants/adafruit_feather/pins_arduino.h
@@ -41,4 +41,9 @@
 #define SPI_HOWMANY    (1u)
 #define WIRE_HOWMANY   (1u)
 
+// D pins
+#define D4             (6u)
+#define D5             (7u)
+#define D6             (8u)
+
 #include "../generic/common.h"
--- a/variants/adafruit_feather/pins_arduino.h
+++ b/variants/adafruit_feather/pins_arduino.h
@@ -42,8 +42,8 @@
 #define WIRE_HOWMANY   (1u)
 
 // D pins
-#define D4             (6u)
-#define D5             (7u)
-#define D6             (8u)
+#define __PIN_D4             (6u)
+#define __PIN_D5             (7u)
+#define __PIN_D6             (8u)
 
 #include "../generic/common.h"
--- a/variants/adafruit_itsybitsy/pins_arduino.h
+++ b/variants/adafruit_itsybitsy/pins_arduino.h
@@ -41,15 +41,15 @@
 #define WIRE_HOWMANY   (2u)
 
 // D pins
-#define D2             (12u)
-#define D3             (5u)
-#define D4             (4u)
-#define D5             (14u)
-#define D7             (6u)
-#define D9             (7u)
-#define D10            (8u)
-#define D11            (9u)
-#define D12            (10u)
-#define D13            (11u)
+#define __PIN_D2             (12u)
+#define __PIN_D3             (5u)
+#define __PIN_D4             (4u)
+#define __PIN_D5             (14u)
+#define __PIN_D7             (6u)
+#define __PIN_D9             (7u)
+#define __PIN_D10            (8u)
+#define __PIN_D11            (9u)
+#define __PIN_D12            (10u)
+#define __PIN_D13            (11u)
 
 #include "../generic/common.h"
--- a/variants/adafruit_itsybitsy/pins_arduino.h
+++ b/variants/adafruit_itsybitsy/pins_arduino.h
@@ -40,4 +40,16 @@
 #define SPI_HOWMANY    (1u)
 #define WIRE_HOWMANY   (2u)
 
+// D pins
+#define D2             (12u)
+#define D3             (5u)
+#define D4             (4u)
+#define D5             (14u)
+#define D7             (6u)
+#define D9             (7u)
+#define D10            (8u)
+#define D11            (9u)
+#define D12            (10u)
+#define D13            (11u)
+
 #include "../generic/common.h"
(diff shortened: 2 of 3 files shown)


variants/adafruit_feather/pins_arduino.h
```diff
@@ -41,4 +41,9 @@
 #define SPI_HOWMANY    (1u)
 #define WIRE_HOWMANY   (1u)
 
+// D pins
+#define D4             (6u)
+#define D5             (7u)
+#define D6             (8u)
+
```
earlephilhower, line 48: Looking at the schematic here https://cdn-learn.adafruit.com/assets/assets/000/116/896/original/adafruit_products_FeatherRP2040_sch.png?1670021606 it looks like there are D7-D13 and D24-D25.  Can you add those here, too?

variants/adafruit_itsybitsy/pins_arduino.h
```diff
@@ -40,4 +40,16 @@
...
+#define D3             (5u)
+#define D4             (4u)
+#define D5             (14u)
+#define D7             (6u)
+#define D9             (7u)
+#define D10            (8u)
+#define D11            (9u)
+#define D12            (10u)
+#define D13            (11u)
+
```
earlephilhower, line 54: Also looks like D24-D25 are included per the schematic here https://cdn-learn.adafruit.com/assets/assets/000/101/925/original/adafruit_products_ItsyRP_sch.png?1619632724

earlephilhower (changes requested):
Thanks!  Could you please button up the last bits and then we can merge!

variants/adafruit_feather/pins_arduino.h
AcThPaU: D7-8 are not included, and D9-13, 24,25 are mapped correctly

variants/adafruit_itsybitsy/pins_arduino.h
AcThPaU: D24 and D25 are mapped correctly
https://github.com/adafruit/Adafruit-ItsyBitsy-RP2040-PCB/blob/main/Adafruit%20ItsyBitsy%20RP2040%20pinout.svg

variants/adafruit_feather/pins_arduino.h
earlephilhower: Actually, that points to another problem.  This `#define` doesn't compile at all.  The issue is the variant pins.h for both chips include "../generic/common.h`.  In there they define
````
static const uint8_t D0 = (0u);
static const uint8_t D1 = (1u);
static const uint8_t D2 = (2u);
static const uint8_t D3 = (3u);
static const uint8_t D4 = (4u);
....
static const uint8_t D27 = (27u);
static const uint8_t D28 = (28u);
static const uint8_t D29 = (29u);
````

So your `#define D4 (6u)` ends up making a line that compiles as
````
static const uint8_t (6u) = (4u);
````

I just tried it locally and it does fail as expected.

````
In file included from /home/earle/Arduino/hardware/pico/rp2040/cores/rp2040/Arduino.h:30,
                 from /tmp/arduino_build_712534/sketch/sketch_jul12a.ino.cpp:1:
/home/earle/Arduino/hardware/pico/rp2040/variants/adafruit_feather/pins_arduino.h:45:25: error: expected unqualified-id before numeric constant
   45 | #define D4             (6u)
      |                         ^~
/home/earle/Arduino/hardware/pico/rp2040/variants/adafruit_feather/../generic/common.h:20:22: note: in expansion of macro 'D4'
   20 | static const uint8_t D4 = (4u);
      |                      ^~
/home/earle/Arduino/hardware/pico/rp2040/variants/adafruit_feather/pins_arduino.h:45:25: error: expected ')' before numeric constant
   45 | #define D4             (6u)
      |                        ~^~
/home/earle/Arduino/hardware/pico/rp2040/variants/adafruit_feather/../generic/common.h:20:22: note: in expansion of macro 'D4'
   20 | static const uint8_t D4 = (4u);
      |                      ^~
````

earlephilhower: What needs to be done is either define all D/A pins and SDA/SCL/etc. (everything in common.h) in each variant (the hacky way) or to rewrite the common to use a predefined constant if defined for each pin.  I.e.
````
...
#ifdef __D4
static const uint8_t D4 = __D4;
#else
static const uint8_t D4 = (4u);
#endif
...
````

There may be a cleaner way of doing this, but off the top of my head I think it would work.

dhalbert:
@ladyada Someone from Adafuit should look this over.

variants/adafruit_feather/pins_arduino.h
earlephilhower: Actually, the suggestion was already implemented for the ADC pins!  Been so long I forgot it, but we have a `__PIN_XX` define

````
...
#ifdef __PIN_A3
static const uint8_t A3 = __PIN_A3;
#else
static const uint8_t A3 = (29u);
#endif
...
````

Doing the digital pins in the same way would be relatively small technical debt I think...

AcThPaU: thanks! I added changes to common.h

variants/adafruit_itsybitsy/pins_arduino.h
ladyada: thanks!

AcThPaU:
OK. I added fix to itsybitsy with the format that new common.h should accept. I need to sleep for the day, and maybe people from Adafruit can fix the rest of the boards.

earlephilhower (approved):
Thanks!  I fixed the 3 Feather pin define names and will give a sanity check and CI run on this now.

ladyada:
reviewd, looks good!

PR #2263, opened 2024-07-12
Add board EVN Alpha

HTY2003:
Hi,

I'd like to add this board (EVN Alpha) to the core. Let me know if there are any issues and I'll try to resolve them promptly.

Thanks for the help!


earlephilhower:
Looks like you ran this on a Windows machine and it changed the linefeeds so now it has a 20K+ line diff.

Please pull latest master and re-run `makeboards.py`.  I believe #2264 should ensure only `\n` is used, even on Windows.

HTY2003:
Should be fixed now. Thanks!

PR #2266, opened 2024-07-13
Cannot switch from AP mode to station mode

andnic:
Clear LwipIntfDev::_netif in LwipIntfDev::end() to allow switching from AP to station modes.

Similar to #884, after starting AP mode it is not possible to switch to station mode. The call to WiFi.begin() completes instantaneously and the AP IP address persists.

Taking clues from #905, the connection state is derived from the presence of an IP address in LwipIntfDev::_netif. Zeroing out _netif in LwipIntfDev::end() remedies the issue and allows switching from station to AP and back successfully.


JAndrassy:
.end() should not clear the static IP configuration so next begin() can use it. 
please try to find a different way to make AP mode work

PR #2268, opened 2024-07-16
Add Elecrow CrowPanel RP2040 DVI board

qq71680264:
board link:https://www.elecrow.com/crowpanel-pico-display-4-3-inch-320-240-hmi-module-tft-lcd-touchscreen-with-rp2040-support-c-c-micropython-lvgl.html

i have already tested this SDK file, it works smoothly with this board.


earlephilhower (changes requested):
Looks like makeboards.py was taken from a much older version, resulting in many diffs in boards.txt (and makeboards, too).  Can you rebase against the current master and re-run makeboards to get the correct boards.txt?

PR #2269, opened 2024-07-16
Add Elecrow CrowPanel RP2040 DVI board

qq71680264:
I've generated boards.txt via makeboards.py in the latest code, added the required files, and tested it to run silky smooth on an Elecrow CrowPanel RP2040 DVI board. This is the first time I've been involved in submitting code, so if I haven't gotten it right somewhere, please remind me and I'll do my best to get it right.
board link：https://www.elecrow.com/crowpanel-pico-display-4-3-inch-320-240-hmi-module-tft-lcd-touchscreen-with-rp2040-support-c-c-micropython-lvgl.html


tools/makeboards.py
```diff
@@ -435,6 +435,9 @@ def MakeBoardJSON(name, vendor_name, product_name, vid, pid, pwr, boarddefine, f
 # EVN
 MakeBoard("evn_alpha", "EVN", "Alpha", "0x2e8a", "0xf00a", 500, "EVN_ALPHA", 16, "boot2_generic_03h_2_padded_checksum", board_url="https://coresg.tech/evn")
 
+# Elecrow
```
earlephilhower, line 438: Alphabetically, this should be placed before ElectronicCat.

earlephilhower (changes requested):
The CI run shows a difference in multiple files from what you've checked in.  You should never edit anything other than makeboards.py and readme.md to add a board.  `Boards.txt` and the JSON files are machine generated.

qq71680264:
Oh, i read the "contrib.rst", and maybe i know how to do it now. The only files that need to be added manually are the ones in the “variants” folder, the rest are generated by makeboards.py, is that correct?

earlephilhower:
> Oh, i read the "contrib.rst", and maybe i know how to do it now. The only files that need to be added manually are the ones in the “variants” folder, the rest are generated by makeboards.py, is that correct?

Yes, that's pretty much it.  The `makeboards.py` script updates the package JSON, the P.IO JSONs, and boards.txt.  None of those outputs should be edited.

PR #2274, opened 2024-07-17
Add undervolting/overvolting according to CPU frequency

DatanoiseTV:
This adds adjustment of the core voltage according to the results from
https://forums.raspberrypi.com/viewtopic.php?p=1820686&sid=a1bc42fab114da743f958a5dbf160925#p1820686

This way the CPU cores will run at the VCORE for the corresponding frequencies.
This allows for operation above 400MHz.

### DIFF
--- a/cores/rp2040/main.cpp
+++ b/cores/rp2040/main.cpp
@@ -22,6 +22,7 @@
 #include "RP2040USB.h"
 #include <pico/stdlib.h>
 #include <pico/multicore.h>
+#include <hardware/vreg.h>
 #include <reent.h>
 
 RP2040 rp2040;
--- a/cores/rp2040/main.cpp
+++ b/cores/rp2040/main.cpp
@@ -80,7 +80,36 @@ static struct _reent *_impure_ptr1 = nullptr;
 
 extern "C" int main() {
 #if F_CPU != 125000000
-    set_sys_clock_khz(F_CPU / 1000, true);
+
+    // Optimal core frequency (under and overvolting) based on
+    // https://forums.raspberrypi.com/viewtopic.php?p=1820686&sid=a1bc42fab114da743f958a5dbf160925#p1820686
+    uint32_t frequency_khz = F_CPU / 1000;
+    if (frequency_khz <= 120000) {
+        vreg_set_voltage(VREG_VOLTAGE_0_85);
+    } else if (frequency_khz <= 150000) {
+        vreg_set_voltage(VREG_VOLTAGE_0_90);
+    } else if (frequency_khz <= 180000) {
+        vreg_set_voltage(VREG_VOLTAGE_0_95);
+    } else if (frequency_khz <= 210000) {
+        vreg_set_voltage(VREG_VOLTAGE_1_00);
+    } else if (frequency_khz <= 240000) {
+        vreg_set_voltage(VREG_VOLTAGE_1_05);
+    } else if (frequency_khz <= 270000) {
+        vreg_set_voltage(VREG_VOLTAGE_1_10);
+    } else if (frequency_khz <= 300000) {
+        vreg_set_voltage(VREG_VOLTAGE_1_15);
+    } else if (frequency_khz <= 330000) {
+        vreg_set_voltage(VREG_VOLTAGE_1_20);
+    } else if (frequency_khz <= 360000) {
+        vreg_set_voltage(VREG_VOLTAGE_1_25);
+    } else if (frequency_khz <= 402000) {
+        vreg_set_voltage(VREG_VOLTAGE_1_30);
+    } else {
+        // If the frequency is out of the safe range, set to the highest safe voltage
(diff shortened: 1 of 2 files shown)


tools/makeboards.py
```diff
@@ -34,9 +34,10 @@ def BuildDebugLevel(name):
         print("%s.menu.dbglvl.%s.build.debug_level=%s" % (name, l[0], l[1]))
 
 def BuildFreq(name):
-    for f in [ 133,  50, 100, 120, 125, 128, 150, 175, 200, 225, 240, 250, 275, 300]:
+    for f in [ 133,  50, 100, 120, 125, 128, 150, 175, 200, 225, 240, 250, 275, 300, 402, 436]:
         warn = ""
         if f > 133: warn = " (Overclock)"
+        if f > 300: warn = " (Experimental, not all chips supported.)"
```
earlephilhower, line 40: This is kind of redundant, no?  "Overclock" is already saying no guarantees IMHO.

cores/rp2040/main.cpp
```diff
@@ -80,7 +81,36 @@ static struct _reent *_impure_ptr1 = nullptr;
 
 extern "C" int main() {
 #if F_CPU != 125000000
-    set_sys_clock_khz(F_CPU / 1000, true);
+
+    // Optimal core frequency (under and overvolting) based on
+    // https://forums.raspberrypi.com/viewtopic.php?p=1820686&sid=a1bc42fab114da743f958a5dbf160925#p1820686
+    uint32_t frequency_khz = F_CPU / 1000;
+    if (frequency_khz <= 120000) {
```
earlephilhower, line 88: I'm very worried about changing the voltage for in-spec operation.  The RPI guys picked a number after, I assume, some serious QC and binning to ensure their boards were solid over the entire temp range and component variations.  That post doesn't include anything about temperature schmooing, and seems to be only on a single part.

What I've found more limiting is the flash SPI speeds at these crazy high multipliers (we're making the [Celeron 300A](https://www.tomshardware.com/picturestory/636-best-overclocking-cpu.html) look like a snail!).  The QSPI is taken as a divisor of the system clock, so there's a 1:1 increase in it and eventually the flash chip doesn't operate properly.

Would you be able to bracket this voltage tweaking under a menu item in the `generic` device?  Since you'll need to tweak the flash divisor in most cases at these high multipliers, an `Over/Undervolt` menu with a `Default Voltage` and an `Auto Over/Undervolt` item would let users take advantage of this

earlephilhower: Another possible option would be a `Voltage` menu with the items `Default` and `Auto` (uses the logic here), `0.85V (<=120MHz)`, `0.90V (<=150MHz)`, ... so that the user would be able to manually select one but have some guidance.

earlephilhower (changes requested):

cores/rp2040/main.cpp
earlephilhower: One more gotcha, when determining the Vin for a device you normally run a test app that turns on all peripherals and burns as much power as possible (since as the I increases, the V on-chip will drop because of the power plane resistance).  His test wasn't doing anything like that AFAICT.  So it would give a lower functional Vin than a fully loaded chip...

DatanoiseTV: > Another possible option would be a `Voltage` menu with the items `Default` and `Auto` (uses the logic here), `0.85V (<=120MHz)`, `0.90V (<=150MHz)`, ... so that the user would be able to manually select one but have some guidance.

I would prefer a method where frequencies above 270 MHz automatically set the core voltage and the flash clock divider, otherwise it will not be very user friendly. With my DSP examples for the Datanoise PicoADK, the core is required to run at 402 MHz.

earlephilhower: 1. I still like the idea of a manual `Voltage` menu on the `generic` board with `Default` being the SDK standard and everything else up to the user's prerogative.  (Set it to 0.8 and 300MHZ, watch your board do nothing...)
2. Adding the 2 add'l frequencies is no big deal and I have no issue with.  Leave it at `Overclock` and we're good.
3. For your specific case, though, using that same infrastructure, you could set the Datanoise board's default clock (it's a define you can set in makeboards.py) and the voltage (would be a new define introduced in 1).  Then your board would "just work" OOTB at the proper frequency without affecting anything else.

For selecting dividers, it's not quite so simple because right now they're just `.S` files that `makeboards` has no idea about other than some ASCII strings for their names.  It might be possible to have it parse the names, but really you should be specifying the proper .S file for your 400mhz operation at 1.3V (or whatever you chose).

I suppose it would be theoretically possible to look thru the datasheets of all the supported flash chips and see what the max QSPI frequency is, but in reality that isn't enough since board design comes in to play.  If there's too much capacitance or induced noise affecting setup/hold, the datasheet #s wouldn't be achievable.  So figuring out where to silently adjust is not so easy.  I'd rather punt that to the overclocker....it's always fun playing w/memory timings in the BIOS, right? :laughing:

PR #2279, opened 2024-07-19
Adding PinMax board

deltaford:
Adding PinMax board


tools/makeboards.py
```diff
@@ -480,6 +480,9 @@ def MakeBoardJSON(name, vendor_name, product_name, vid, pid, pwr, boarddefine, f
 MakeBoard("pimoroni_plasma2040", "Pimoroni", "Plasma2040", "0x2e8a", "0x100a", 500, "PIMORONI_PLASMA2040", 2, "boot2_w25q080_2_padded_checksum")
 MakeBoard("pimoroni_tiny2040", "Pimoroni", "Tiny2040", "0x2e8a", "0x100a", 500, "PIMORONI_TINY2040", 2, "boot2_w25q64jv_4_padded_checksum")
 
+# Pintronix
+MakeBoard("PinMax", "Pintronix", "PinMax", "0x2e8a", "0x9101", 250, "RASPBERRY_PI_PICO", 4, "boot2_w25q080_2_padded_checksum")
```
earlephilhower, line 484: Could you please use the naming format and custom define like the others for the variant?  So
````
MakeBoard("pintronix_pinmax", "Pintronix", "PinMax", "0x2e8a", "0x9101", 250, "PINTRONIX_PINMAX", 4, "boot2_w25q080_2_padded_checksum")
````

earlephilhower (changes requested):

tools/makeboards.py
deltaford: Yes, absolutely

earlephilhower:
Just FYI, you're probably doing this already, but you also need to re-run the script and rename the variant directory to match.  The improperly names JSON needs to be removed, too, and the newly generated one added.

deltaford:
Ok, thanks, will redo and resubmit.

PR #2319, opened 2024-08-04
Add checks for Adafruit TinyUSB to USB libs

earlephilhower:
The included USB libraries are not compatible with Adafruit TinyUSB, so add a #error message when they're built with the define set.


theHDM:
how hard / easy is it to get SingleFileDrive working together with adafruit tinyUSB? suppose i want to have access to adafruit's MIDI over USB library (which i know works) but also want a PC to be able to read littleFS files off the RP2040.

Issue #776, opened 2022-08-20
Servo sweep only giving 90 degrees of movement and not 180ish


johnmholmes:
Hi I have been trying the Arduino pic board manager and have found I am only getting about 90 Degrees of movement and not 180 ish. I tried it first on the Cryton maker pi and then bought a original Raspberry Pi Pico  board. Both boards give the same results.

The servo has been tested on an Arduino Nano and works with full sweep movement. I also tried the RP2040_ISR_servo and this gives me full sweep on both Pico boards.

I have also tested if with both versions of the Arduino Ide. There is no compile errors showing do you have any ideas what I may need to change to allow for closer to 180 movement.

earlephilhower:
The defaults in the Servo library here are a little conservative here to avoid damage in the case of over-driving.  The pulse widths individual servos, especially the no-name or clones, need are sometimes variable.

You can set the min and max servo pulse width in the `attach` command.  See https://www.arduino.cc/reference/en/libraries/servo/attach/

You'd do something like `attach(pin, 1000, 3000)` for example, to double the max pulse length.

johnmholmes:
Thank you Just given that a try and used what I had used before for Arduino Nano values   "attach(15, 600, 2400) " and its work spot on.

Thanks for all the good work you have done and for a fast response.

PR #2326, opened 2024-08-08
Expand documentation to clarify pin notation and Servo usage

AIWintermuteAI:
It is not explained anywhere that unlike official Arduino mbed cores, you need to prepend D before number for digital pins.
Another rather large difference is that servo PWM ranges are set very conservatively according to https://github.com/earlephilhower/arduino-pico/issues/776.

I think this can be two common gotchas for people trying to port their code from Arduino mbed nano core, so perhaps worth adding them to the docs.


docs/digital.rst
```diff
@@ -6,6 +6,10 @@ Board-Specific Pins
 The Raspberry Pi Pico RP2040 chip supports up to 30 digital I/O pins,
 however not all boards provide access to all pins.
 
+Pin Notation
+-----------
+Keep in mind that for using Digital Pins, you need to specify the number of the pin with letter D prepended to it, e.g. D4.
```
earlephilhower, line 11: This is not really the case in general.  All calls in the core work at a GPIO # level.  On the official Pico boards and many others, the PCB pins are labeled directly as the GPIO pins from the chip they're attached to.  For some boards where they didn't follow this, yes you do either need to use the Dx they silk screen (assuming they actually made a full variant file w/this mapping...OTW it'll still be a 1:1 mapping) or the real GPIO it's connected to.

Maybe something like
> When using analog or digital I/Os, if you supply an integer it specifies the RP2040 GPIO pin to use.  Using `Dx` or `Ax` notation (like `D4` or `A3`) may be necessary on boards without a direct PCB pin to GPIO mapping.

docs/digital.rst
```diff
@@ -6,6 +6,10 @@ Board-Specific Pins
 The Raspberry Pi Pico RP2040 chip supports up to 30 digital I/O pins,
 however not all boards provide access to all pins.
 
+Pin Notation
+-----------
```
earlephilhower, line 10: The doc generator complains if the `-`s don't match the size of the name, could you extend these accordingly?

docs/servo.rst
```diff
@@ -11,3 +11,10 @@ require the use of a PIO machine.
 See the Arduino standard
 `Servo documentation <https://www.arduino.cc/reference/en/libraries/servo/>`_
 for detailed usage instructions.  There is also an included ``sweep`` example.
+
+Pulse Width Defaults
+-----------
```
earlephilhower, line 16: Same here re: # of -s

earlephilhower (changes requested):

AIWintermuteAI:
Ooops. Sorry, got too busy with my main work recently. I can still address the comments and get the PR ready, if you want.

earlephilhower:
Sure, if you're up for it!

docs/digital.rst
AIWintermuteAI: Thank you for the explanation. I used your wording with minor capitalization changes.

docs/digital.rst
AIWintermuteAI: Done!

docs/servo.rst
AIWintermuteAI: Done!

earlephilhower (approved):
Thanks for the clean up!

PR #2336, opened 2024-08-17
WIP - Migrate to 2.0.0 SDK

earlephilhower:
There was a lot of breakage with the 2.0.0 SDK, with some bits being just plain diabolical like never initializing global clocks or assigning interrupt vectors.  Nothing wrong w/the SDK, just the way it's being used here is as a libraryvs. recompiling from scratch with a CMake project.

Working through the issues on the RP2040, to be combined with work on the Pico 2/RP2350 once a PR or board is available.  Basic RP2040 blink and LittleFS SpeedTest running, but this is far from release worthy even on the original Pico.


earlephilhower:
WiFi on the PicoW works for basic sanity tests from the examples.  Able to connect to telnet and TLS servers with `WiFiClient` and `BearSSL_Validation` examples.

PR #2342, opened 2024-08-20
Adds ilabs board and PSRAM support.

PontusO:
Very light testing done but the board comes up and prints over USB now. PSRAM also works, currently only with 8MB and only up to 150MHz CPU clock speed.


earlephilhower:
Now that this is clean in CI, I'll try and pull this and SparkFun into something coherent this weekend.  We should be able to take the PSRAM linkage and add a `__psram_start` symbol that the allocator can use (`_psram_end` will be calculated per the chip readout).  Also would like to compare the clock/timing setup here and there.  Looks like SparkFun has compensated for `F_CLK` and uses the explicit SDK counts vs. hardcoded #s.  If they differ, I'll add in a `#define` in the variant and a `#ifndef` in the clock routine.  The PSRAM setup can also be part of the SDK global runtime inits so it "just works", hopefully. :)

PontusO:
> Now that this is clean in CI, I'll try and pull this and SparkFun into something coherent this weekend. We should be able to take the PSRAM linkage and add a `__psram_start` symbol that the allocator can use (`_psram_end` will be calculated per the chip readout). Also would like to compare the clock/timing setup here and there. Looks like SparkFun has compensated for `F_CLK` and uses the explicit SDK counts vs. hardcoded #s. If they differ, I'll add in a `#define` in the variant and a `#ifndef` in the clock routine. The PSRAM setup can also be part of the SDK global runtime inits so it "just works", hopefully. :)

Sounds good. I'll try to help out with getting some of the stuff up that I've been working on for a long time now. I have started to do some more broad testing with running the different examples. So far most things work right out of the box but not everything. Most noticeably Adafruit TinyUSB does not build yet. I'll check in on that tomorrow as it is important to us here.
Got some nice WiFi action going with WiFiEspAT so we're definitely moving in the right direction =)

earlephilhower:
Thanks for the update, especially considering the circumstances.  But, isn't PSRAM size calculated at runtime?
````
....
...
    _psram_size = 1024 * 1024; // 1 MiB
    uint8_t size_id = eid >> 5;
    if (eid == 0x26 || size_id == 2) {
        _psram_size *= 8;
    } else if (size_id == 0) {
        _psram_size *= 2;
    } else if (size_id == 1) {
        _psram_size *= 4;
    }
...
````

For the linker we can put in a dummy size of 16MB (or whatever is the max possible) which is how we build the SDK with (for flash size)...

PontusO:
That is very true and this may not be needed at all. What you do get is a link time check that the allocated memory fits in the memory device of the board. I will leave that up to you to decide if it should be in or not 🙂

PontusO:
Adafruit TinyUSB now build and works but it is not picking up the correct vid/pid.

PR #2395, opened 2024-08-31
Add serial inversion for UART and SerialPIO

earlephilhower:
Use real GPIO pad inversion to allow inverted RX, TX, and controls for the hardware UART and software PIO-emulated serial ports.

Adds ``setInvertTX(bool)`` and ``setInvertRX(bool)`` calls to both ports, with ``setInvertControl(bool)`` for the HW UARTS.


Kinqdos:
Tested it with my sbus receiver and it works fine!

earlephilhower:
Thanks for the update!

PR #2404, opened 2024-09-01
FS doc update

jlbirccyn:
FSInfo64 removed and FSInfo updated. File name length limitation of LittleFS changed to 255 characters


earlephilhower (approved):
Thanks!

PR #2407, opened 2024-09-02
Allow reconfiguring individual FreeRTOS config items - fixes #2398

### LINKED ISSUE #2398: Reconfigure FreeRTOS
Hi there!

Would it be possible to use your own `FreeRTOSConfig.h` file?

The one included in the library (`libraries/FreeRTOS/src/FreeRTOSConfig.h`) e.g. doesn't have configUSE_APPLICATION_TASK_TAG set which I use to debug task switches.

I found a way around this by patching the `FreeRTOSConfig.h` included in arduino-pico, but figured there could be a cleaner solution to this problem.
I see 2 possible solutions: 
- allow inclusion of one's own `FreeRTOSConfig.h`
- allow overwriting any of the defines in a global configuration, by wrapping them in `#ifndef`s, like

> #ifndef configUSE_APPLICATION_TASK_TAG
>  #define configUSE_APPLICATION_TASK_TAG    0
> #endif

What's your thought on this?

Yveaux:


libraries/FreeRTOS/src/FreeRTOSConfig.h
```diff
@@ -1,104 +1,231 @@
...
 #define configUSE_MALLOC_FAILED_HOOK   1
+#endif
+#ifndef configCHECK_FOR_STACK_OVERFLOW
 #define configCHECK_FOR_STACK_OVERFLOW 2
+#endif
 
 /* Cortex-M specific definitions. */
+#ifndef configPRIO_BITS
 #undef __NVIC_PRIO_BITS
 #ifdef __NVIC_PRIO_BITS
```
earlephilhower, line 195: Granted I cut-n-paste this file (or more likely it came already cut-n-pasted from upstream) but the sequence

````
#undef X
#ifdef X
````
is just wrong on so many levels. :laughing:

earlephilhower (approved):

PR #2411, opened 2024-09-04
Fix RP2040 FP calls in ROM

### LINKED ISSUE #2410: Problem with RP2040 floating point with version 4.01?
This is so weird that it must be a problem in my setup... but it is happening in my desktop and my notebook. The following sketch should turn on the Pico LED after a single floating point multiplication is done:

```
double X; // declared here so the multiplication is not removed by optimization

void setup() {
  pinMode(LED_BUILTIN, OUTPUT);

  double T = millis();
  X = -1.3 * T;

}

void loop() {
  delay(100);
}

```

It works fine with version 3.95 but fails with version 4.01 (the LED stays unlit). 

Tested with Arduino IDE 2.3.2 under Windows 10, with a Raspberry Pi Pico and default configurations.

In case someone is curious, this problem showed up when I tried to run a Whetstone benchmark with the new version. I isolated the problem to a single multiplication.

earlephilhower:
A typo in the CMakefile resulted in ROM FPU calls not being added in the Pico library.  Fix the file and rebuild libraries.

Fixes #2410


dquadros:
Great work! I spent some time trying to understanding why my program was not working, now it always make sense.

PR #2413, opened 2024-09-04
Add Datanoise PicoADK v2 (RP2350).

DatanoiseTV:


DatanoiseTV:
@earlephilhower How can I add variants of the v2 board? E.g. there is a version without PSRAM and a version with 8MB PSRAM.

earlephilhower:
If it's just PSRAM then I might special-case `if boards==adk2: BuildPSRAM()`.  Then there's the board and you can pick from a sub-menu the PSRAM included.

DatanoiseTV:
> If it's just PSRAM then I might special-case `if boards==adk2: BuildPSRAM()`. Then there's the board and you can pick from a sub-menu the PSRAM included.

Is BuildPSRAM() something you need to implement or is there something I could do? The V2 is coming out in the next couple of weeks. Let me know if you want one!

earlephilhower:
https://github.com/earlephilhower/arduino-pico/blob/902f709f6f343719a09f5acb46c11f19dd4142d8/tools/makeboards.py#L51

DatanoiseTV:
@earlephilhower Ok, done!

DatanoiseTV:
The PSRAM_CS is fixed at GPIO0 when the PSRAM is used. What is the best place to put it?
Also, please check the BuildPSRAM() if it is at the correct spot.

earlephilhower:
The CS define is normally set in the variant file.

earlephilhower:
Also, I think the psram function should be called in the next if block , [not where you've got it now](https://github.com/earlephilhower/arduino-pico/blob/e0e856eadb7253c60cdd984fcdc87eb6ad3c33d4/tools/makeboards.py#L305).  The current code will miss the flash filesystem size completely, I think.

DatanoiseTV:
> Also, I think the psram function should be called in the next if block , [not where you've got it now](https://github.com/earlephilhower/arduino-pico/blob/e0e856eadb7253c60cdd984fcdc87eb6ad3c33d4/tools/makeboards.py#L305). The current code will miss the flash filesystem size completely, I think.

Done.

tools/makeboards.py
```diff
@@ -298,7 +298,10 @@ def MakeBoard(name, chip, vendor_name, product_name, vid, pid, pwr, boarddefine,
         BuildFlashMenu(name, chip, 8*1024*1024, [0, 7*1024*1024, 4*1024*1024, 2*1024*1024])
         BuildFlashMenu(name, chip, 16*1024*1024, [0, 15*1024*1024, 14*1024*1024, 12*1024*1024, 8*1024*1024, 4*1024*1024, 2*1024*1024])
     elif name == "datanoisetv_picoadk_v2":
-        BuildPSRAM(name)
+            BuildPSRAM(name)
+            BuildPSRAMCS(name)
```
earlephilhower, line 302: `BukldPSRAMCS`, `BuildPSRAMFREQ`, and `BuildFreq` aren't needed here.  You have a hard-wired board and know the CS pin (which should be a `#define PSRAM_CS xxx` in `variants/datanoisetv_v2/pins_arduino.h`).  There's no reason to have these as a menu.  Only the first PSRAM size via `buildpsram` is needed here.  See the `challenger_2350_bconnect` or other RP2350 board w/PSRAM for more info...

earlephilhower (changes requested):

tools/makeboards.py
DatanoiseTV: So I should just leave BuildPSRAM? Could you check the latest commit and maybe make necessary changes if possible? I am currently working on some example code.

DatanoiseTV:
I have implemented the changes. Could you check?

earlephilhower:
The boards file is still not matching the generated output of makeboards, so something is still not quite up to date in the commits to the PR...

tools/makeboards.py
earlephilhower: If your board always has PSRAM then there's no need for these menus.  Define the proper pin/clock/CS in the variant.h like in the other 2350+PSRAM boards and it will "just work."

If there are 2 boards, 1 with and 1 without, then only the BuildPSRAM menu makes sense to have because everything else if defined by the PCB.

`BuildFreq` is called 3 lines below, so the instance here is duplicate  That's not legal and I'm not sure what the IDE will do in that case.  Best case you end up with some weird dup's menu items.

DatanoiseTV:
> The boards file is still not matching the generated output of makeboards, so something is still not quite up to date in the commits to the PR...

I aplologise. I had some lack of sleep recently. Just pulled from master, merged and reworked it.

boards.txt
```diff
@@ -10390,6 +10390,279 @@ datanoisetv_picoadk.menu.uploadmethod.picoprobe_cmsis_dap.upload.maximum_data_si
...
+datanoisetv_picoadk_v2.menu.flash.4194304_2097152.build.fs_start=270524416
+datanoisetv_picoadk_v2.menu.flash.4194304_2097152.build.fs_end=272621568
+datanoisetv_picoadk_v2.menu.flash.4194304_3145728=4MB (Sketch: 1MB, FS: 3MB)
+datanoisetv_picoadk_v2.menu.flash.4194304_3145728.upload.maximum_size=1040384
+datanoisetv_picoadk_v2.menu.flash.4194304_3145728.build.flash_total=4194304
+datanoisetv_picoadk_v2.menu.flash.4194304_3145728.build.flash_length=1040384
+datanoisetv_picoadk_v2.menu.flash.4194304_3145728.build.eeprom_start=272621568
+datanoisetv_picoadk_v2.menu.flash.4194304_3145728.build.fs_start=269475840
+datanoisetv_picoadk_v2.menu.flash.4194304_3145728.build.fs_end=272621568
+datanoisetv_picoadk_v2.menu.freq.150=150 MHz
```
earlephilhower, line 10507: This still has the frequency menu items twice.  Please verify the Python logic you've implemented in the change.  Otherwise, much cleaner and ready to go!

earlephilhower (changes requested):

tools/makeboards.py
```diff
@@ -326,6 +326,8 @@ def MakeBoard(name, chip, vendor_name, product_name, vid, pid, pwr, boarddefine,
             BuildPSRAMCS(name)
             BuildPSRAM(name)
             BuildPSRAMFreq(name)
+        elif name == "datanoisetv_picoadk_v2":
+            BuildFreq(name, 400)
```
earlephilhower, line 330: Sorry, but again you still have 2 frequency menus being generated.  Line 323 and line 330.

earlephilhower (changes requested):

earlephilhower (approved):
Now it looks OK.  Are we ready for merge on this, or are you still working on things WRT VID/etc.?

DatanoiseTV:
> Now it looks OK. Are we ready for merge on this, or are you still working on things WRT VID/etc.?
Ready to merge.

PR #2445, opened 2024-09-12
add "#define PICO_RP2350B 1" since this board is based on RP2350B

SamHalvoe:
I think adding "#define PICO_RP2350B 1" is correct, but I'm not sure if the analog pins have to start at 40 and if I have to change other things besides adding "#define PICO_RP2350B 1". Have a look at [ppico_plus_2_pinout_diagram.pdf](https://github.com/user-attachments/files/16983840/ppico_plus_2_pinout_diagram.pdf).


earlephilhower:
Ah, did not know the existing board was B based.

> ...if the analog pins have to start at 40...

Yes, they do need to be at 40+.  On the -B variant the 8 inputs to the ADC come from GPIOs 40...47.  It's hardwired that way and trying to `analogRead(28)` (for ex) on the -B will not work.

earlephilhower:
Yeah, looking at that diagram seems like it's not gonna work at all for ADC.  Per the datasheed, the 2350B ADCs only go to GPIOs 40+ inside the package:

image

Maybe there's a typo in the pretty-form image?  Is there a schematic you can check?

earlephilhower:
Yup, the pretty colored image is wrong.  Schematic shows the A0...3 pins on the right spor (40+):
image

https://cdn.shopify.com/s/files/1/0174/1800/files/Pimoroni_Pico_Plus_2_Schematic.pdf?v=1724926880

SamHalvoe:
Okay, but whats about the GP on the right? I'm not so good at reading schematics.

SamHalvoe:
What I mean is, they label them both with e.g. GP26 and ADC 40.
grafik

earlephilhower:
Something like
`D26 = 40`
`A0 = 40`
`D27 = 41`
`A1 = 41`
...

SamHalvoe:
Okay. Should I define A3 as ```A3 = 43```?

SamHalvoe:
Okay, so which would be best? There is no fourth analog pin on this board.

SamHalvoe:
In my first pull request I had to define A3, because otherwise the build failed.

earlephilhower:
Yeah, the problem is whoever wrote the original file didn't use the defaults in `generic` for pin statics and instead rolled their own.  The generic one will set sane defaults for everything depending on the CPU type.

Are you sure you need an A3 to build?  The 4.0.1 core still expected `A3` to exist as part of the RP2040, but that logic was replaced with a "there are 4 analog pins starting @ 28(?) on the RP2040 and RP2350A, and 8 analog pins starting @ 40 on the RP2350B).

Worst case, define `A3 = 43` and be done with it.  I don't see and refs in the core to that pin and if a user tries to read a pin that doesn't exist...we can safely ignore that!

SamHalvoe:
I have to admit I wrote the original file ;)
I could use the defaults of generic, if that would be better.

earlephilhower:
No worries.  At this point it's a mismash in `variants` anyway, and this is plumbing nobody should see.  Just adjusting the A pins and setting the flag should get you to a point where users can access all IOs and do whatever they need...

SamHalvoe:
Okay, I commited my changes. Thank you for your time (:

earlephilhower (approved):
Thx!

Andy2No:
Apologies if I'm looking at this wrong, but that seems to say

static const uint8_t A0 = (26u);
static const uint8_t A1 = (27u);
static const uint8_t A2 = (28u);
static const uint8_t A3 = (22u);

and the schematic linked to above shows the ADC0-2 pins as GP40-42 with each one wired to the corresponding one of GP26-28 via a 1k resistor, with GPIO43/ADC3 wired to VSYS_SENSE, with no resistor.

The way I read that is that you have to have the four ADCs defined as GP40-43, and to use ADC0-2 you'd need to disable the corresponding one of GP26-GP28 so they're not fighting over what the header pin does via the resistor that connects them.

Andy2No:
I see.  I was looking at the wrong file for the changed pins, and it now says

static const uint8_t A0 = (40u);
static const uint8_t A1 = (41u);
static const uint8_t A2 = (42u);
static const uint8_t A3 = (43u);

So I guess the trick is just to not use pinMode() on the GP26-GP28 pin corresponding to an ADC that's used.

PR #2446, opened 2024-09-12
solderparty_rp2350_stamp_xl: Properly mark as using RP2350B

arturo182:
I saw the define was needed for the PGA2350, so I think it needs to be here as well.


earlephilhower:
D'oh!  I actually use the XL as the `libpico` SDK board since it forces the inclusion of the -B logic (which is commented out on the -A).

Thx!

arturo182:
Yes, I noticed that and then went to double check if it was actually declaring itself as a -B, but it is now :)

Issue #2428, opened 2024-09-07
Feature request: Include FreeRTOS for RP2350 (from raspberry pi implementation?)


fietser28:
Hello,

Thank you for creating Arduino-pico, I'm a very happy user using rp2040! I'm also impressed how far the RP2350 support already is. Unfortunately for me the main project I'm creating depends on FreeRTOS.

As you stated there is currently no support for FreeRTOS. I've noticed however that raspberry pi have created a FreeRTOS clone with RP2350 support (for ARM M33 cores and based on existing implementation).
The repository is: [https://github.com/raspberrypi/FreeRTOS-Kernel](https://github.com/raspberrypi/FreeRTOS-Kernel)

Would this help to get FreeRTOS  working for rp2350s in Arduino-pico without to much effort?

earlephilhower:
#2406 has been there for a while.  There are subtle issues with the 2350 implementation vs. the 2040 one that are causing weird mutex-related behaviour in stress test apps.  If you have hacked the FreeRTOS internals, please give it a look and see if you can discover why the `tests/freertos_syscallmutex` fails to make progress in certain tasks.

fietser28:
I scanned the issues but not the branches and pull request.

I did some digging and maybe found something: Adding a delay in the start of the threads solves the problem.

You can find a working test at: [https://github.com/fietser28/arduino-pico-freertos-test](https://github.com/fietser28/arduino-pico-freertos-test)

earlephilhower:
Cool, thanks!  Doesn't pinpoint anything but now there's a hack I can examine to see what goes differently inside the kernel.  Should be no need to sleep before the `while(1)` especially since there's a  `delay` in every loop...

fietser28:
I did some tests to try to help pinpoint and can share the following:

The main test case is after 10 seconds of running: from loop a call to vTaskCoreAffinitySet is done to change the affinity of loop2. This is similar to the original behaviour where after a createTask a vTaskCoreAffinitySet is done but the thread is already running then (that's the way Arduino-pico works) and I avoided that scenario with the delays at the start of the threads. 
Using this code:
1. I've copied the content from the *.ino to compile on plain Pico-SDK + FreeRTOS. There the sketch doesn't crash, it works.
[...]

PR #2450, opened 2024-09-13
Added mutex around malloc/free/... (not for PSRAM yet)

fietser28:
See discusion in #2428 for more background. I did some testing and there was no need to use the FreeRTOS heap management functions. Somehow it seems the wrapping of malloc in malloc-lock.cpp undoes the Pico-SDK wrapping (where there is a pico sdk mutex around the malloc function). I've added similar mutex to malloc-lock.cpp. This seems to work (can't generate a crash/lock anymore). 

Note I didn't put the mutexes around the PSRAM calls yet. Can't test this, have no PICO with PSRAM yet.

### DIFF
--- a/cores/rp2040/malloc-lock.cpp
+++ b/cores/rp2040/malloc-lock.cpp
@@ -21,6 +21,7 @@
 #include <Arduino.h>
 #include <malloc.h>
 #include <reent.h>
+#include "pico/mutex.h"
 #include "psram.h"
 
 extern "C" void *__real_malloc(size_t size);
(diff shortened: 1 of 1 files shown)


earlephilhower:
Err, I believe this s unnecessary and duplicated.  Newlib `malloc` already has locking protection implemented using either SDK or FreeRTOS mutexes at build time.  Non-FreeRTOS multicore tests on the 2350/2040 and FreeRTOS tests on the 2040 run properly already.

earlephilhower:
SDK mutexes like `auto_init_mutex(malloc_lock_mutex);` do NOT function properly in FreeRTOS.  They only have a "core ID" as the owner...for FreeRTOS the owner needs to be "task ID" (and hence the wrappers in [lock.cpp](https://github.com/earlephilhower/arduino-pico/blob/master/cores/rp2040/lock.cpp).

fietser28:
You are right. This is not a solution.

Issue #2454, opened 2024-09-14
I2C issue: Ceva BNO085 IMU is not correctly detected


Rasmusfk:
When running the i2c_scanner.ino program the BNO085 is not correctly detected, the address it is found at changes from time to time, and is not the correct I2C address (0x4B). 
The SparkFun_BNO08x_Arduino_Library library does not work either.
4 other I2C devices I have works fine.

Making the following change in libraries/Wire/src/Wire.cpp in the TwoWire::endTransmission() function fixes the problem:
```
-        return _probe(_addr, _sda, _scl, _clkHz) ? 0 : 2;
+        return i2c_read_timeout_us(_i2c, _addr, _buff, 1, false, 1000) < 0 ? 2 : 0;
```

The other 4 I2C devices I have also still works after this change.

I will post pull request.

Regards
Rasmus

PR #2455, opened 2024-09-14
Fixes issue #2454: I2C issue: Ceva BNO085 IMU is not correctly detected

Rasmusfk:
This fixes issue #2454, and has been tested with 4 other I2C IC's which still works after the fix.

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -296,63 +296,6 @@ static bool _clockStretch(pin_size_t pin) {
     return digitalRead(pin);
 }
 
-bool _probe(int addr, pin_size_t sda, pin_size_t scl, int freq) {
-    int delay = (1000000 / freq) / 2;
-    bool ack = false;
-
-    pinMode(sda, INPUT_PULLUP);
-    pinMode(scl, INPUT_PULLUP);
-    gpio_set_function(scl, GPIO_FUNC_SIO);
-    gpio_set_function(sda, GPIO_FUNC_SIO);
-
-    digitalWrite(sda, HIGH);
-    sleep_us(delay);
-    digitalWrite(scl, HIGH);
-    if (!_clockStretch(scl)) {
-        goto stop;
-    }
-    digitalWrite(sda, LOW);
-    sleep_us(delay);
-    digitalWrite(scl, LOW);
-    sleep_us(delay);
-    for (int i = 0; i < 8; i++) {
-        addr <<= 1;
-        digitalWrite(sda, (addr & (1 << 7)) ? HIGH : LOW);
-        sleep_us(delay);
-        digitalWrite(scl, HIGH);
-        sleep_us(delay);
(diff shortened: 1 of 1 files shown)


Rasmusfk:
Oh, and by the way this also removes the now unused _probe() function.

earlephilhower (changes requested):
Thanks, but this isn't going to work in general  Translating a 0-byte write into a 1-byte read was discussed and dismissed when the core came out.  In your particular case it may fix things, but in general there were significant issues seen.

The Adafruit folks chimed in on the existing implementation way back when, https://github.com/earlephilhower/arduino-pico/issues/38 (and other issues/prs later).  IIRC, it boiled down to 0-byte writes are guaranteed no-ops on the device, while reads to invalid registers could either (falsely) fail or change chip state causing an error later on..

If detection is iffy with the current code on your setup (and FWIW I know several drone folks using the core so they might be using the same chip if not PCB) then it's possible there's a pull-up/down strength or timing issue.

It's also possible the existing bit banging core is sampling on the wrong edge or not obeying setup/hold timing...if you see a timing or edge issue in the bitbang code, I'm happy to get PRs to adjust that.  But we can't replace a write with a read here.

dquadros:
> Thanks, but this isn't going to work in general Translating a 0-byte write into a 1-byte read was discussed and dismissed when the core came out. In your particular case it may fix things, but in general there were significant issues seen.
 

Not knowing the history of this feature, I was going to ask exactly this.

Rasmusfk:
Ok, thanks! I just assumed that as the Pico SDK used the read-1 method, then it would be ok in general:
https://www.raspberrypi.com/documentation/pico-sdk/hardware.html#group_hardware_i2c
I'll look more into the timing, and see if I can identify the issue.

Rasmusfk:
So this turned out to be a bit of a rabbit hole. The I2C spec. specifies clock stretching, which the BNO085 uses, but it doesn't specify any upper limit for how long a target device can stretch the clock:
https://learn.adafruit.com/working-with-i2c-devices/clock-stretching

The BNO085 is correctly detected, when increasing the clock stretch timeout from 100us to 500us like this in Wire.cpp:

```
static bool _clockStretch(pin_size_t pin) {
    auto end = time_us_64() + 500; // <-- increased from 100 to 500

    while ((time_us_64() < end) && (!digitalRead(pin))) { /* noop */ }
    return digitalRead(pin);
}
```

There is not anything magic about 500us other that it is the shortest timeout where the BNO085 is still reliably detected. Other I2C devices might need (much?) longer timeouts.

In particular for the BNO085 the increased timeout is only needed when reading the ACK/NACK:


```
...
 if (!_clockStretch(scl)) {
      goto stop;
  }

  ack = digitalRead(sda) == LOW; 
```

Would you accept a pull request where the timeout in _clockStretch() is increased to i.e. 5000 (to be on the safe side)?
The change would have minimal impact on how long ie the i2c_scanner example takes to run.

PR #2458, opened 2024-09-16
I2C lib: Handle timeout in write()

tommag:
This PR handles the timeout in I2C write() function. This avoids hanging in peripheral mode, when replying to a request with more bytes than the controller expects. 

Let me know if any changes are needed ! And thanks for the hard work on this integration :)


earlephilhower (approved):
LGTM, thanks!

PR #2459, opened 2024-09-16
Add Adafruit Feather RP2350 with HSTX

BlitzCityDIY:
Adding board definition for Adafruit Feather RP2350. This includes a delay for XOSC.


earlephilhower:
Looks like there are merge issues, can you take a look at it, @BlitzCityDIY ?  FYI, RP2350B patch needed a new set of libs built with different CMake options.  Your new libs w/HSTX turned on might need to be rebuilt from head options...

earlephilhower (approved):
LGTM.  One thing, though, is there any possible configuration on the HSTX port?  Should there be `#define`s for the pinout in the `variant.h` file, or does it always live at the same IO location w/o GPIO muxing?

tannewt:
The pins don't vary due to the pin mux. The HSTX can vary what comes out each HSTX pin on its own. You could swap color channels for example.

earlephilhower:
Got it, thx.  Just noted the same from the RP2350 datasheet.

There's one old file, I think, causing CI failure now.  The JSON file should be regenerated and not include `pico-debug`.  See the style CI diff (note that pico-debug is no longer maintained or in OpenOCD):

````
diff --git a/tools/json/adafruit_feather_rp2350_hstx.json b/tools/json/adafruit_feather_rp2350_hstx.json
index 05a783a..4ed8480 100644
--- a/tools/json/adafruit_feather_rp2350_hstx.json
+++ b/tools/json/adafruit_feather_rp2350_hstx.json
@@ -47,8 +47,7 @@
             "jlink",
             "raspberrypi-swd",
             "picotool",
-            "picoprobe",
-            "pico-debug"
+            "picoprobe"
         ]
     },
     "url": "https://www.raspberrypi.org/products/raspberry-pi-pico/",
Error: Process completed with exit code 1.
````

tannewt:
Thank you for the fast reviews!

PR #2468, opened 2024-09-17
Adds support for ESP Hosted to wifi/ble boards and adds a menu item for ESP WiFi selection

PontusO:
This PR adds direct support for ESP Hosted for iLabs wifi/ble boards. It also adds a meny item that allows you to select between the two depending on what is installed on the esp32 device.


PontusO:
It seems there is a local ESPHost in the CI environment that overrides the new submodule, and there is an issue with the library itself that needs attending.

earlephilhower:
You can delete the GH action line here
https://github.com/earlephilhower/arduino-pico/blob/ccdc76c279b887249700a3e37d56e14f8ba2d0de/tests/common.sh#L190
to avoid the ESPHost checkout (added to ensure the `lwip_ESPHost` stuff actually keeps building).

earlephilhower:
You can also add the single failing ESPHost-library ino to the exclude list https://github.com/earlephilhower/arduino-pico/blob/ccdc76c279b887249700a3e37d56e14f8ba2d0de/tests/common.sh#L28 until the upstream repo gets it cleared up.

tools/makeboards.py
```diff
@@ -274,9 +274,16 @@ def BuildGlobalMenuList():
     print("menu.boot2=Boot Stage 2")
     print("menu.wificountry=WiFi Region")
     print("menu.usbstack=USB Stack")
+    print("menu.espwifitype=ESP wifi type")
```
earlephilhower, line 277: Minor thing, but could it be `ESP WiFi Type` (capitalization diff) to match the existing menu formats?

PR #2469, opened 2024-09-17
Remove CFG_TUSB_DEBUG for rp2350

Yveaux:
This change removes the default define of setting CFG_TUSB_DEBUG to 0 for rp2350 targets and allows a user to change the TUSB debug level at compile time.

This default is not present for rp2040 targets (https://github.com/earlephilhower/arduino-pico/blob/163b209d6c60733f110b02b9cb5f0b3c48970955/lib/rp2040/platform_def.txt)

And it defaults to 0 anyway: https://github.com/earlephilhower/arduino-pico/blob/163b209d6c60733f110b02b9cb5f0b3c48970955/include/rp2350/tusb_config.h#L46)


earlephilhower (approved):
Seems fine.  I just cut-n-pasted the entire `Cmake` define list for the rp2350 and I get many others are really defaults as well.

PR #2479, opened 2024-09-19
I2C lib : implement "bus recovery" feature on timeout reset

tommag:
This feature is described in RP2040 datasheet chapter 4.3.13.1 but not implemented in hardware (see https://github.com/raspberrypi/pico-sdk/issues/1595) 
This commit implements it in software, based on code from the probe() function.

My I2C bus is much more robust with this addition :)


earlephilhower (approved):
LGTM, thanks for the report and PR!

PR #2487, opened 2024-09-22
Make USB manufacturer and product name customizable

misaka19465:
Some devices may pick ready to use rp2040 boards, but need an custom product name, to make configuring easier for example. This patch added customizable manufacturer and product name through `build_flags` (this is my present solution. better solution could be found in the future).

### DIFF
--- a/cores/rp2040/RP2040USB.cpp
+++ b/cores/rp2040/RP2040USB.cpp
@@ -324,8 +324,16 @@ const uint16_t *tud_descriptor_string_cb(uint8_t index, uint16_t langid) {
 
     static const char *const usbd_desc_str[] = {
         [USBD_STR_0] = "",
+#ifdef CUSTOM_USB_MANUFACTURER
+        [USBD_STR_MANUF] = CUSTOM_USB_MANUFACTURER,
+#else
         [USBD_STR_MANUF] = USB_MANUFACTURER,
+#endif
+#ifdef CUSTOM_USB_PRODUCT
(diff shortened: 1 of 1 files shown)


earlephilhower (changes requested):
These values already are defined by the user during the build process on the command line or the P.IO boards.json file.  Make a custom board file and you can define them any way you like, without needing to add anything.

maxgerhardt:
It already is customizible.

https://github.com/earlephilhower/arduino-pico/blob/a6ab6e1f95bc1428d667d55ea7173c0744acc03c/tools/platformio-build.py#L287-L297

Allows you to set this in the `platformio.ini`, e.g.
```ini
[env:pico]
platform = [raspberrypi](https://github.com/maxgerhardt/platform-raspberrypi.git)
board = rpipico
framework = arduino
; Set custom manufacturer and product
board_build.arduino.earlephilhower.usb_manufacturer = Custom Manufacturer
board_build.arduino.earlephilhower.usb_product = Ultra Cool Product
```
grafik

We need a documentation update here for PlatformIO, not a code change.

maxgerhardt:
Well, mostly.. Found out that for changing USB VID and PID, the PlatformIO build defines the macros `USB_PID`, `USB_VID`, but these only seem to be used in Adafruit TinyUSB.. The regular USB stack uses `USBD_PID`, `USBD_VID`. Also defining those makes the VID+PID freely redifinable, too.

grafik

I will file a PR shortly for docs + builder script update.

misaka19465:
Thanks for your works. I'll close this.

PR #2491, opened 2024-09-23
Add RISC-V support

earlephilhower:
This PR's pretty good to go. I'm still not sure it'll be part of the mainstream, but other than the GCC update and a fix to locking, no ARM-related changes were done so it shouldn't actually have any negative impact on the main platform.

Platform.IO runs will fail until it starts pointing to the new GCC 14.2 toolchain. @maxgerhardt any suggestions? Any way of using different tools programmatically?

RISC-V/ARM can be selected through the menus and most things "just work" under RISC-V. Known issues are OTA and some FreeRTOS-related USB stuff not fully implemented.

A new toolchain, GCC 14.2, is required since RISC-V support in 12.4 is not usable, so git clones need to do a get.py to install. This also will increase tool installation size since both ARM and RISCV compiler chains are needed.


earlephilhower:
If you pull this PR and run `get.py` on Linux, then select the Pico2-RiscV and then Tools->C++ Exceptions->Enabled you can build a UF2 (for blink.ino at least).  [The exceptions disabled special build in the new GCC14.2 toolchain seems to have had a hiccup and can't link yet.]

However, the binary seems to be DOA,.  I'm not having any luck getting OpenOCD running for R5 (not even on the SDK stuff), so if anyone has any good documents on getting R5 debugging running it's be appreciated!

earlephilhower:
The core can now build and run RISC-V code (only).  Tested PWMAudio, Blink, and Fade, and all run on quick inspection.  Probably many different peripheral libraries not there yet, but basic sanity is fine.

To try it out, make sure
* Select Tools->Board->Raspberry Pi Pico 2 Risc-V
* Select Tools->C++ Exceptions->Enabled

GDB and OpenOCD work, but I think (for all R5 debug on the RP2350) you need to already have a running RISCV app, connect w/OpenOCD, and then load your real test app via GDB/etc.

earlephilhower:
P.IO builds here aren't going to work for the RP2350-RISCV until the upstream platform.io builder is updated to select the right compiler.  But the IDE and arduino-cli should work for folks now.

maxgerhardt:
I'm working on the PIO integration right now.

Some things stand out:
* the board JSON files for RP2350 boards have their boot2.S files all replaced with `none.S`? Then the information what boot2 file to use for the regular ARM build got lost?
* the RISC-V toolchain packages from https://github.com/earlephilhower/pico-quick-toolchain/releases, e.g. [aarch64-linux-gnu.riscv32-unknown-elf-8b9df28.240926.tar.gz](https://github.com/earlephilhower/pico-quick-toolchain/releases/download/4.0.0/aarch64-linux-gnu.riscv32-unknown-elf-8b9df28.240926.tar.gz) still has the same `toolchain-rp2040-earlephilhower` name. Can we rename that to e.g. `toolchain-rp2040-earlephilhower-riscv`? I'm afraid PlatformIO will redownload / re-unpack the whole toolchain package when a user switches between an ARM and a RISC-V build. We could have both installed in parallel under different package names.

maxgerhardt:
The toolchain URLs and toolchain switch was implemented (see [commits/develop/](https://github.com/maxgerhardt/platform-raspberrypi/commits/develop/)). The examples all build now again, for RP2040, RP2350 ARM + RISC-V. Users can switch to the RISC-V build with

```
; RP2350 based (ARM)
[env:rpipico2]
board = rpipico2

; RP2350 based (RISC-V)
[env:rpipico2-riscv]
board = rpipico2
board_build.mcu = rp2350-riscv
```

It's still weird because the RISC-V build still builds the regular boot2.S file *and it works*, LED blinking and all. (I have not yet synced the PIO board JSON changes). How does that work here? Does the RISC-V build not need a boot2 file to integrate with the external flash chip?

earlephilhower:
Thanks!

> the board JSON files for RP2350 boards have their boot2.S files all replaced with none.S? Then the information what boot2 file to use for the regular ARM build got lost?
> ...
> Does the RISC-V build not need a boot2 file to integrate with the external flash chip?

The boot2 isn't actually used anywhere on the RP2350.  On the 2040 ROM they called a specific address where it lived and it was responsible for setting up the QSPI interface.  On the 2350 the ROM has add'l intelligence which they say get you running "good enough" with whatever flash.  If the app wants to adjust the QSPI timing/etc. then it can do so itself programmatically.  We don't.  The boot2 files were just being added to the binary image but never accessed.

>Also, why do we need special board definitions for RP2350 boards when in RISC-V mode? (variants/rpipico2rv/pins_arduino.h)

A leftover, dropped from the latest pushes.  It was a board before, now it's just a build option...

> the RISC-V toolchain packages... Can we rename that to e.g. toolchain-rp2040-earlephilhower-riscv...

Will do next build.  Also going to try and reduce the R5 GCC image size because it now includes every R5 architecture option...it's a 3hr/32 parallel job monster now, though, so might be next week when the new files get up there.

maxgerhardt:
Can you clear the cache for PlatformIO in the Githu actions once per <https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#viewing-cache-entries>? It's still using the old platform version and with it the old toolchain, that's why compiling fails.

earlephilhower:
@maxgerhardt , the toolchain 4.0.1 was rebuilt and uploaded with the fixes JSON entries and smaller d/l and install sizes.  When you update the P.IO wrapper let me know and I'll blow away the 4.0.0 one.

znmeb:
How does a PlatformIO end user with an RP2350 board go about designating that the compile / upload should use RISC-V rather than Arm Cortex M33?

earlephilhower:
https://github.com/earlephilhower/arduino-pico/blob/master/docs/platformio.rst#selecting-the-cpu-architecture

maxgerhardt:
Full `platformio.ini` example with some chippies:
```ini
; common settings
[env]
framework = arduino

; RP2040 based
[env:pico]
board = rpipico

[env:picow]
board = rpipicow

; RP2350 based (ARM)
[env:rpipico2]
board = rpipico2

; RP2350 based (RISC-V)
[env:rpipico2-riscv]
board = rpipico2
board_build.mcu = rp2350-riscv
```

PR #2504, opened 2024-09-30
Update PIO platform before compile, add RISC-V PIO build to CI

maxgerhardt:
Should eliminate the need to occassionally delete the cache for PlatformIO.

The cache is still restored at the stard and the state at the end gets cached, so we still retain the speed advantage if nothing changed in the platform.


maxgerhardt:
I can see the update taking place correctly inside the Gitlab runner logs, and the RISC-V build passes. This should be good to merge.

PR #2527, opened 2024-10-06
Document setting the PSRAM size for PlatformIO

maxgerhardt:
Was still missing in the docs.


maxgerhardt:
CI failure is due to a bug in the platform that I already fixed, but the platform hasn't updated here properly. Looking into it.

maxgerhardt:
This should be it.

PR #2533, opened 2024-10-10
Add seeed xiao rp2350 board

Cincinnatu:


earlephilhower (changes requested):
Where is the updated makeboards?  That needs to be checked in.

Cincinnatu:
> Where is the updated makeboards? That needs to be checked in.
This is the link to the Seeed Xiao RP2350 board that I added recently: https://www.seeedstudio.com/Seeed-XIAO-RP2350-p-5944.html. Do you need us to provide any other information?

earlephilhower:
No, what you need to check in is `tools/makeboards.py` which (I hope) you added the new board to.  If you just cut-n-paste or edited `boards.txt` in any way that's not going to work.  You need to update `makeboards.py`, rerun it, and check in the new script and files.  See the [docs for more info](https://github.com/earlephilhower/arduino-pico/blob/master/docs/contrib.rst#adding-a-new-board).

earlephilhower (approved):
Much better, thanks

PR #2539, opened 2024-10-12
Preserve PSRAM QMI interface around flash ops

### LINKED ISSUE #2537: Flash operations break PSRAM interface
I have a program using LittleFS that works OK when not using PSRAM, but hangs when using PSRAM.

Here's a minimal demonstration:
````text
#include <LittleFS.h>

uint32_t Workspace[10000] PSRAM;
const int len = 1000;

bool checkdata () {
  for (int i=0; i<len; i++) if (Workspace[i] != i) return false;
  return true;
}

void savedata () {
  for (int i=0; i<len; i++) Workspace[i] = i;
  if (checkdata()) Serial.println("Data setup OK");
  LittleFS.begin();
  File file;
  file = LittleFS.open("/DATA.DAT", "w");
  file.close();
}

void setup() {
  Serial.begin(9600);
}

void loop() {
  savedata();
  Serial.println("Saved");
  if (!checkdata()) Serial.println("Data error");
  delay(5000);
}
````
The program simply opens and closes a file on LittleFS.

Running the program displays **Data error** indicating that LittleFS has corrupted the array **Workspace**. If you remove the **PSRAM** attribute from **Workspace** the program runs without error.

### PR DESCRIPTION
**Pull request by earlephilhower**
The flash ROM routines seem to overwrite the QMI configuration we set for PSRAM, rendering it unreadable after any erase or write or ID command.

Wrap the 4 flash control functions to preserve the QMI state on the RP2350.  On the RP2040, simply pass through the call.

Fixes #2537

### DIFF
--- a/cores/rp2040/flash_wrapper.cpp
+++ b/cores/rp2040/flash_wrapper.cpp
@@ -1,0 +1,73 @@
+/*
+    Flash wrappers to protect PSRAM access on the RP2350
+
+    Copyright (c) 2024 Earle F. Philhower, III <earlephilhower@yahoo.com>
+
+    This library is free software; you can redistribute it and/or
+    modify it under the terms of the GNU Lesser General Public
+    License as published by the Free Software Foundation; either
+    version 2.1 of the License, or (at your option) any later version.
+
+    This library is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by maxgerhardt** [comment]
PR probably failing for RP2040 because
```cpp
#ifdef PICO_RP2350
#include <hardware/structs/qmi.h>
#endif
```
is not hit, but then no `#include` is done, and all those types like `uint32_t` are missing.

PR #2559, opened 2024-10-25
Add ISR check to critical section

ldursw:
`taskENTER_CRITICAL()` and `taskEXIT_CRITICAL()` cannot be used in an interrupt context.

This PR allows the usage of `noInterrupts()` and `interrupts()` inside an interrupt. Libraries that call these functions should now work instead of hanging.


earlephilhower (approved):
Thanks!

PR #2582, opened 2024-11-04
Add SimpleMDNS, IGMP, and .local lookup

earlephilhower:


earlephilhower:
@jaguilar, since you're doing FreeRTOS and MDNS, can you give this PR and the new `SimpleMDNS` library a test?

There were multiple LWIP tweaks needed to actually allow the MDNS responder to work, like increasing the internal LWIP timers.  W/O it you'd `panic` as soon as you actually did some network I/O.

Basic testing replacing `LEAmDMS` in OTA and HTTPUpdateServer seem to "just work."

jaguilar:
Yes, I will test it.

jaguilar:
For anyone testing something like this from platformio in the future, the magic is:

```
[env:pico]
platform_packages = 
		framework-arduinopico @ https://github.com/earlephilhower/arduino-pico.git#1a05d9261df14c5ecabf6c9dc8af28b0ca6def9e
```

And I can confirm it works. I am able to resolve .local addresses after flashing with the rebuilt program. I did not test the MDNS responder because I do not know how it works and I have no code to use it at the moment.

jaguilar:
One other really great thing about this feature -- even though my server has a static IP my client was having some inconsistency reconnecting to it when I was manually loading the IP address. Now that we're over to mdns, I have restarted the server several times and seen no client reconnect failures. So happy!

earlephilhower:
@GUVWAF `begin` can only fail if it's already running or there are no `netif`s available.  Want to do a PR to add the `bool` return to it and make it even better?

GUVWAF:
@earlephilhower Yes, will do that.

PR #2591, opened 2024-11-08
Modify the pin definitions for XIAO RP2350

Cincinnatu:


earlephilhower:
Again, what exactly are you changing here and why rewrite it in this way instead of using the existing infra to redefine pins?

Cincinnatu:
Here is the pinout for the XIAO RP2350. I need to define the pins that will be used so that users can easily use them in Arduino, as well as ensure compatibility with other Arduino library files.
image

earlephilhower:
Ah, I see.  The person who added the board originally didn't actually get the pin mappings at all, it seems.  Thanks for the update, then!

PR #2599, opened 2024-11-11
SimpleMDNS: let `begin` return a bool

GUVWAF:
As discussed in #2582, this makes SimpleMDNS an even better drop-in replacement for LEAmDNS.


earlephilhower (approved):
Thx!

PR #2600, opened 2024-11-11
Fix SPI communication with overclocked Pico W

maxgerhardt:
Just a basis for a possible fix for https://github.com/earlephilhower/arduino-pico/discussions/2590.

Open points:
* instead of a simple div by 4 at a certain frequency, try to compute the (fractional) divider such that we always try to reach a certain SPI clock speed, such as 66.5MHz
* move fix to the lwip_cyw43439 library instead of the Pico W variant, to make it more generic? I wanted the call to be before the first cyw init call though


earlephilhower:
Let's do this as part of #2616.  I copied your changes here in addition to the refactor, so I think we're good.  If not, we can reopen here and figure something out.

PR #2604, opened 2024-11-13
lwIP_Ethernet: rename copied cyw43 mac function

### LINKED ISSUE #2602: esphome fails to compile with 4.2.0 release due to conflicting decls of cyw43_hal_generate_laa_mac
When including both `WiFi.h` and `cyw43.h`, I get this compile error:
```
~/.platformio/packages/framework-arduinopico/libraries/lwIP_Ethernet/src/LwipIntfDev.h:57:17: error: 'void cyw43_hal_generate_laa_mac(int, uint8_t*)' was declared 'extern' and later 'static' [-fpermissive]
```

`cyw43_hal_generate_laa_mac` is defined here as `static`:
https://github.com/earlephilhower/arduino-pico/blob/db4b167aaafc33d235d2b1f1949fcae8d107c10e/libraries/lwIP_Ethernet/src/LwipIntfDev.h#L57

While it is declared here as non-static:
https://github.com/raspberrypi/pico-sdk/blob/efe2103f9b28458a1615ff096054479743ade236/src/rp2_common/pico_cyw43_driver/include/cyw43_configport.h#L125

Removing the static from `LwipIntfDev.h` will fail with multiple definitions during linking, which can be solved by adding `__attribute__((weak))` to the definition in `LwipIntfDev.h`.

afflux:
Prevent symbol conflicts when including both Wifi.h and cyw43.h

Fixes #2602

### DIFF
--- a/libraries/lwIP_Ethernet/src/LwipIntfDev.h
+++ b/libraries/lwIP_Ethernet/src/LwipIntfDev.h
@@ -53,15 +53,13 @@
 
 // Dup'd to avoid CYW43 dependency
 // Generate a mac address if one is not set in otp
-extern "C" {
-    static void cyw43_hal_generate_laa_mac(__unused int idx, uint8_t buf[6]) {
-        pico_unique_board_id_t board_id;
-        pico_get_unique_board_id(&board_id);
-        memcpy(buf, &board_id.id[2], 6);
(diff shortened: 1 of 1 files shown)


earlephilhower (approved):
Thx!

PR #2628, opened 2024-11-21
Add Adafruit Floppsy board

jepler:
The pinout here is for the "B" board revision but the final retail version is likely to be very similar.

ping @ladyada


earlephilhower:
Looks like minor astyle complaints.  If you move the indent to after the `//` it will all line up and keep CI heppy.

jepler:
Thanks, I'll take a look.

jepler:
Thank you!

PR #2653, opened 2024-11-30
Increase LWIP timeouts to avoid MDNS panics

### LINKED ISSUE #2650: SimpleMDNS in combination with WiFiClient leads to panic
Using a Pico W (RP2040), SimpleMDNS in combination with `WiFiClient` leads to a panic due to an assert in LWIP:

`sys_timeout: timeout != NULL, pool MEMP_SYS_TIMEOUT is empty`

Here’s a MCVE:
```cpp
#include <Arduino.h>
#include <WiFi.h>
#include <SimpleMDNS.h>

void setup() {
  Serial.begin(115200);
  delay(5000);
  WiFi.begin("ssid", "psk");
  while (WiFi.status() != WL_CONNECTED) {
    delay(500);
  }
  Serial.println("Starting MDNS service");
  MDNS.begin("test");
  MDNS.addService("http", "tcp", 80);
  
  WiFiClient client;
  Serial.println("Connecting...");
  client.connect("www.google.com", 80);
}

void loop() {
  delay(1000);
  Serial.println("...");
}
```

It happens both for core version 4.2.1 and 4.3.0. When replacing SimpleMDNS with LEAmDNS, this works.

I found a reference to this error here: https://github.com/raspberrypi/pico-sdk/issues/1281. So I bumped `LWIP_DNS` from 1 to 2 in `lwipopts.h` and recompiled `libpico`, and then the MCVE works. However, I'm not sure if this is a suitable fix or we should find out why SimpleMDNS needs more timers than LEAmDNS.

earlephilhower:
Bump up the LWIP_ARP setting to increase the LWIP timeout pool to avoid a panic when using SimpleMDNS and TCP clients.

Fixes #2650


earlephilhower:
@GUVWAF can you give this a try with your real application?  I've added 2 add'l LWIP timer slots and rebuilt everything.  Should be microscopic RAM impact and hopefully no OOM panics...

GUVWAF:
Thanks for the fix and the explanation. 

Just tried this with the application (Meshtastic, which offers an API over TCP and optionally connects to an MQTT server), and it works fine now.

PR #2673, opened 2024-12-05
Added logic so that the build.mcu value is written to a board definition

gigapod:
The value of baord.mcu is needed when building with Arduino libraries that include archive files (.ar).

### DIFF
--- a/tools/makeboards.py
+++ b/tools/makeboards.py
@@ -238,8 +238,9 @@ def BuildHeader(name, chip, chaintuple, chipoptions, vendor_name, product_name, 
     print("%s.build.usbpid=-DUSBD_PID=%s" % (name, main_pid))
     print("%s.build.usbpwr=-DUSBD_MAX_POWER_MA=%s" % (name, pwr))
     print("%s.build.board=%s" % (name, boarddefine))
-#    print("%s.build.mcu=cortex-m0plus" % (name))
+
     if chip == "rp2040":  # RP2350 has menu for this later on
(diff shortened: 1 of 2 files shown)


tools/makeboards.py
```diff
@@ -260,6 +261,8 @@ def BuildHeader(name, chip, chaintuple, chipoptions, vendor_name, product_name,
     print('%s.build.usb_product="%s"' % (name, product_name))
     if ((chip == "rp2350") or (chip == "rp2350-riscv")) and (name != "generic_rp2350"):
         print("%s.build.psram_length=0x%d00000" % (name, psramsize))
+    if chip == "rp2350":
+        print("%s.build.mcu=cortex-m33" % (name))
```
earlephilhower, line 265: It's a minor thing, but we do support RISCV mode for everything here (even WiFi and BT).  Is it possible to set the `build.mcu` as part of the `BuildArch` function for the RP2350 instead of hardcoding it here as an ARM?  The R5 mcu should be `rv32imac` if I understand it correctly...

earlephilhower (changes requested):
Thanks!  Small request for the R5 folks...

tools/makeboards.py
gigapod: Changes made - and verified they work using our rp2350 board in ARM "mode".

earlephilhower: Awesome, thanks!

earlephilhower (approved):

PR #2697, opened 2024-12-17
don't go through runtime initializers when there is no OTA command

afflux:
Leave that to the main app instead, so we don't reset peripherals twice.

See discussion here: https://github.com/earlephilhower/arduino-pico/discussions/2686


afflux:
I've verified that my problem is solved with this change, but I haven't given OTA much testing at this point.

afflux:
OTA works on my Pico W

earlephilhower (changes requested):
You need to rebuild all 3 OTAs and add them to the PR, please.  The IDE doesn't even look at the `/ota` directory or the source files there.  they are included as an opaque blob since they have duplicated symbols from the main app.

afflux:
Done. Note that the OTA build currently tries to copy riscv header files somewhere and fails:

```
cp: cannot create regular file '../../include/rp2350-riscv/pico_base/pico/.': No such file or directory
```

Let me know if you want me to clean that up in this PR as well

afflux:
One thing to note: I'm reordering the runtime_init priorities a little, because reading the LFS to check if there is an OTA available requires working memset/memcmp/memcpy, which is wired to the bootrom functions and therefore requires `__aeabi_mem_init` to have run. I have tested this on rp2040, but I don't know if rp2350{,-riscv} needs other `runtime_init` functions.

https://github.com/raspberrypi/pico-sdk/blob/2.1.0/src/rp2_common/pico_runtime_init/include/pico/runtime_init.h

And now that I think of it, it might make sense to also `__aeabi_{bits,float,double}_init` to run before `check_ota()`, just in case one of those functions end up being called by LittleFS in some code path I haven't seen yet.

earlephilhower:
> `cp: cannot create regular file '../../include/rp2350-riscv/pico_base/pico/.': No such file or directory`
> Let me know if you want me to clean that up in this PR as well

Could you, please?  I have it in my own tree but I forgot to `git add include/rp2350-riscv/pico_base/pico/ota_command.h`

afflux:
There doesn't appear to be anything referencing `include/rp2350-riscv`, instead the riscv build will also use the `include/rp2350` headers, so I would've just disabled the copy for riscv:

https://github.com/earlephilhower/arduino-pico/blob/66af35957852c3b2ba7e9797caa94ef16073d34a/lib/rp2350-riscv/platform_inc.txt#L1-L2

earlephilhower:
The problem is then you have to special case the R5 build.  We still need it copied over in the rp2350 and rp2040 one.  It's doable, but maybe not so clean in cmake?

earlephilhower:
A quick run of OTAFromFile on the Pico2 using both ARM and RISC-V modes seems to still work, so this looks good!

earlephilhower (changes requested):
Looks good, no effect on Pico2 ARM or RISC-V (there's no mem-ops there, ROM doesn't include much at all and is very slow).

Float/double aren't in LittleFS so no worries about including that (softfp would blow up the 12KB we reserve for OTA code so this seems safe) but they could be moved up.

The bit operations **probably** will be used, good point.  I'm not sure the filesystem needs to do CLZs and POPCOUNT, but we have a GZIP decompressor that can run under OTA which in my experience is going to want those special builtins.

earlephilhower (approved):
Thanks, looks good now.  Tested Pico2 and Pico and GZip compressed OTA, no issues seen.

PR #2698, opened 2024-12-17
Rudimentary Pimoroni Plasma 2350 support

doctea:
Hi!  Simple bodge-together of a variants pins_arduino.h file for _pimoroni_plasma2350_ boards.

Not thoroughly tested at all so may be some errors on pin definitions, but is working for me so far.


earlephilhower:
Thanks, but new boards need an updated makeboards.py and JSON file, too.  See the [docs](https://github.com/earlephilhower/arduino-pico/blob/master/docs/contrib.rst#adding-a-new-board) for instructions...

doctea:
@earlephilhower thanks!  I believe I've now made the correct changes and tracked the necessary files.  Please could you let me know if I've still missed something?

Also, importantly for I2C on Plasma RP2**0 users, I've fixed a problem in the pins definition where ___WIRE0_DEVICE_ was defined to use i2c1 instead of i2c0.  Making this change suddenly made i2c start working for me on my Plasma RP2350 project!

I've got a feeling that the same change will be necessary for RP2040 -- I've never tested this, but presume the same issue is likely the same and presume nobody else ever has ever tried this or reported the problem, so I've updated the arduino_pins.h for the RP2040 board here, too!

earlephilhower (approved):
LGTM.  Thanks for the fix on the other board, too.

PR #2704, opened 2024-12-19
Fix issue with undefined LED_BUILTIN on Seeed Xiao RP2350

Cincinnatu:


earlephilhower (approved):
Thanks.  This is the problem with defining the pins in a self contained file, vs. using the `../generic/common.h` header and just setting any override defines.

PR #2708, opened 2024-12-20
Overclock to 153.6 MHz (instead of 147.6 MHz) for I²S 48 kHz sample rate

magy00:
Increasing `clk_sys` by another 6 MHz would allow for exact 96 kHz and 192 kHz sampling rates (apart from the ±30 ppm of Pico's 12 MHz crystal itself).

### DIFF
--- a/tools/makeboards.py
+++ b/tools/makeboards.py
@@ -39,7 +39,7 @@ def BuildDebugLevel(name):
 
 def BuildFreq(name, defmhz):
     out = 0
-    for f in [ defmhz, 50, 100, 120, 125, 128, 133, 150, 175, 200, 225, 240, 250, 275, 300]:
+    for f in [ defmhz, 50, 100, 120, 125, 128, 133, 150, 153.6, 175, 200, 225, 240, 250, 275, 300]:
         warn = ""
         if f > defmhz: warn = " (Overclock)"
         if (out == 1) and (f == defmhz):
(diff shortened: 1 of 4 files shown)


tools/makeboards.py
```diff
@@ -39,7 +39,7 @@ def BuildDebugLevel(name):
 
 def BuildFreq(name, defmhz):
     out = 0
-    for f in [ defmhz, 50, 100, 120, 125, 128, 133, 150, 175, 200, 225, 240, 250, 275, 300]:
+    for f in [ defmhz, 50, 100, 120, 125, 128, 133, 150, 153.6, 175, 200, 225, 240, 250, 275, 300]:
```
earlephilhower, line 42: This requires re-running the script and checking in the updated boards.txt or CI will fail.

earlephilhower (changes requested):

earlephilhower:
> Should `I2S::setSysClk()` return `true` if `clk_sys` was (successfully) changed, or should it return `false` only if any change attempt was unsuccessful (or should it `panic()` instead)?

In theory if there's a return code I'd not panic and return true if the machine state on output of the function is what I requested (even if it was already good).  So your returning of the set_sys_clk is fine.

earlephilhower (approved):
Thanks!

frohro:
This is a really nice improvement!  Thanks!

PR #2716, opened 2024-12-23
Add Defcon32 Badge RP2350A board support

MeteorTheProto:
I used the pin lists and stuff so it should work. I haven't been able to test it because I cant figure out how to get a release version.


earlephilhower:
This breaks the packaging here and adds a random variants.h file in the root directory.  It also defines a bunch of things in the pins file that are not used at all in the core (the raw SDK using cmake is the only place they'd ever have an effect).

In any case, though, like I said in the issue, this isn't a generally available board and I'm against adding it.  There were maybe 100 built and that's the sum total of potential users.  And I don't think the average defcon attendee is interested in Arduino.  They've been writing raw SDK code since pre-release days!

MeteorTheProto:
Alright. I have barely begun my days with microcontrollers and the badge was a gift from a friend's mom who went to Defcon.

earlephilhower:
You can definitely use the generic RP2350 profile and adjust pins in your app to run stuff, or [read up](https://github.com/earlephilhower/arduino-pico/blob/master/docs/install.rst#installing-via-git) on using the `git` version where you can have Arduino run your local checked-out and modified copy of the core.

PR #2769, opened 2025-01-20
PIOProgram: Replace __pioHighGPIO with pio_get_gpio_base()

### LINKED ISSUE #2768: PIOProgram unaware of outside PIO usage
I have the following sketch:

```
#include <Servo.h>

Servo s1;
Servo s2;

void setup()
{
  Serial.begin(115200);
  delay(1000);
  Serial.println("Hello, world!");

  // !!! ORDER MATTERS HERE !!! //
  s1.attach(25, 500, 2500);
  s2.attach(7, 500, 2500);
  // !!! ORDER MATTERS HERE !!! //
}

void loop()
{
  s1.write(0);
  s2.write(0);
  delay(1000);
  s1.write(180);
  s2.write(180);
  delay(1000);
}
```

This is being executed on a board with an RP2350B, where `LED_BUILTIN` is actually on a GPIO pin of the RM2 radio module. The clock and data pins to the RM2 are on GPIO >= 32. I can't share much more about this particular board yet 🤫

As is, `s1` works fine, but `s2` does not. Oscilloscope shows no PWM signal at all, just constant 0V.

In the `// !!! ORDER MATTERS HERE !!! //` section, if I instead call `s2.attach()` then `s1.attach()`, or if I call `digitalWrite()` after attaching either `s1` or `s2`, then both `s1` and `s2` work fine.

I believe the root problem is the use of `__pioHighGPIO` in [PIOProgram.cpp](https://github.com/earlephilhower/arduino-pico/blob/master/cores/rp2040/PIOProgram.cpp). With the sketch above, the following happens:

1. Calling `digitalWrite(LED_BUILTIN, HIGH);` configures PIO 2.
    * PIO is needed here because the LED is actually on a GPIO pin of the RM2, and communication with the RM2 uses PIO.
    * The [CYW43 driver calls `pio_claim_free_sm_and_add_program_for_gpio_range()`](https://github.com/raspberrypi/pico-sdk/blob/95ea6acad131124694cda1c162c52cd30e0aece0/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.c#L119). Since it's the first thing to use PIO, it picks PIO block 2, SM 0 (higher PIO blocks have higher priority).
    * The clock and data pins to the RM2 are on GPIO >=32 on this particular board, so this also sets the GPIO base to 16.
    * The [CYW43 PIO program](https://github.com/raspberrypi/pico-sdk/blob/95ea6acad131124694cda1c162c52cd30e0aece0/src/rp2_common/pico_cyw43_driver/cyw43_bus_pio_spi.pio#L20-L32) is only 6 instructions long, so there's still 26 instructions available in PIO block 2.
2. Calling `s1.attach()` incorrectly sets `__pioHighGPIO[2]`.
    * As [`PIOProgram::prepare()`](https://github.com/earlephilhower/arduino-pico/blob/e20c973bf524a34ff9bc16df0c797ce0338134c3/cores/rp2040/PIOProgram.cpp#L53-L139) works through its process, it eventually concludes that [it needs to call `pio_claim_free_sm_and_add_program_for_gpio_range()`](https://github.com/earlephilhower/arduino-pico/blob/e20c973bf524a34ff9bc16df0c797ce0338134c3/cores/rp2040/PIOProgram.cpp#L118).
    * Because [the servo PIO program](https://github.com/earlephilhower/arduino-pico/blob/e20c973bf524a34ff9bc16df0c797ce0338134c3/libraries/Servo/src/servo.pio.h#L21-L32) is only 8 instructions long, and 24 instructions are still available in PIO block 2, and SM 1/2/3 are all available in PIO block 2, and `s1` uses a GPIO >= 16, it adds the servo program to PIO block 2, and assigns SM 1 for `s1`.
    * [`PIOProgram::prepare()` then sets `__pioHighGPIO[2]` to `false`](https://github.com/earlephilhower/arduino-pico/blob/e20c973bf524a34ff9bc16df0c797ce0338134c3/cores/rp2040/PIOProgram.cpp#L126) because [`s1` doesn't _need_ it](https://github.com/earlephilhower/arduino-pico/blob/e20c973bf524a34ff9bc16df0c797ce0338134c3/cores/rp2040/PIOProgram.cpp#L67), but that differs from what the GPIO base was actually set to for PIO block 2.
3. Calling `s2.attach()` incorrectly uses PIO block 2.
    * Because `__pioHighGPIO[2]` is false, `PIOProgram::prepare()` concludes it can use PIO block 2 for `s2`, despite the GPIO base is set to 16, and `s2` uses GPIO < 16.

My suggested solution is use [`pio_get_gpio_base()`](https://github.com/raspberrypi/pico-sdk/blob/95ea6acad131124694cda1c162c52cd30e0aece0/src/rp2_common/hardware_pio/include/hardware/pio.h#L759-L776) instead of `__pioHighGPIO` in [PIOProgram.cpp](https://github.com/earlephilhower/arduino-pico/blob/master/cores/rp2040/PIOProgram.cpp). I have tested this on my end and it appears to work fine (PR incoming).

I'm also skeptical about the necessity of `__pioAllocated`. Both it and `__pioHighGPIO` are treated as the source of truth for how each PIO block is configured, but that's only true if the PIOProgram class is _only_ thing messing with the PIO. IMO it would be better to use the Pico SDK API to determine whether a PIO block has been allocated (if that's even necessary in the first place).

sfe-SparkFro:
Fixes #2768 

I've also changed `needsHigh` (`true` or `false`) to `gpioBaseNeeded` (`0` or `16`) to help simplify some of the checks and prints later on. This matches the original behavior of `PIOProgram.cpp` (at least as far as my quick test showed), though I think it may be better to change it to some kind of tertiary behavior, because GPIO base 0 and 16 are both valid for GPIO 16-31. But that's an optimization better had in a different PR, because this is just meant to be a bug fix.

### DIFF
--- a/cores/rp2040/PIOProgram.cpp
+++ b/cores/rp2040/PIOProgram.cpp
@@ -33,7 +33,6 @@
 
 static std::map<const pio_program_t *, int> __pioMap[PIOCNT];
 static bool __pioAllocated[PIOCNT];
-static bool __pioHighGPIO[PIOCNT];
 auto_init_mutex(_pioMutex);
 
 PIOProgram::PIOProgram(const pio_program_t *pgm) {
@@ -64,16 +63,16 @@ bool PIOProgram::prepare(PIO *pio, int *sm, int *offset, int start, int cnt) {
     return ret;
 #endif
 
-    bool needsHigh = (start + cnt) >= 32;
-    DEBUGV("PIOProgram %p: Searching for high=%d, pins %d-%d\n", _pgm, needsHigh ? 1 : 0, start, start + cnt - 1);
+    uint gpioBaseNeeded = ((start + cnt) >= 32) ? 16 : 0;
+    DEBUGV("PIOProgram %p: Searching for base=%d, pins %d-%d\n", _pgm, gpioBaseNeeded, start, start + cnt - 1);
 
     // If it's already loaded into PIO IRAM, try and allocate in that specific PIO
     for (int o = 0; o < PIOCNT; o++) {
         auto p = __pioMap[o].find(_pgm);
-        if ((p != __pioMap[o].end()) && (__pioHighGPIO[o] == needsHigh)) {
+        if ((p != __pioMap[o].end()) && (pio_get_gpio_base(pio_get_instance(o)) == gpioBaseNeeded)) {
             int idx = pio_claim_unused_sm(pi[o], false);
             if (idx >= 0) {
-                DEBUGV("PIOProgram %p: Reusing IMEM ON PIO %p(high=%d) for pins %d-%d\n", _pgm, pi[o], __pioHighGPIO[o] ? 1 : 0, start, start + cnt - 1);
+                DEBUGV("PIOProgram %p: Reusing IMEM ON PIO %p(base=%d) for pins %d-%d\n", _pgm, pi[o], pio_get_gpio_base(pio_get_instance(o)), start, start + cnt - 1);
                 _pio = pi[o];
                 _sm = idx;
                 *pio = pi[o];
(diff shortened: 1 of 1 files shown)


earlephilhower (approved):
Great, thanks foe finding and fixing this.  LGTM!

PR #2770, opened 2025-01-21
Change duty cycle of PIO Tone to 50%

magy00:
Move `side`-setting of pin one instruction up and replace `jmp` with `.wrap` to get a 50% duty cycle.

### DIFF
--- a/cores/rp2040/Tone.cpp
+++ b/cores/rp2040/Tone.cpp
@@ -71,10 +71,7 @@ void tone(uint8_t pin, unsigned int frequency, unsigned long duration) {
         return;    // Weird deadlock case
     }
 
-    int us = 1'000'000 / frequency / 2;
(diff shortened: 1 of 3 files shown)


earlephilhower:
Neat.  Tone was one of the first PIO programs I wrote and I think it was just 1 sysclk off in terms of duty cycle.  For audio stuff, 1/133,000,000s didn't seem like an issue to me at the time.  How'd you notice it was off by a sysclk?

BTW, nice job reducing the instruction count by 1.

@Andy2No it already works atomically, so glitch free.  It basically pulls a counter load value, sets the output state, and counts down to 0 before pulling a new one (or the same one as before if not updated).

earlephilhower (approved):

Andy2No:
Cool.  Thanks, @earlephilhower and @magy00

PR #2777, opened 2025-01-24
Add buffer read to `AudioBufferManager` and `I2S`.

relic-se:
I quickly noticed after our discussion in https://github.com/earlephilhower/arduino-pico/pull/2775 that a buffered read was not available in the I2S library or the AudioBufferManager. I've added the necessary functionality to do this as well as a second loopback example demonstrating how to properly implement it. The documentation has also been updated with this method.

Notes:
- I have noticed performance improvements with this feature properly implemented.
- I've fixed an error in the documentation regarding the return value of `size_t I2S::write(const uint8_t *buffer, size_t size)`. I'm good with having that function and the new `size_t I2S::write(const uint8_t *buffer, size_t size)` return the number of words written. But to match the nature of the `size_t size` argument (size in bytes), the return could be scaled by `sizeof(uint32_t)`.

### DIFF
--- a/libraries/I2S/examples/I2SLoopback_Buffer/I2SLoopback_Buffer.ino
+++ b/libraries/I2S/examples/I2SLoopback_Buffer/I2SLoopback_Buffer.ino
@@ -1,0 +1,38 @@
+/*
+   I2S bi-directional input and output buffered loopback example
+   Released to the Public Domain by Cooper Dalrymple
+*/
+
+#include <I2S.h>
+
+I2S i2s(INPUT_PULLUP);
+
+#define SIZE 256
+int16_t buffer[SIZE];
+
+void setup() {
+  Serial.begin(115200);
+
+  i2s.setDOUT(0);
+  i2s.setDIN(1);
+  i2s.setBCLK(2); // Note: LRCLK = BCLK + 1
+  i2s.setBitsPerSample(16);
+  i2s.setFrequency(22050);
+  i2s.setBuffers(6, SIZE * sizeof(int16_t) / sizeof(int32_t));
+  i2s.begin();
+
+  size_t count, index;
+  while (1) {
+    count = i2s.read((const uint8_t *)&buffer, SIZE * sizeof(int16_t)) * sizeof(int32_t) / sizeof(int16_t);
+    index = 0;
+    while (index < count) {
+        buffer[index++]; // left
+        buffer[index++]; // right
+    }
+    i2s.write((const uint8_t *)&buffer, count * sizeof(int16_t));
+  }
+}
+
+void loop() {
+  /* Nothing here */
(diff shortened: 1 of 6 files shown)


libraries/I2S/examples/I2SLoopback_Buffer/I2SLoopback_Buffer.ino
```diff
@@ -0,0 +1,38 @@
...
+  i2s.setFrequency(22050);
+  i2s.setBuffers(6, SIZE * sizeof(int16_t) / sizeof(uint32_t));
+  i2s.begin();
+
+  size_t count, index;
+  while (1) {
+    count = i2s.read((const uint8_t *)&buffer, SIZE * sizeof(int16_t)) * sizeof(uint32_t) / sizeof(int16_t);
+    index = 0;
+    while (index < count) {
+      buffer[index++]; // left
```
earlephilhower, line 29: Were you trying to do anything here?  These are no-ops...maybe swap left and right channels or something to show data use?

relic-se: I was just showing how to get access to the samples. There's no processing happening, but swapping channels wouldn't be the worst idea. By the way, I've noticed that when converting from the int32_t words to a int16_t buffer, the right channel is packed first (ie: rrllrrll vs llrrllrr). So this example is actually a bit off in that regard. Can you confirm this?

libraries/I2S/src/I2S.cpp
```diff
@@ -559,6 +559,14 @@ bool I2S::read32(int32_t *l, int32_t *r) {
     return true;
 }
 
+size_t I2S::read(const uint8_t *buffer, size_t size) {
```
earlephilhower, line 562: Same `const` thought.  POSIX `write` is non-const, only `read` is.  We're overwriting data in the ptr so it's really not const anyway (wouldn't be very useful of a read if we couldn't get new data :laughing: )

earlephilhower (changes requested):
Minor `const` question, but OTW thanks very much for the contribution!  I focused my effort on making the output path faster since I use it for https://github.com/earlephilhower/BackgroundAudio and https://github.com/earlephilhower/ESP8266Audio so you're hitting these bottlenecks.  Thx again for taking the time to help alleviate them!

relic-se:
@earlephilhower Everything should be good to go now. Thanks for the review!

earlephilhower (approved):
Great, thx!

relic-se:
@earlephilhower  Thanks for catching that last `const`!

earlephilhower:
No worries.  C casting footgun avoided.  Probably others in my code still lurking... Thanks again!

PR #2790, opened 2025-02-03
PlatformIO PSRam Pimoroni Pico 2 (2350) Demo

webzep:
Being a bit of a noob in this area, I had a lot of trouble getting PSRAM working on PIO with the pimoroni pico plus 2.

So here is a demo for future noobs like myself that need a bit of hand holding.

Thanks for your work and relentlessly patient attitude Earle, its refreshing to deal with a friendly maintainer.

### DIFF
--- a/libraries/rp2350/examples/PlatformioPSRAMDemo/PsramAllocator.h
+++ b/libraries/rp2350/examples/PlatformioPSRAMDemo/PsramAllocator.h
@@ -1,0 +1,38 @@
+#ifndef PSRAM_ALLOCATOR_H
+#define PSRAM_ALLOCATOR_H
+
+#include <Arduino.h>
+
+extern "C"
+{
+	void *pmalloc(size_t size);
+}
+
+template <class T>
+struct PsramAllocator
+{
+	using value_type = T;
+
+	PsramAllocator() noexcept {}
+	template <class U>
+	PsramAllocator(const PsramAllocator<U> &) noexcept {}
+
+	T *allocate(std::size_t n)
+	{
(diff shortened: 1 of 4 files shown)


maxgerhardt:
The way I see it, the PlatformIO specific part is just about configuring

```
build_flags = -fexceptions
              -DRP2350_PSRAM_CS=47
board_build.pico_boot2_name = boot2_psram8.S
board_build.pico_psram_size = 8
board_upload.psram_length = 8388608
```

in regards to that: 
1. Enabling exceptions is also documented in https://arduino-pico.readthedocs.io/en/latest/platformio.html#c-exceptions
2. Configuring the PSRAM CS will be documented through #2792 
3. Setting the Boot2 source file will be documented through #2792. Also the `board_build.pico_psram_size` config name is wrong / unused? It should have no effect because `board_build.arduino.earlephilhower.boot2_source` is used
5. PSRAM length was already documented at https://arduino-pico.readthedocs.io/en/latest/platformio.html#psram-size. Also `board_build.pico_psram_size` should have no effect because, the next line is actually correctly configuring it for 8MB.

earlephilhower:
While this probably isn't necessary as a P.IO example, the idea of implementing an`std::allocator` which uses PSRAM seems like a neat trick that might save some grief.  One of those things you don't need until you **really** need it.

If the P.IO bit isn't quite right in the configuration here, maybe we can drop that bit and leave just the actual INO as an example of a PSRAM allocator?  (BTW, CI is failing due to formatting.  You need to run `tools/restyle.sh` or copy the existing INO example(Arduino) style w/2 space indents to get that to pass.

PR #2795, opened 2025-02-08
add new RP2350 Metro!

ladyada:
just tested this fancy new hardware :)


earlephilhower (changes requested):
Nice, but we need to get the generated platform.io JSON file from `tools/json/` added to the commit, please.

ladyada:
woops late night commits! re-trying now with files

earlephilhower (approved):
Thx!

ladyada:
thank you!

PR #2796, opened 2025-02-08
fix: Wire::writeReadAsync always malloc'ing DMA buffer

veecle-stefan:
The async implementation of I2C (Wire) seems to allocate the DMA buffer for every call because the check, whether the current buffer is large enough (`if (_dmaSendBufferLen < bufLen) {`) is always `true` because `_dmaSendBufferLen` is never actually set anywhere. 

Simple fix: added a line after successful `malloc` that sets the `_dmaSendBufferLen = bufLen`. 

PS: Shouldn't we use `new uint16_t[bufLen]` here, since we're in a C++ object? It probably maps to `malloc` but it would at least avoid one cast.

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -507,7 +507,7 @@ bool TwoWire::writeReadAsync(uint8_t address, const void *wbuffer, size_t wbytes
     abortAsync();
 
     // Create or enlarge dma command buffer, we need one entry for every i2c byte we want to write/read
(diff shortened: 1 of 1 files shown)


earlephilhower:
Great catch, thanks!

I'm fine leaving `malloc/free` in place of `new[]/delete[]` here.  For primitives there's no consistency here and it depends on what I was feeling like the day I wrote the code.  The one gotcha, though, is that `new` could `throw` which we can't catch (unless you explicitly enable exceptions in the IDE and blow up your code size).  They map down to `malloc` and `free` in any case, like you said...

PR #2797, opened 2025-02-08
Move to Arduino API 10501

earlephilhower:
Track upstream Arduino API headers

### DIFF
--- a/cores/rp2040/Arduino.h
+++ b/cores/rp2040/Arduino.h
@@ -110,7 +110,7 @@ extern bool __isFreeRTOS;
 #endif
 
 #ifndef PGM_VOID_P
-#define PGM_VOID_P void *
+#define PGM_VOID_P const void *
 #endif
 
 #ifdef __cplusplus
@@ -158,4 +158,8 @@ constexpr uint64_t __bitset(const int (&a)[N], size_t i = 0U) {
 #else
 #define __GPIOCNT 30
 #define __FIRSTANALOGGPIO 26
+#endif
+
+#ifdef __cplusplus
+using namespace arduino;
 #endif
--- a/cores/rp2040/SerialPIO.h
+++ b/cores/rp2040/SerialPIO.h
@@ -29,7 +29,7 @@
 
 extern "C" typedef struct uart_inst uart_inst_t;
 
-class SerialPIO : public HardwareSerial {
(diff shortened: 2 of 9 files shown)


cores/rp2040/Arduino.h
```diff
@@ -159,3 +159,7 @@ constexpr uint64_t __bitset(const int (&a)[N], size_t i = 0U) {
 #define __GPIOCNT 30
 #define __FIRSTANALOGGPIO 26
 #endif
+
+#ifdef __cplusplus
+using namespace arduino;
+#endif
```
mathieucarbou, line 165: Hi,

Is it possible that this namespace introduction causes this build failure in many projects ?

Example: https://github.com/ayushsharma82/NetWizard/actions/runs/13421632127/job/37495411747#step:15:2

I had to fix the CI to 4.4.3 for ESPAsyncWebServer: https://github.com/ESP32Async/ESPAsyncWebServer/commit/306dcf6740018eee8d34581aa0464992f62aaff8

Thanks!

@earlephilhower @ayushsharma82

earlephilhower: I suppose it's possible to cause a problem, but very unlikely because in the prior release there was a blanket `using namespace arduino` in the API headers under `Print`.  https://github.com/arduino/ArduinoCore-API/commit/db53d3d51879ec2253ed4288a926213b75cff633  So it's always been there, just moved to our own file and not the API sources with this PR (I forked the API when the project was just started).

The removal in the latest Arduino APIs actually caused problems with libraries here (TinyUSB off the top of my head).  I just re-instated it here so it's effectively a no-op.

There were some API additions/changes with `String` (and bug fixes, the reason for moving fwd) so I think they're the reason CI is failing for you there, not the `using` statement.  GCC knows to use the `arduino::String`, but can't figure out a cast that will work with the code in that file, it seems.  Extra explicit casting might help you there...

earlephilhower: Actually looks like you can just explicitly make etag a String and make progress (bad hand written DIFF follows...)
https://github.com/ESP32Async/ESPAsyncWebServer/blob/306dcf6740018eee8d34581aa0464992f62aaff8/src/WebHandlers.cpp#L225
````
- etag = request->_tempFile.size();
+ etag = String((int)request->_tempFile.size());
````

PR #2798, opened 2025-02-09
added I2C TwoWire::busIdle to be checked by writeReadAsync

veecle-stefan:
**Problem**  
When using `writeReadAsync` in a multi-threaded environment (e.g. with FreeRTOS), the DMA interrupt fires **immediately** after the TX FIFO is loaded—not when the I2C peripheral has actually finished clocking out those bytes. As a result, a *new* `writeReadAsync` call might start too early and interrupt the ongoing hardware transfer. This issue only affects write-only transactions (`rbytes == 0`), because read transactions already use a different interrupt that waits for RX completion.

**Solution**  
This PR adds a `bool busIdle()` function that checks if the I2C hardware is truly idle (i.e., no master activity and the transmit FIFO is empty). Inside `writeReadAsync`, if the bus is still busy, we return `false`—just as we already do if a DMA transfer is in progress. This allows the caller to detect the “bus still busy” case and retry the operation once the bus becomes free.

**Benefits**  
- Prevents partial writes when a new transaction starts too soon.  
- Allows cleaner multi-thread or asynchronous usage of `writeReadAsync` by giving callers a straightforward way to avoid interrupting a still-active write.  
- In testing, this improved stability and reduced CPU load significantly when accessing complex sensors (e.g. Adafruit BNO085) at higher rates.


earlephilhower (approved):
Nice addition, looks good to me!

PR #2819, opened 2025-02-23
Support Makerbase MKSTHR36 and MKSTHR42 Boards

michael-ring:
The two Makerbase Boards described here:

https://makerbase3d.com/product/mks-thr36-thr42-board

offer a nice solution for people that need to build Focussers for Astro Equipment or simply need only one Stepper Motor for their project.

I have created (hopefully) all the required files to make those boards work, at least for me, they work just fine with those changes.


earlephilhower:
You're going to need to define `PIN_SERIAL(1/2)_(RX/TX)` for the two boards to get a working build.  Looks like some were partially set in the header, but not all 4 settings for each board.

michael-ring:
Sorry, should work now

earlephilhower:
No worries, and thanks for the contribution!

PR #2825, opened 2025-02-25
Define HSTX pins on Adafruit boards with HSTX or DVI connectors

jepler:
These macros are used by Adafruit_dvhstx to put the signals on the right HSTX pins.


jepler:
ping @BlitzCityDIY @ladyada

ladyada (approved):

jepler:
thank you!

PR #2831, opened 2025-02-28
early out of (un)maskInterrupts() if no GPIO interrupts need to be masked

jepler:
My application is designed to generate HSTX data on core0 in interrupts, but also uses hardware SPI for SD card access.

It turns out that the amount of time spent in `maskInterrupts`/`unmaskInterrupts` with interrupts disabled, even with an empty `_usingIRQs`, is too long.

Add a quick check and avoid touching the interrupt disable flag if there's not actually any GPIO interrupt to (un)mask.

@ladyada this fixes the hstx video glitches during SPI SD card access

### DIFF
--- a/libraries/SPI/src/SPIHelper.h
+++ b/libraries/SPI/src/SPIHelper.h
@@ -108,6 +108,7 @@ public:
         @brief Disables any GPIO interrupts registered before an SPI transaction begins
     */
     void maskInterrupts() {
+        if(_usingIRQs.empty()) { return; }
         noInterrupts(); // Avoid possible race conditions if IRQ comes in while main app is in middle of this
         // Disable any IRQs that are being used for SPI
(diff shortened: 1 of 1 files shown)


earlephilhower (commented):
Wow, that's tight timing.  I wonder if the STL is doing actual memory allocations/etc. in the `for (auto x: y)` iteration even with an empty list...

You'll need to adjust formatting using `tools/restyle.sh` or just manually indent/format the new short-circuit path to get through CI.

-edit- update formatting -edit-
````
 if (_usingIRQs.empty()) {
    return;
}
````

OTW, LGTM!

jepler:
Thanks, I've tried to update the branch with the correct style.

earlephilhower (approved):
Third time's the charm. :laughing:

jepler:
thank you

PR #2836, opened 2025-03-03
Add upcoming SparkFun IoT RedBoard RP2350

SFE-Brudnerd:
Requires pico-sdk [PR#2329](https://github.com/raspberrypi/pico-sdk/pull/2329) to be merged.


earlephilhower:
We use a single generic `libpico` built from the SDK and don't depend on the SDK's board.h files.  So, this should be fine as-is.  If CI passes I'll merge and you'll be set...

SFE-Brudnerd:
Great, thanks for letting me know!

Issue #2830, opened 2025-02-28
A2DP would benefit from exposing more BluetoothHCI functions


InfiniteCoder01:
Exposing `bool scanAsyncDone();` and `std::vector<BTDeviceInfo> scanAsyncResult();` in A2DPSource would be useful for handling async connections. Might even be better to do something along the lines `BluetoothHCI &getHCI() { return _hci; }`

earlephilhower:
I didn't really think about the async scanning methods, they kind of came along as I refactored things so obviously missing bits like that still exist.  Would you like to make a PR implementing the `scanAsyncDone` and `scanAsyncResult` methods?

InfiniteCoder01:
Size in bytes, samples and uint32_ts seems to also be mixed up in A2DPSource. Should I also fix it?

PR #2839, opened 2025-03-05
A2DP: scanAsyncDone & scanAsyncResults; fix write and availableForWrite

InfiniteCoder01:
See #2830


earlephilhower (approved):
Thanks.  The availableForWrite fix is a good start to what I was doing with #2765...

PR #2843, opened 2025-03-10
Bidirectional TDM Support

efairbanks:
Adding support for bidirectional TDM support. Sets are on falling BCLK to maintain consistency with other TDM/I2S code here, and TDM implementation assumes that this is I2S-compatible TDM.

Tested all four channels working on AK4619 4i4o codec with the following register settings:

```
0x01: 0b10101100
0x02: 0b00011100
0x03: 0b00000011
```

See AK4619 datasheet for more details, though this should work with any I2S-compatible TDM codec.


earlephilhower:
Haven't tried it yet, but thanks!

For the pio->h conversion, it looks like you ran an old pioasm executable, hence the missing field in the generated structures.  We ship a cli pioasm as part of the tools install here, you'll find it same spot as the other exes like gcc and openocd.

efairbanks:
Ok @earlephilhower done. This is tested and working on an RP2350 hooked up to an AK4619 codec. Hopefully this is helpful to other folks who want to use codecs like this AsahiKASEI one or those Cirrus Logic codecs that are so popular.

earlephilhower:
CI failure needs the formatting fixes shown in the diff.  If you can run `./tools/restyle.sh` it'll reformat things for you automatically.

PR #2847, opened 2025-03-11
Add SparkFun XRP Controller

sfe-SparkFro:
Upcoming board from us!


package/package_pico_index.template.json
```diff
@@ -317,6 +317,9 @@
                   {
                      "name": "SparkFun XRP Controller (Beta)"
                   },
+                  {
```
earlephilhower, line 320: Should we remove the other variant?  `SparkFun XRP Controller (Beta)`?

sfe-SparkFro: No, please keep both. This is for a new version of the same product, but we're planning to maintain support for the old board for a while.

PR #2852, opened 2025-03-13
Redo UF2 discovery for Windows compatibility

earlephilhower:
Windows Python doesn't seem to kill the worker Thread properly when the IDE is exited, leading to a) multiple Python3 instances on a PC after many uses and b) errors updating the core when it tries to re-install Python3 while still having the older version's EXE loaded and in use.

Works around https://github.com/arduino/arduino-cli/issues/2867

### DIFF
--- a/tools/pluggable_discovery.py
+++ b/tools/pluggable_discovery.py
@@ -5,56 +5,59 @@ import sys
 import time
 import threading
 
-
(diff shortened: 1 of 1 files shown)


earlephilhower:
Alas, this does seem to leave dangling processes when called from arduino-cli.

Issue #2932, opened 2025-05-01
Faster parity calculations for SerialPIO when using multiple instances


ahmedarif193:
Right now, SerialPIO calculates parity with a bit loop that gets slower with more data bits. If you're running several soft serial ports (like 4+), it starts eating CPU for no good reason.

A small lookup table (256 bytes) could do the same job way faster, basically constant time, just a few instructions per byte versus 60+ instructions for 8-bit data for example.

Tradeoffs:

- Speed: about 10–12x faster per check
- RAM: 256B global LUT (shared by all instances)
- Determinism: fixed cost per check

I'm happy to send a PR for this.

earlephilhower:
Fixed by #2933

PR #2933, opened 2025-05-01
optimize parity calculations in SerialPIO (#2932)

ahmedarif193:
use LUT for faster parity checks, reduces CPU load with multiple instances

### DIFF
--- a/cores/rp2040/SerialPIO.cpp
+++ b/cores/rp2040/SerialPIO.cpp
@@ -68,13 +68,10 @@ static PIOProgram *_getRxProgram(int bits) {
 }
 // ------------------------------------------------------------------------
 
-// TODO - this works, but there must be a faster/better way...
-static int __not_in_flash_func(_parity)(int bits, int data) {
-    int p = 0;
-    for (int b = 0; b < bits; b++) {
-        p ^= (data & (1 << b)) ? 1 : 0;
-    }
-    return p;
+static int __not_in_flash_func(_parity)(int data) {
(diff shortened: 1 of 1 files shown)


earlephilhower (changes requested):
Thanks, but this is pretty expensive in terms of ROM vs. what it's (hopefully) accelerating.  256 bytes of bytewide 1s and 0s just feels wasteful.  And since it's in ROM it could be a very slow access (XIP is nice, but nobody's going to call it fast...).

I'm not sure that parity on any bitrate that the SerialPIO operates at is a limiting factor, but if you really think it is then why not something simpler with much lower space requirements. 

For example, (stolen from http://www.graphics.stanford.edu/%7Eseander/bithacks.html#ParityParallel) you can do parity in ~5 instructions without LUT by utilizing a constant as a parity bitmap:
````
static int __not_in_flash_func(_parity)(int data) {
    data ^= data >> 4;
    data &= 0xf;
    return (0x6996 >> data) & 1;
}
````
There shouldn't be a need to mask off bits before doing the math in the above or a LUT.  That was there to short-circuit in the original for loop case.

ahmedarif193:
that bit hack is way more brilliant. I'm using this for running 8 rx channels on my rp2040 with real-time constraints alongside other critical tasks. cpu usage matters when you're maxing out a cortex-m0 cpu because that cpu has potential indeed

earlephilhower (approved):
Thx, LGTM!

PR #2937, opened 2025-05-05
Fix ADCInput masking

### LINKED ISSUE #2935: ADCInput + I2S project loopbak stop working on new plugin versions.
Good morning, I'm using a combination of the ADCInput API and I2S as an output device. In general in the project that uses Raspberry Pico Pi 2, I read a buffer of 128 samples, filter them, and write them back to the I2S output. Everything worked perfectly with previous versions of the plugin (4.3.1) but with the new versions the acquisition of the signal from ADC seems to be extremely slowed down generating continuous noise.

I also tryed to make a simple test of loopbak reading and writing at the same frequencies
using A0 as sampling PIN and PCM5202 as Output DAC.

The samples rate tested was 8000 16000 and more ... 
The BUFFER_SIZE was various 32 ...
The setup:

  i2s.setBCLK(I2S_BCLK);
  i2s.setDATA(I2S_DOUT);
  i2s.setBitsPerSample(16);

  if (!i2s.begin(sampleRate)) {
    while (1); // do nothing
  }

  adcIn.setBuffers(4, BUFFER_SIZE);

  if (!adcIn.begin(sampleRate)) {
    while (1); 
  }

... The loop 

while (adcIn.available() > 0) {

    newSample = adcIn.read();
    
     i2s.write(newSample);
     i2s.write(newSample);
   
  };

All works correctly on 4.3.1 , but not in the new release of plugin and use It with success in several projects in ham radio processing signal.

The complete  project Is : https://github.com/gcallipo/RadioDSP-Pico

Thanks for the great effort to produce this library.
Giuseppe.

earlephilhower:
Fixes #2935

The PICO_2350A patch flipped the logic of the ADCInput bitmask calcs causing the 2040 and 2350A to use bitmasks for the 2350B (i.e. wrong GPIOs) and vice versa.

### DIFF
--- a/libraries/ADCInput/src/ADCInput.cpp
+++ b/libraries/ADCInput/src/ADCInput.cpp
@@ -50,11 +50,6 @@ bool ADCInput::setBuffers(size_t buffers, size_t bufferWords) {
 int ADCInput::_mask(pin_size_t p) {
     switch (p) {
 #if defined(PICO_RP2350) && !PICO_RP2350A // RP2350B
-    case 26: return 1;
-    case 27: return 2;
-    case 28: return 4;
-    case 29: return 8;
-#else // Starts at 40 and there are 8 of them
     case 40: return 1;
     case 41: return 2;
     case 42: return 4;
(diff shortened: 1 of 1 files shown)


gcallipo:
I checked and can confirm the fix solved the issue #2935 
Thanks.

PR #2938, opened 2025-05-06
Make Arduino IDE ask for OTA password

progamler:
Arduino OTA currently not working when its password Protected. 
adding this line enables the configuration of the Password


earlephilhower (approved):
Thanks!

PR #2962, opened 2025-05-18
Added 'TwoWire::setBufferSize()' to Wire library

Lorandil:
Hi,

As you suggested I added the missing **'setBufferSize()'** method to the Wire library,
The method name and #define are taken from the ESP32 code.

I had to make some design decisions:

- because most people won't use this feature, I chose to allocate the default buffer size already in the c'tor to prevent unncessary heap fragmenation
- I implemented a lower limit to the buffer size (32 bytes) - what do you think?
- the memory is only freed if a new buffer size is required, this again should keep the heap clean.

The code worked fine in my test environment ;)

Cheers,
Sven


earlephilhower (changes requested):
Looks good, thanks for implementing the option.

You'll need to clean up some minor formatting issues shown in CI (use tools/restyle.sh under Linux to automatically fix those) and some warnings about signedness you need to explicitly cast away.

Lorandil:
I fixed an oversight of mine (buffer is also used on write) and added the casts.
Hoping for the best :wink:

Lorandil:
> You'll need to clean up some minor formatting issues shown in CI (use tools/restyle.sh under Linux to automatically fix those) and some warnings about signedness you need to explicitly cast away.

I would gladly resolve the issue, but I really don't get what the problem is.
The check's output doesn't give me any clues what's wrong (at least none I discovered) and there is no 'restyle.sh' in tools (or maybe I'm just blind :wink:).

I would really appreciate, if you could point me in the right direction here...

Regards,
Sven

earlephilhower:
You need to remove the spaces on blank lines that the CI identified, please.

earlephilhower:
And if you have astyle and this repo, you can just run `./tests/restyle.sh` to clean them up automatically.

earlephilhower:
Took a little while to get there, but it got there in the end. :laughing:

PR #2976, opened 2025-05-30
Force DHCP request on WiFiMulti ::run connection

### LINKED ISSUE #2974: WiFi issue !
Arduino Ide RPI Pico W library 4.5.4 installed

two different WiFi's with DHCP and different SSID and different ip ranges

Same behavior with WiFIMulti or while doing manually if SSID no more available WiFi.mode(WIFI_OFF) then delay(500) then WiFi.mode(WIFI_STA) and at last WiFi.begin(ssid,passwd)

it establishes a connection , gets IP etc and this connection works
after connection lost and a new connection is established to another SSID IP address is not updated the Pico W stays with the address received by the first successful connection . On this case the new connection doesn't work cause the Pico address doesn't fit to the range of the new SSID.

earlephilhower:
The IP address is not cleared (even if it is DHCP acquired) between ::begins on the Ethernet classes by design (#1866).  But it is possible that the WiFi networks being polled by WiFiMulti are different network configurations, so force a DHCP request explicitly before (re)connecting).

Fixes #2974

### DIFF
--- a/libraries/WiFi/src/WiFiMulti.cpp
+++ b/libraries/WiFi/src/WiFiMulti.cpp
@@ -109,6 +109,9 @@ uint8_t WiFiMulti::run(uint32_t to) {
                j->bssid[0], j->bssid[1], j->bssid[2], j->bssid[3], j->bssid[4], j->bssid[5], j->rssi);
     }
 
+    // Force DHCP request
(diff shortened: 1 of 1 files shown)


earlephilhower:
Closing as this treats the symptom and not the cause.  See #2977

PR #2999, opened 2025-06-18
Add support for W6300-EVB-Pico and W6300-EVB-Pico2 boards

wiznet-mason:
Hello,

This PR adds support for the W6300-EVB-Pico and W6300-EVB-Pico2 boards from WIZnet. 
These boards combine RP2040 or RP2350 with WIZnet's W6300 Ethernet controller via a QSPI interface implemented using PIO.

https://docs.wiznet.io/Product/iEthernet/W6300/w6300-evb-pico
https://docs.wiznet.io/Product/iEthernet/W6300/w6300-evb-pico2

Thank you very much for your consideration!

### DIFF
--- a/libraries/lwIP_w6300/src/utility/w6300.cpp
+++ b/libraries/lwIP_w6300/src/utility/w6300.cpp
@@ -51,8 +51,7 @@ wiznet_pio_qspi_handle_t wiznet_pio_qspi_handle = NULL;
     Check physical link
     @return true when physical link is up
 */
-bool Wiznet6300::isLinked()
-{
+bool Wiznet6300::isLinked() {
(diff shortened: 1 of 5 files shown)


wiznet-mason:
Hi,

Thanks for your feedback. However, since W6300  requires PIO to  emulate GPIOs to be QSPI pins, therefore we included definitions for all related pins in w6300.h. Thus it is not needed to define separately.

Upon tests on our side, everything works smoothly. Let us know if this approach is wrong and we will fix it for this chip and others too.

wiznet-mason:
Thank you for merging.
Should you have any issues, please contact me.

PR #3000, opened 2025-06-18
Add Waveshare RP2350 Plus

shubham0x13:
Add support for [Waveshare RP2350 Plus](https://www.waveshare.com/wiki/RP2350-Plus) boards with 4MB and 16MB flash variants.


earlephilhower (approved):
Thx!

PR #3001, opened 2025-06-19
Fix JLink device names for RP2350 in PIO boards

maxgerhardt:
Fixes https://github.com/maxgerhardt/platform-raspberrypi/issues/104 at the source.

Now the JLink `-device` arguments should be correct, see
* https://www.segger.com/supported-devices/search/RP2350
* https://www.segger.com/supported-devices/search/RP2040

(Does interfere with other PRs that add boards, so either those PRs or this PR will have to re-run the makeboards.py generation.)

### DIFF
--- a/tools/json/waveshare_rp2350_plus.json
+++ b/tools/json/waveshare_rp2350_plus.json
@@ -25,7 +25,7 @@
         "variant": "waveshare_rp2350_plus"
     },
     "debug": {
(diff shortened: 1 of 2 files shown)


earlephilhower:
@maxgerhardt the other PRs just merged, can you please re-run makeboards to get the 3 new board JSONs?

PR #3016, opened 2025-06-28
Refactor Waveshare RP2040 Plus with flash size menu

shubham0x13:
Replaces separate board entries for Waveshare RP2040 Plus 4MB and 16MB variants with a flash size menu.


earlephilhower (approved):
Thanks for finding and fixing this!

PR #3026, opened 2025-07-06
add DVI pin mapping to pins_arduino.h

SamHalvoe:
I use the Adafruit library https://github.com/adafruit/Adafruit-DVI-HSTX, which implements DVI over HSTX.
You need the correct pin mapping for tzhe HSTX pins to use this library. Adafruit HSTX boards define this mapping.
I would like to add this to the Pimoroni Pico Plus 2, so that you can use the Adafruit library out of the box without headaches how to map the HSTX pins.

### DIFF
--- a/variants/pimoroni_pico_plus_2/pins_arduino.h
+++ b/variants/pimoroni_pico_plus_2/pins_arduino.h
@@ -51,6 +51,16 @@
 
 #define PICO_RP2350A 0 // RP2350B
 
+// DVI connector
+#define PIN_CKN (15u)
+#define PIN_CKP (14u)
+#define PIN_D0N (13u)
+#define PIN_D0P (12u)
+#define PIN_D1N (19u)
+#define PIN_D1P (18u)
+#define PIN_D2N (17u)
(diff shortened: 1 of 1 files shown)


earlephilhower:
Thanks, but aren't the pins fixed by the hardware?  Meaning there should be one definition for all RP2350 boards somewhere, or just assumed by the library?

SamHalvoe:
I think the HSTX is fixed, but the mapping to the GPIO is specific to each board. But can't tell for sure.
See the Adafruit HSTX board, it has a different [mapping](https://github.com/earlephilhower/arduino-pico/blob/master/variants%2Fadafruit_feather_rp2350_hstx%2Fpins_arduino.h#L44-L52) than the Pimoroni Pico Plus 2.

SamHalvoe:
I assume that the Raspberry Pi Pico 2 and Raspberry Pi Pico 2 W have the same mapping as the Pimoroni boards.

earlephilhower:
Ah, got it.  The pins you can use are fixed, but the function of each GPIO is configurable.  Thx!

PR #3034, opened 2025-07-11
Rewrite HTTPClient chunked transfer-encoding

### LINKED ISSUE #3029: HTTP client drops data with chunked transport
In reference to https://github.com/earlephilhower/BackgroundAudio/issues/61

I tried to get the background audio HTTP stream example to work, but its constantly stuttering without actually running behind. According to the error messages, there's large amounts of corruption in the received file.

I investigated by having the HTTP client print out the first three bytes of the file. I did not receive "ID3" as expected, so I investigated further. I modified the HTTP stream example as shown:

```cpp
/**
   StreamHTTPClient.ino

    Created on: 24.05.2015

*/

#include <Arduino.h>
#include <WiFi.h>
#include <HTTPClient.h>

#ifndef STASSID
#define STASSID "<redacted>"
#define STAPSK "<redacted>"
#endif

const char *ssid = STASSID;
const char *pass = STAPSK;

WiFiMulti WiFiMulti;

void setup() {

  Serial.begin(11520000);
  // Serial.setDebugOutput(true);

  Serial.println();
  Serial.println();
  Serial.println();

  for (uint8_t t = 4; t > 0; t--) {
    Serial.printf("[SETUP] WAIT %d...\n", t);
    Serial.flush();
    delay(1000);
  }

  WiFi.mode(WIFI_STA);
  WiFiMulti.addAP(ssid, pass);
}

void loop() {
  // wait for WiFi connection
  if ((WiFiMulti.run() == WL_CONNECTED)) {

    Serial.print("[HTTPS] begin...\n");

    HTTPClient https;
    https.setInsecure();

    if (https.begin("https://path/to/my/mp3")) {

      Serial.print("[HTTPS] GET...\n");
      // start connection and send HTTP header
      int httpCode = https.GET();
      if (httpCode > 0) {
        // HTTP header has been send and Server response header has been handled
        Serial.printf("[HTTPS] GET... code: %d\n", httpCode);

        // file found at server
        if (httpCode == HTTP_CODE_OK) {

          // get length of document (is -1 when Server sends no Content-Length header)
          int len = https.getSize();

          // create buffer for read
          static uint8_t buff[128] = { 0 };

          // read all data from server
          while (https.connected() && (len > 0 || len == -1)) {
            // get available data size
            size_t size = https.getStreamPtr()->available();

            if (size) {
              // read up to 128 byte
              int c = https.getStreamPtr()->readBytes(buff, ((size > sizeof(buff)) ? sizeof(buff) : size));

              // write it to Serial
              Serial.write(buff, c);

              if (len > 0) {
                len -= c;
              }
            }
            delay(1);
          }

          Serial.println();
          Serial.print("[HTTPS] connection closed or file end.\n");
        }
      } else {
        Serial.printf("[HTTPS] GET... failed, error: %s\n", https.errorToString(httpCode).c_str());
      }

      https.end();
    } else {
      Serial.printf("Unable to connect\n");
    }
  }

  Serial.println("Wait 10s before the next round...");
  delay(10000);
}

```

And I read this onto my computer with:
```bash
(stty raw; cat > received.log) < /dev/ttyACM0
```

I trimmed the file in a hex editor so that it only displayed the MP3. This file *did* start with ID3 as expected, but there were massive differences. The audio stuttered on playback on my PC just as it did on the Pi.

It seems like the HTTP client is dropping characters every now and then. This likely has something to do with the larger flow rate since I'm able to maintain a websocket connection for long periods of time with no problem. It is worth noting: the server *will* attempt to send the whole file unless the Pi slows it down. Does the Pi not rate-limit the TCP connections, and instead discard the data?

earlephilhower:
Fixes #3029

The chunked decoder seems to have had some race conditions resulting in lost data.

The HTTPClient::getStreamPtr returned the raw WiFiClient which included all of the chunked block size markers, requiring the user to manually parse the chunked encoding.

Remove the existing chunked handling, add a HTTPStream as a WiFiClient subclass.  The HTTPClient will return this HTTPStream which will always properly cut out chunk markers when present (and pass things raw for non-encoded streams).  Use this new HTTPStream class to replace the existing handling in HTTPClient

Update the fingerprint in the StreamHTTPSClient example because their cert was updated.


KB1RD:
It would be very helpful to add a `isEof()` function or similar to detect the end of a chunked/normal HTTP stream, if its not too much trouble. I don't see a way to tell the difference between an EOF and no data, unless I'm mistaken.

KB1RD:
I'm noticing one small issue: The write function does not appear to behave in the same way as the one before the changes. My outbound websocket connection now seems unable to write any data. Will continue investigating.

EDIT: This is the code writing to the websocket:
```cpp
bool WebsocketConnection::sendString(const char* str) {
  size_t len = strlen(str);
  // I'm only supporting sending short packets rn
  if (len >= 126) { return false; }

  // Write header
  client->write(0x81);
  client->write((uint8_t)len); // I'm also not masking because I'm really lazy, and I know I've configured my reverse proxy properly

  // Write body
  client->write((const uint8_t*)str, len);

  return true;
}
```
The 0x81 and the 0x1b gets written to the websocket (first two writes), but the larger write from a buffer never ends up getting sent.

EDIT 2: The solution is to add:
```cpp
    size_t write(const uint8_t *buf, size_t size) {
        return _conn->write(buf, size);
    }
```
to the HTTPStream. I'm surprised the WiFi client doesn't rely on the interface single-char `write` method to do this, though.

earlephilhower:
Thanks for the update and good catch.  It's a pain this is a `WiFiClient*` and not a `Stream *` and complicates things.

At a minimum I should pass through all "reasonable" virtual calls.  Let me look at what might make sense...

PR #3039, opened 2025-07-16
Added a function to allow inverting the output from a GPIO pin

invectorlabs:
This allows a user to simply by calling the setPinInvert function to invert the output of a GPIO pin instead of rewriting existing code.... if required =) I am not sure if the function declaration is correctly positioned though so please advice.

### DIFF
--- a/cores/rp2040/wiring_digital.cpp
+++ b/cores/rp2040/wiring_digital.cpp
@@ -25,13 +25,27 @@ extern void __clearADCPin(pin_size_t p);
 
 static PinMode _pm[__GPIOCNT];
 
+extern "C" void setPinInvert(pin_size_t ulPin, bool invert) __attribute__((weak, alias("__setPinInvert")));
+extern "C" void __setPinInvert(pin_size_t ulPin, bool invert) {
+    if (ulPin >= __GPIOCNT) {
+        DEBUGCORE("ERROR: Illegal pin in setPinInvert (%d)\n", ulPin);
+        return;
+    }
+
+    if (invert)
+        gpio_set_outover(ulPin, GPIO_OVERRIDE_INVERT);
+    else
+        gpio_set_outover(ulPin, GPIO_OVERRIDE_NORMAL);
+}
+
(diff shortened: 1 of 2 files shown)


earlephilhower:
Thanks, but could we add the declaration to `Arduino.h` like the other [simple additions](https://github.com/earlephilhower/arduino-pico/blob/e7008b634c4c1bb4c68992fc1e347e8ea3bf860e/cores/rp2040/Arduino.h#L103) that don't need the Arduino-Core API headers to be changed (I try and sync them up every now and then).

Also, a line or two in the docs about this and an example (something simple like blink.ino but with toggling the invert pin) would be much appreciated.  It's non-standard so users won't be able to discover it otherwise...

cores/rp2040/wiring_digital.cpp
```diff
@@ -90,7 +104,6 @@ extern "C" void __digitalWrite(pin_size_t ulPin, PinStatus ulVal) {
         DEBUGCORE("ERROR: Illegal pin in pinMode (%d)\n", ulPin);
         return;
     }
-    gpio_set_function(ulPin, GPIO_FUNC_SIO);
```
earlephilhower, line 93: While adding this in `pinMode` seems fine, I think deleting here will break the case where you have someone else controlling the pin (i.e. `tone` which uses PIO) and then you stop that process and then `digitalWrite`.  Let me whip up a simple test case...

earlephilhower: Yes, this breaks normal functionality.  Can you re-add the line?

Simple example blinks after analogWrite with `master`, remains PWN (connected to PIO) with this patch
````
void setup() {
  pinMode(LED_BUILTIN, OUTPUT);
  analogWriteRange(100);
  analogWrite(LED_BUILTIN, 50);
  delay(1000);
  analogWrite(LED_BUILTIN, 70);
  delay(1000);
  delay(1000);
}

void loop() {
  delay(1000);
  delay(1000);
}
````

earlephilhower (changes requested):

earlephilhower:
Closing just to clean things up.  If you do ever get the chance to go back to this, we can reopen.  Thx!

PR #3049, opened 2025-07-29
Update to Pico-SDK 2.2.0 release, remove multi-lib LWIP, clean up CMake, clean up P.IO build hardcoded defines

earlephilhower:
Doesn't seem to have many additions since the develop branch we were using, but this one is nice and official and tagged.


earlephilhower:
https://github.com/raspberrypi/pico-sdk/issues/2607

earlephilhower:
@maxgerhardt if you could look over the changes I've done in your `platform-build.py` I would really appreciate the check!

The defines set in the P.IO builds never really matched the ones in the IDE because they relied on cut-n-pasting (and editing for proper formatting) updates every time.  Now, the Arduino IDE and the P.IO should pull chip-level defines from a GCC `@` file that has some common values for all chips and special ones for the 2040/2350/riscv.

maxgerhardt:
Changes to the `platformio-build.py` are reviewed and perfectly fine. Definitely sensible to pull them out of the shared files. I also tested the latest commit inside VSCode + PlatformIO and building the firmware and uploading worked perfectly fine.

PR #3059, opened 2025-08-05
Add support for custom CMake flags and positional target arguments in build script

### LINKED ISSUE #3058: Encancement: Support Custom CMake Flags and Selecting Targets in Build Script
Right now, the `make-libpico.sh` script does not let users pass custom CMake flags or specify build targets. This limitation makes it harder to adapt to local development and more complex continuous integration (CI) workflows.
## Motivation
Sometimes, developers only need to build specific targets. For instance, I have just the RP2040 Connect board, so it would be really helpful to compile only what I need to speed up my development process.

In addition, we often require custom CMake flags for different setups. For example, I use the Picoprobe and need a multicore build, which calls for some extra flags.
## Proposal
- Add support for positional arguments to specify build targets.
- Include a `--flags` option to allow for custom CMake flags to be forwarded to CMake.
## Expected Outcome
- Backwards compatibility will be maintained.
- Users will experience improved flexibility for both local development and CI builds.

schkovich:
- Refactor `make-libpico.sh` to accept positional arguments for target selection and a `--flags` option to forward custom CMake flags.
- Update environment variable exports for clarity.
- Add `LIB_PICO_MULTICORE` CMake option to `CMakeLists.txt` and propagate it to target compile definitions.
- Preserve backwards compatibility and improve build flexibility for local and CI workflows.

### Example Usage:
```bash
/make-libpico.sh rp2040 --flags -DPICO_STDIO_UART=1 -DPICO_STDIO_USB=0 -DLIB_PICO_MULTICORE=1
```

### Tested:
- Verified that the script works as before with no arguments.
- Confirmed that custom flags are correctly passed to CMake and reflected in the build.

Closes #3058

### DIFF
--- a/tools/libpico/CMakeLists.txt
+++ b/tools/libpico/CMakeLists.txt
@@ -1,5 +1,7 @@
 cmake_minimum_required(VERSION 3.12)
 
+option(LIB_PICO_MULTICORE "Enable multicore support" 0)
+
 set(cpu $ENV{CPU})
 message("Building for CPU ${cpu}")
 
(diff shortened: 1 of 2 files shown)


schkovich:
Puff... that requires a change to pico-sdk. Let me see if I can manage to do it.

PR #3063, opened 2025-08-08
WIP - Make FreeRTOS a first-class citizen

earlephilhower:
With the expanded performance and memory of the Pico 2, having a full operating system with threads could really improve developer life and get the best possible algorithm performance.

Start by moving FreeRTOS checks from a global bool set by weak function linkage checks to a compiler definition.  This can allow the build process to differ between bare metal and FreeRTOS builds (i.e. async_context implementations) and save a few bytes of program space.


earlephilhower:
The non-CYW43 drivers look pretty good here.  We have complete control over their architecture making it much cleaner than the CYW43 driver included in the SDK.  I've had it handling ~30 HTTP server requests/second(!) and 10 Tasks(threads) of MOTD clients over a slow W5100 wired interface for >12hrs w/o a hiccup.

~The CYW32 driver in this mode is slow to connect initially but seems relatively stable.  It will probably need to be rewritten from scratch, using the CYW32 library and dropping any of the SDK driver stuff, because `async_context` and the LWIP thread just don't play nicely together.  But, it does work (ioctls are slow!) more or less.  It still thinks there is an async_context_freertos servicing things so I would not be surprised if things go boom (or take a long time) under load testing...~

If you're doing wired Ethernet and FreeRTOS and LWIP, I'd recommend giving this a try.  If you're using Platform.IO, add `-D__FREERTOS=1` to your .ini file to use this.  OTW, select the `Operating System->FreeRTOS SMP` from the IDE menus.

PR #3067, opened 2025-08-09
Support the Pimoroni Explorer board.

### LINKED ISSUE #3062: Feature Request: Support Pimoroni Pico Explorer
The Pimoroni Pico Explorer is a development board featuring with the RP2350 controller, 6 on-board buttons, 4 servo outputs, an LCD screen and a small breadboard. It would be great to control these from the Arduino environment.

libewa:
Fixes #3062


earlephilhower (approved):
Looks good at a glance.  Thanks for defining the hardwired pinouts!  Is this ready to merge?

libewa:
One more thing: Is it possible to add additional documentation, like how to use the screen with Adafruit GFX?

earlephilhower:
Normally that is done by the manufacturers at their own websites.  Adafruit, for example, has probably hundreds of examples on their site for their boards. 

Here, there's not really a good place for that kind of board specific things.  Plus it would probably need third party libraries which we're not distributing.

PR #3122, opened 2025-09-08
Fix warning for FatFS lfn size

ldursw:
`FILINFO.fname` has size 65 to accomodate a 64 character file name and the null byte. `_lfn` is defined with size 64 which could truncate the null byte if the file name is exactly 64 characters long.

```
In file included from framework-arduinopico\libraries\FatFS\src\FatFS.cpp:25:
framework-arduinopico\libraries\FatFS\src\FatFS.h: In member function 'virtual bool fatfs::FatFSDirImpl::next()':
framework-arduinopico\libraries\FatFS\src\FatFS.h:443:24: warning: 'char* strncpy(char*, const char*, size_t)' output may be truncated copying 64 bytes from a string 
of length 64 [-Wstringop-truncation]
  443 |                 strncpy(_lfn, fno.fname, sizeof(_lfn));
      |                 ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/libraries/FatFS/src/FatFS.h
+++ b/libraries/FatFS/src/FatFS.h
@@ -459,7 +459,7 @@ protected:
     FatFSImpl*              _fs;
     std::shared_ptr<DIR>    _dir;
     bool                    _valid;
-    char                    _lfn[64];
+    char                    _lfn[65];
     time_t                  _time;
     time_t                  _creation;
     std::shared_ptr<char>   _dirPath;


ldursw:
The project where I received the warning has many custom configurations and scripts. I tried to reduce it, and I think I found a minimal set that reproduces the issue. First, the project must be compiled with `-O3` optimization instead of the default `-Os`, and it must use `-Wall -Wextra`.

platformio.ini:
```ini
build_unflags = -Os
build_flags =
    -O3
    -Wall
    -Wextra
    -Wno-unused-parameter
    -Wno-type-limits
```

earlephilhower:
Got it!  I can reproduce with the ListFiles-USB FatFS example but, like you said, only with O3, and only on ARM (not R5).  I'll see if I can put something in CI for that...maybe have the Pico2 ARM builds -O3 not -Os.

PR #3126, opened 2025-09-11
Fix double promotion warnings.

martinbudden:
Fixes warnings given when code is compiled with  `-Wdouble-promotion`


earlephilhower:
Thanks!

PR #3154, opened 2025-09-25
Reduce LWIP flash 5.4K by cleaning debug prints

earlephilhower:
In non-debug cases the string to `panic()` isn't actually sent to the Serial port, or anywhere in fact.

When there is no debug port defined, replace the panic(str) of LWIP_PLATFORM_ASSERT with panic("lwip") instead.  This doesn't effect debugging because the source line of the LWIP_ASSERT will still show the full string (GDB shows the source).

Also, LWIP_DEBUG is a flag by definition, not value, so fix lwipopts.h to not define it or the LWIP_X debug defines unless LWIP_DEBUG is set by the compliation.

Saves 5480 bytes of flash for LWIP (Pico W, 2W, Ethernet) apps.

### DIFF
--- a/tests/build-rp2350.sh
+++ b/tests/build-rp2350.sh
@@ -15,7 +15,7 @@ elif [ "$BUILD_PARITY" = "odd" ]; then
     rem=1
 fi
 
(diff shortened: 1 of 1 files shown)


earlephilhower:
Original CI failure seems to show a GCC bug for the M3 (Pico2) where it thinks the DHCP `options[]` is 0-length, while it is always set to the min (`68U`) (verified with `#if DHCP_OPTIONS_LEN < 68 / #error` manually added in my own testing.  The M0+ (Pico) doesn't show the error even at `-O3`, so I have to conclude the ARM-M3 GCC has a bug with -O3...

Issue #3173, opened 2025-10-02
HTTPClient doesn't preserve request headers over multiple redirects


skaringa:
I'm using the following code to make a HTTP request:

      http.begin(url);
      http.setReuse(true);
      http.setFollowRedirects(HTTPC_FORCE_FOLLOW_REDIRECTS);
      http.addHeader("Icy-MetaData", "1");
      int code = http.GET();
      if (code != HTTP_CODE_OK) {
          // ...

On the first request to *url*, the headers added with *addHeader* are correctly sent to the target host.
However, if one or more redirects happen during the *GET* method, then those headers are not sent again to the redirected URLs.

I would expect to re-send the headers over multiple redirects, because the description of the redirection modes states clearly:

> `HTTPC_FORCE_FOLLOW_REDIRECTS` - all redirections will be followed,
        regardless of a used method. New request will use the same method,
        and they **will include the same body data and the same headers**.
        In the sense of the RFC, it's just like every redirection is confirmed.

The reason for the wrong behavior is that the method *HTTPClient::clear* is called multiple times during handling of the request, but it clears all headers.

I fixed this locally with the following patch:

    diff -p HTTPClient.cpp ~/.arduino15/packages/rp2040/hardware/rp2040/5.1.0/libraries/HTTPClient/src/HTTPClient.cpp 
    *** HTTPClient.cpp	2025-10-02 17:58:12.619006816 +0200
    --- ~.arduino15/packages/rp2040/hardware/rp2040/5.1.0/libraries/HTTPClient/src/HTTPClient.cpp	2025-10-02 18:07:18.688822906 +0200
    *************** public:
    *** 104,109 ****
    --- 104,110 ----
      
      
      void HTTPClient::clear() {
    +       DEBUG_HTTPCLIENT("[HTTP-Client][clear]\n");
          _returnCode = 0;
          _size = -1;
          _headers = "";
    *************** int HTTPClient::sendRequest(const char *
    *** 578,591 ****
          int code;
          bool redirect = false;
          uint16_t redirectCount = 0;
          do {
    !         // wipe out any existing headers from previous request
              for (size_t i = 0; i < _headerKeysCount; i++) {
                  if (_currentHeaders[i].value.length() > 0) {
                      _currentHeaders[i].value = "";
                  }
              }
      
              DEBUG_HTTPCLIENT("[HTTP-Client][sendRequest] type: '%s' redirCount: %d\n", type, redirectCount);
      
              // connect to server
[...]

PR #3175, opened 2025-10-02
HTTPClient preserves headers over redirects (#3173)

skaringa:
HTTP headers are preserved over redirects.
Additional re-directed requests will include the same headers like the original request.

Fixes issue #3173


libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -578,14 +578,23 @@ int HTTPClient::sendRequest(const char * type, const uint8_t * payload, size_t s
     int code;
     bool redirect = false;
     uint16_t redirectCount = 0;
+    String headers;
```
earlephilhower, line 581: Is there something missing here?  I see you make a local `String headers` but it is never set to anything so just blank `""`.

libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -578,14 +578,23 @@ int HTTPClient::sendRequest(const char * type, const uint8_t * payload, size_t s
...
+        // wipe out any existing response headers from previous request
         for (size_t i = 0; i < _headerKeysCount; i++) {
             if (_currentHeaders[i].value.length() > 0) {
                 _currentHeaders[i].value = "";
             }
         }
 
+        // keep request headers for redirects
+        // fixes #3173
+        if (redirect) {
```
earlephilhower, line 592: And here, it will clear `this->_headers` (since `String headers` is never set to anything

libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -578,14 +578,23 @@ int HTTPClient::sendRequest(const char * type, const uint8_t * payload, size_t s
...
+            _headers = headers;
+        } else {
+            headers = _headers;
```
earlephilhower, line 595: And here it will set the local variable to the `this->_headers` but never use it...

earlephilhower (changes requested):
Looks like the log here will just clear the `_headers` member variable on a `redirect`, but I thought you were trying to save those headers.  Maybe there's more to the diff?

libraries/HTTPClient/src/HTTPClient.cpp
skaringa: `headers` is set later during the first pass of the do..while loop.

earlephilhower (approved):
Thanks for the explanation.  I missed the fact this was in a do..while.  `HTTPClient` is one of the classes ported from the ESP8266 and has had a lot of authors and some interesting logic.

Anyway, LGTM now that I've got your explanation and a better look at the code!

PR #3225, opened 2025-10-27
[UART] Saves the achieved uart baud rate and adds SerialUART::getAcheivedBaud()

ngist:
The pico SDK uart_init() makes it's best attempt to achieve the the baud rate specified, but it is by no means guaranteed to do so. It does however helpfully compute the nominal baud rate it actually achieved. 

See:
https://github.com/raspberrypi/pico-sdk/blob/a1438dff1d38bd9c65dbd693f0e5db4b9ae91779/src/rp2_common/hardware_uart/include/hardware/uart.h#L273

As currently implemented in this the achieved baud rate value is discarded. ~~This change simply returns this value from begin functions so that users of it can extract the result if desired.~~ This change saves the result and adds a getter method to allow higher level access to the result. This allows things like throwing a warning or error if the difference exceeds some bounds, or simply checking the achieved value while debugging.


earlephilhower:
It looks like this not matching the Arduino "standard" is fatal.  The virtual methods have fixed signatures in `HardwareSerial.h` and changing return types isn't really an overload that can be done.
````
In file included from /home/runner/arduino_ide/hardware/pico/rp2040/cores/rp2040/Arduino.h:138,
                 from /home/runner/work/arduino-pico/arduino-pico/build.tmp/sketch/DigitalReadSerial.ino.cpp:1:
/home/runner/arduino_ide/hardware/pico/rp2040/cores/rp2040/SerialUART.h:67:9: error: conflicting return type specified for 'virtual int SerialUART::begin(long unsigned int)'
   67 |     int begin(unsigned long baud = 115200) override {
      |         ^~~~~
In file included from /home/runner/arduino_ide/hardware/pico/rp2040/cores/rp2040/api/HardwareSerial.h:2,
                 from /home/runner/arduino_ide/hardware/pico/rp2040/cores/rp2040/SerialUSB.h:24,
                 from /home/runner/arduino_ide/hardware/pico/rp2040/cores/rp2040/Arduino.h:135:
/home/runner/arduino_ide/hardware/pico/rp2040/cores/rp2040/api/../../../ArduinoCore-API/api/HardwareSerial.h:91:18: note: overridden function is 'virtual void arduino::HardwareSerial::begin(long unsigned int)'
   91 |     virtual void begin(unsigned long) = 0;
      |                  ^~~~~
/home/runner/arduino_ide/hardware/pico/rp2040/cores/rp2040/SerialUART.h:70:9: error: conflicting return type specified for 'virtual int SerialUART::begin(long unsigned int, uint16_t)'
   70 |     int begin(unsigned long baud, uint16_t config) override;
      |         ^~~~~
/home/runner/arduino_ide/hardware/pico/rp2040/cores/rp2040/api/../../../ArduinoCore-API/api/HardwareSerial.h:92:18: note: overridden function is 'virtual void arduino::HardwareSerial::begin(long unsigned int, uint16_t)'
   92 |     virtual void begin(unsigned long baudrate, uint16_t config) = 0;
      |                  ^~~~~
exit status 1
````

earlephilhower:
One option would be to store the uart_init speed in class variable add a non-virtual class method like `int getRealBaud() { return _realBps;`

ngist:
That was my initial inclination but I thought this would be simpler. I’ll
post the change shortly. Thanks

ngist:
OK updated with the getter method. Local CI died along with my laptop battery so unfortunately i didn't pretest this.

Issue #3157, opened 2025-09-28
Second call to WiFi.begin() + ArduinoOTA running seems to block forever on RP2040


ledereyes:
Hi, I am trying to implement proper WiFi reconnect handling, but it seems the second time I am calling WiFi.begin it blocks forever. This is my logic called from loop:

```
  if (!WiFi.isConnected())
  {
    Serial.println("connection lost");
    if (WiFi.begin(ssid, pass) != WL_CONNECTED) // seems to block forever
    {
      Serial.println("wifi reconnect failed");
    }
    else
    {
      Serial.println("wifi reconnect success");
    }
  }
```

I am triggering the disconnect by powering off and on my router. The last message I get over Serial is "connection lost" which makes me believe that WiFi.being blocks.
What I tried so far:
1. Using the async version `WiFi.beginNoBlock` but it also never returns.
2. Adding various delays here and there.
3. Calling `WiFi.disconnect` and `WiFi.end` before the reconnect.
4. An RP2350

These are my versions:
```
[env:rpipicow]
platform = https://github.com/maxgerhardt/platform-raspberrypi.git#a4b555a5f29961fd76e278b3c5ba56f0067207d0
framework = arduino
board_build.core = earlephilhower
board_build.filesystem_size = 0.5m
board = rpipicow
```

Thanks for any help on the issue!

earlephilhower:
I can't reproduce any such failure here.  I set up another PicoW as the server (using the CaptivePortal example and renaming the WiFi network to "picow" because I didn't want to have to type that loooong name it used) and ran the following test which does what you've described above:
````
#include <WiFi.h>

void setup() {
  delay(5000);
  WiFi.setHostname("DUT");
  Serial.println("Starting WiFi");
  WiFi.begin("picow");
  Serial.printf("IP: %s\n", WiFi.localIP().toString().c_str());
}

void loop() {
  static int l = 0;
  Serial.printf("loop %d\n", l);
  if (!WiFi.connected()) {
    Serial.println("Attempting reconnect...");
    Serial.flush();
    if (WiFi.begin("picow") == WL_CONNECTED) {
      Serial.printf("IP: %s\n", WiFi.localIP().toString().c_str());
    } else {
      Serial.println("Failed\n");
    }
  }
  delay(50); // do something;
  l++;
}
````

[...]

PR #3250, opened 2025-11-27
Disable IRQs on RP2040 during HW division

earlephilhower:
The divider is a shared HW resource, and when an IRQ comes in (i.e. when a packet is processed by LWIP and the user's callbacks) its state can be corrupted silently and randomly.

Change the Pico-SDK defaults to disable IRQs during division operations, avoiding the issue by disallowing the LWIP callback to happen until after division is completed.

Fixes #3212


yohine:
I found the real cause. The 64-bit timer loop is unrelated to the WiFi library. There is nothing wrong with Arduino library and SDK. The WiFi reconnection issue is resolved in #3212. 

Sorry, I made it extra work for you here.

yohine:
I'm currently running stability tests on my environment, applying 5.4.3 and update #3250. Operation under the same conditions as the previous task appears to be fine for now. It's too early to draw any conclusions about long-term stability, but I'll report any issues later.

I also performed another test: the stability of the WiFi library in an RP2040 multi-core environment. 

In projects of a complexity involving SPI and I2C access, multi-core and WiFi are extremely unstable, even without an OS. I have not been able to get WiFi to work on multi-core with update #3250. 

However, even if it can't be resolved, it's not a problem. When I started my project, the library was still version 5.2.0 or earlier, and even WiFi.beginNoBlock would block for several seconds. Therefore, I considered separating WiFi initialization on multi-core to avoid blocking the main task. 

The blocking issue was resolved with an update to #3157 CYW43shim.cpp, and now begin itself does not block. Therefore, multi-core operation is no longer required.

earlephilhower:
Per the RPI guys, this is not needed.  The divider_hardware.S file will save/restore before/after any division operation, so IRQ and app-level are fine.

PR #3257, opened 2025-12-07
Adjust SerialPIO flush timeouts

### LINKED ISSUE #3256: PIO Serial.flush - seems to take too much time
Hello!

Using ArduinoIDE for RP2040 programming. Use serial transmission (PIOSerial, 115k) for sending data, with code:

```
(...)
int SerialBufferSize=512;
SerialPIO PIOSerial3 (10, 11, SerialBufferSize);
(...)
HardwareSerial* port=nullptr;
port = &PIOSerial3; 
(...)

gpio_put(DIR_PIN, true); //start sending

port->write(txFrameBuffer, totalLen); //sending

gpio_put(DIR_PIN, false); //sending end
```

Of course it fails, because “sending end” information raise to early - DIR_PIN going low during transmission (upper: data frame, lower DIR_PIN state):

Image

So, lets use Serial.flush() which check when output buffer will be empty (all data sent):

```
gpio_put(DIR_PIN, true);

port->write(txFrameBuffer, totalLen);

port->flush();  

gpio_put(DIR_PIN, false);
```

Much better, but.. why Serial.flush() tooks so much time longer? All Serial goes out, but DIR_PIN is HIGH for over 170us after real transmission ends?

Image

Lets do other test - change serial speed to 9600:

Image

The same situation, too long “tail” but seems almost fine

So, speed up to 961k:

Image

Seems Serial.flush() takes about 10x more time than real transmission time! As I see, Serial.flush() executes not less than ~1ms (?)

My questions are:

- Does Serial.flush() should works like that?
- How proper check when serial transmission ends?

Best regards!

earlephilhower:
Fixes #3256

Reduce the minimum timeout for SerialPIO::flush from 1ms to 1us.

### DIFF
--- a/cores/rp2040/SerialPIO.cpp
+++ b/cores/rp2040/SerialPIO.cpp
@@ -355,10 +355,10 @@ void SerialPIO::flush() {
         return;
     }
     while (!pio_sm_is_tx_fifo_empty(_txPIO, _txSM)) {
-        delay(1); // Wait for all FIFO to be read
+        /* noop */ // Busy wait for all FIFO to be read
     }
     // Could have 1 byte being transmitted, so wait for bit times
-    delay((1000 * (_txBits + 1)) / _baud);
+    delayMicroseconds((1000000 * (_txBits + 3 /* start + stop + parity */)) / _baud);
 }
 
 size_t SerialPIO::write(uint8_t c) {


earlephilhower:
@zmechu, please give this PR a try with your o'scope and report back.  It should reduce the excessive serialpio.flush times significantly.

PR #3288, opened 2026-01-05
Attempt at adding support for Waveshare RP2350B-Plus-W

Lannakin:
I've done my best to follow the guide for contributing and:

Edited `makeboards.py`, although I am not sure if I understood what these parameters should be:

```python
    elif name == "waveshare_rp2350_plus_w":
        BuildFlashMenu(name, chip, 4*1024*1024, [*smallfs, 1024*1024, 2*1024*1024, 3*1024*1024])
        BuildFlashMenu(name, chip, 16*1024*1024, [0, 15*1024*1024, 14*1024*1024, 12*1024*1024, 
```

(I do not know if the first BuildFlashMenu is correct to be included)
I also was unsure of what parameters 6, 7, and 10 should be, as the number I got for the device ID via `udevadm` attributes walk is quite different from the others' IDs (but the manufacturer code matched so).
```python
MakeBoard("waveshare_rp2350_plus_w", "rp2350", "Waveshare", "RP2350 Plus W", "0x2e8a", "0x0009", 500, "WAVESHARE_RP2350_PLUS_W", 16, 0, "none", ["PICO_CYW43_SUPPORTED=1", "CYW43_PIN_WL_DYNAMIC=1"])
```

I guess the 7th parameter is power in mA?  The schematic says the LDO is `ME6217C33M5G`, and the datasheet for that says it can output 800mA (this LDO is also used on the Waveshare RP2350B-Eth and RP2350B-Tiny), so I guess I'll update that and rerun the script.

I also added `digital.cpp`, `pins_arduino.h`, and `init.cpp`.  The comments at the top of `digital.cpp` and `pins_arduino.h` include the datasheets I used as references.

### DIFF
--- a/variants/waveshare_rp2350b_plus_w/pins_arduino.h
+++ b/variants/waveshare_rp2350b_plus_w/pins_arduino.h
@@ -67,8 +67,7 @@
 
 // LED_BUILTIN
 #define LED_BUILTIN     PIN_LED
-#define PIN_LED         (64u) // LED1 uses the Raspberry Pi Radio Module 2
-                              // "B1_GPIO0"
+#define PIN_LED         (64u) // LED1 uses the Raspberry Pi Radio Module 2 ("B1_GPIO0")
 
 // Serial
 #define PIN_SERIAL1_TX  (0u)


Lannakin:
Note:  updated the `pwr` parameter in `makeboards.py` and reran the script.

tools/makeboards.py
```diff
@@ -725,6 +728,7 @@ def MakeBoardJSON(name, chip, vendor_name, product_name, vid, pid, pwr, boarddef
 MakeBoard("waveshare_rp2350_zero", "rp2350", "Waveshare", "RP2350 Zero", "0x2e8a", "0x10B0", 500, "WAVESHARE_RP2350_ZERO", 4, 0, "none")
 MakeBoard("waveshare_rp2350_pizero", "rp2350", "Waveshare", "RP2350 PiZero", "0x2e8a", "0x000F", 500, "WAVESHARE_RP2350_PIZERO", 16, 0, "none")
 MakeBoard("waveshare_rp2350_plus", "rp2350", "Waveshare", "RP2350 Plus", "0x2e8a", "0x10B1", 500, "WAVESHARE_RP2350_PLUS", 4, 0, "none")
+MakeBoard("waveshare_rp2350b_plus_w", "rp2350", "Waveshare", "RP2350B Plus W", "0x2e8a", "0x0009", 800, "WAVESHARE_RP2350B_PLUS_W", 16, 0, "none", ["PICO_CYW43_SUPPORTED=1", "CYW43_PIN_WL_DYNAMIC=1"])
```
earlephilhower: > I guess the 7th parameter is power in mA? The schematic says the LDO is ME6217C33M5G, and the datasheet for that says it can output 800mA (this LDO is also used on the Waveshare RP2350B-Eth and RP2350B-Tiny), so I guess I'll update that and rerun the script.

On reflection, I think you may want to leave the pwr to 500mA, not 800.

This is the power limit sent in the USB descriptor and not really the upper limit of the onboard regulators.  If the nominal power really is 0.8A(!), then leave it...if this board has a bunch of WS2812s or something.  But OTW putting this higher than the more typical 250 or 500 could cause USB hubs and some onboard ports to disable it because they think it would go over their power budget.

> I also was unsure of what parameters 6, 7, and 10 should be, as the number I got for the device ID via udevadm attributes walk is quite different from the others' IDs (but the manufacturer code matched so).

6 is the USB PID.  If Waveshare has one they use, it can go here.  OTW, just pick something like you did is fine.  It's all SW defined so you can set it as you like

10 is PSRAM size.  If there's 8MB of PSRAM onboard, set it to 8 and define the PSRAM_CS pin like in the Adafruit_Fruitjam pins_arduino (and others).

PR #3294, opened 2026-01-08
The SPISlave library was modified in order to make possible its use in a FreeRTOS app.

tesanoff:
Hello.

I tried to use SPISlave in my FreeRTOS-based project but I found out that the library is not quite "FreeRTOS-ready" yet.

This is mostly about its callbacks. When a user calls a typical `*FromISR()` function, it's necessary to call `portYIELD()` depending on the value the `*FromISR()` function returns via the last parameter.

The problem with the current implementation of SPISlave is that there's no proper place for the user to make that call of `portYIELD()`.

The proposed change adds another parameter to both `SPISlaveRecvHandler` and `SPISlaveSentHandler`, so that the user can just use that parameter each time when a FreeRTOS-API function is called. Here's an example:

```
void    KbdMessenger::sentCallback(BaseType_t *hptw){
    // this callback must not work if the instance is not fully initialized yet.
    if(!_is_initialized){
        return;
    }

    // Here we set the "ready for sending" flag
    xEventGroupSetBitsFromISR(_flags, F_READY_FOR_SENDING, hptw);
}
```

If the user uses this `hptw` parameter that way, the modified SPISlave library will take care about the rest and it will call `portYIELD()` if required.

This modification was tested and flawlessly works in my application.

If FreeRTOS is not used, then SPISlave remains exactly the same, without any extra parameters in callback handlers.

### DIFF
--- a/libraries/SPISlave/src/SPISlave.cpp
+++ b/libraries/SPISlave/src/SPISlave.cpp
@@ -191,7 +191,7 @@ void SPISlaveClass::_handleIRQ() {
         }
     } while (spi_is_writable(_spi) & _dataLeft);
 #ifdef __FREERTOS
-    portYIELD_FROM_ISR(xHigherPriorityTaskWoken);    
+    portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
 #endif
     // Disable the TX FIFO IRQ if there is still no data to send or we'd always be stuck in an IRQ
     // Will be re-enabled once user does a setData


earlephilhower:
I'm kind of surprised you can get reliable SPISlave under FreeRTOS in any case given the task swapping additional random latency spikes in the app.

In an ideal world, I'd like to have a more "FreeRTOS native" organization than this.  Like on the Ethernet drivers, where we get the IRQ and then just signal a task which does the real work at Task_t level and doesn't need to jump through hoops or operate at IRQ level.

But I can see here, the latency concerns really mean your solution is the best pragmatic one.  So, thanks!

tesanoff:
Thanks!

PR #3302, opened 2026-01-10
Add include guard to EthernetLinkStatus enum to prevent multiple definitions

RainerMinixhofer:
When both Ethernet and WiFi libraries are included, the EthernetLinkStatus enum
is defined in both Ethernet.h and LwipIntfDev.h, causing a compilation error.
This guard ensures the enum is defined only once.

### DIFF
--- a/libraries/lwIP_Ethernet/src/LwipIntfDev.h
+++ b/libraries/lwIP_Ethernet/src/LwipIntfDev.h
@@ -66,12 +66,14 @@ static void _cyw43_hal_generate_laa_mac(__unused int idx, uint8_t buf[6]) {
 }
 
 
-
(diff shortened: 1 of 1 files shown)


earlephilhower:
Thanks, seems fine.

Are you actually using wired and Wireless Ethernet at the same time?  I'm not sure if LWIP and the backend here play nicely with multiple interfaces.  I haven't added anything to stop it working, but I've not actually tested or debugged multiple Ethernet devices...

RainerMinixhofer:
Just trying it out. I'm planning to use WiFi only as fallback when no Ethernet is available, but this was prevented by the compile error on the enum...

Issue #3308, opened 2026-01-16
Add symbol definition to check the filesystem size in sketch


jlbirccyn:
Hello,

I needed to check the correct flash partitioning is selected in the Tools menu so I added ```-DFILESYSTEM_START``` and ```-DFILESYSTEM_END``` in the ```compiler.defines``` in platform.txt:

```
compiler.defines=-DFILESYSTEM_START={build.fs_start} -DFILESYSTEM_END={build.fs_end} {build.os} ...
```

So that I can do something like that in a sketch:

```C
#if FILESYSTEM_END - FILESYSTEM_START != 1048576
#error "select '2MB (Sketch: 1MB, FS: 1MB)' in 'Tools > Flash size' menu"
#endif
```

Do you want a pull request for this, maybe with different symbol names ?

Best regards and happy new year for 2026

earlephilhower:
I honestly don't remember the reasoning for making the FS location be a linker-created symbol vs. using a `-D` on the build command line.  It was inherited from the [ESP8266 core](https://github.com/esp8266/Arduino) I (and tons of others, of course) worked on, and was set up before I started contributing there so I'm oblivious as to the origin.  My own [monkey ladder experiment](https://factschology.com/factschology-articles-podcast/monkey-ladder-experiment-truth) reproduction.  🐒 🪜 😆 

I'm fine if you'd like to do a PR here.  Maybe shorten to FS_START/FS_END just because I'm a lazy typist, but adding to the defines seems benign...

jlbirccyn:
Fine, I'll do that.

PR #3310, opened 2026-01-17
Added symbols FS_START and FS_END in compiler.defines

jlbirccyn:
These symbols allow you to verify during compilation that the appropriate memory partitioning has been selected in the Tools > Flash Size menu for example.

See issue #3308

### DIFF
--- a/tools/platformio-build.py
+++ b/tools/platformio-build.py
@@ -132,7 +132,9 @@ env.Append(
         ("BOARD_NAME", '\\"%s\\"' % env.subst("$BOARD")),
         # at this point, the main.py builder script hasn't updated upload.maximum_size yet,
         # so it's the original value for the full flash.
-        ("PICO_FLASH_SIZE_BYTES", board.get("upload.maximum_size"))
+        ("PICO_FLASH_SIZE_BYTES", board.get("upload.maximum_size")),
+        ("FS_START", "$FS_START"),
+        ("FS_END", "$FS_END")
     ]
 )
 


maxgerhardt:
Can you equalize this with PlatformIO's build logic, too? Otherwise the Arduino IDE and PlatformIO will build it differently

https://github.com/earlephilhower/arduino-pico/blob/8f7a2be50143c323a8375deba5ecfe6362be98a1/tools/platformio-build.py#L127-L137

would need these two new entries

```py
        ("FS_START", "$FS_START"),
        ("FS_END", "$FS_END"),
```

plus the comma in the (previously) last entry of the list.

That env variable is exposed in https://github.com/maxgerhardt/platform-raspberrypi/blob/cc24cfef37ed22ca9f2a6aead28c2deb76c39f24/builder/main.py#L93-L94

jlbirccyn:
Of course. Just did it (I hope it is ok)

maxgerhardt:
Looks perfect, will retest on hardware later.

earlephilhower (approved):
Thanks, LGTM.  I'll wait for @maxgerhardt to update on his testing and then merge.

maxgerhardt:
Tested in PlatformIO, works perfectly.

Did
```ini
[env:rpipico]
platform = raspberrypi
framework = arduino
board = rpipico
; allocate 512k for LittleFS
board_build.filesystem_size = 0.5m
```
And checked

```cpp
#if FS_END - FS_START != (1024 * 1024)
#error "Wrong FS size!"
#endif
```
triggers the error, 
```cpp
#if FS_END - FS_START != (512 * 1024)
#error "Wrong FS size!"
#endif
```
does not, the macro values are correctly exposed.

grafik

Good to go!

PR #3313, opened 2026-01-19
[Feature] Add Waveshare RP2040 LoRa variant board

Lannakin:
```txt
added Waveshare RP2040 LoRa to README.md
added Waveshare RP2040 to makeboards.py
added ./variants/waveshare_rp2040_lora/pins_arduino.h

ran ./tools/makeboards.py
```

Not to brag, but I still have a lot to learn about Git'ing 💀 I initially pushed everything onto my main branch.


earlephilhower:
Thanks!  It takes a bit to get into the `git` mindset, but everyone goes through the startup pains!

https://xkcd.com/1597/

Lannakin:
Oh no, that comic is exactly what I've been doing!  If I'm unsure of how to phrase phrase a query, I hesitate for about 15 seconds deciding whether I want to learn to do something properly...and then just delete the repo.  Sometimes I'll even do this to avoid making multiple commits (as they reveal ~~the madness within~~ ~~my numerous flaws~~ I don't read good).

PR #3344, opened 2026-02-01
Async SPI Callbacks

earlynerd:
Adds callback support for async SPI transfers via onTransferComplete()

users can register a callback that fires automatically when the SPI transfer completes
Uses the RX DMA channel completion interrupt as the trigger, since SPI is full-duplex and RX completes exactly when the last bit clocks out

Follows existing patterns from AudioBufferManager and Wire for DMA IRQ handling
 API

  // Simple callback
  SPI.onTransferComplete(myCallback);

  // Callback with user context
  SPI.onTransferComplete(myCallback, contextPtr);

  // Disable callback
  SPI.onTransferComplete(nullptr);

  - DMA IRQ is always enabled for async transfers; internal handler performs cleanup
  - finishedAsync() now just checks a flag (cleanup moved to IRQ handler)
  - Polling mode still works - just don't register a callback

### DIFF
--- a/libraries/SPI/examples/SPItoMyselfAsync/SPItoMyselfAsync.ino
+++ b/libraries/SPI/examples/SPItoMyselfAsync/SPItoMyselfAsync.ino
@@ -2,9 +2,7 @@
 // Core0 runs as an SPI master and initiates a transmission to the slave
 // Core1 runs the SPI Slave mode and provides a unique reply to messages from the master
 //
-// This example demonstrates two modes:
-// 1. Callback-based async transfer (preferred) - uses onTransferComplete()
-// 2. Polling-based async transfer (legacy) - uses finishedAsync()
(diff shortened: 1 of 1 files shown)


libraries/SPI/src/SPI.h
```diff
@@ -75,8 +75,9 @@ class SPIClassRP2040 : public arduino::HardwareSPI {
     */
     void transfer(const void *txbuf, void *rxbuf, size_t count) override;
 
-    // DMA/asynchronous transfers.  Do not combime with synchronous runs or bad stuff will happen
+    // DMA/asynchronous transfers.  Do not combine with synchronous runs or bad stuff will happen.
```
earlephilhower: Huh, good catch.  I wonder why this wasn't seen by the `codespell` check.

PR #3372, opened 2026-02-11
lwip_wrap.h: wrap more functions

functionpointer:
While debugging https://github.com/earlephilhower/arduino-pico/pull/3265 i have found several LWIP functions that were called without the mutex being held. This was found by implementing `LWIP_ASSERT_CORE_LOCKED` with `async_context_lock_check(__getEthernetContext());`, as seen here: https://github.com/earlephilhower/arduino-pico/pull/3408

This PR wraps these functions to ensure integrity of LWIP.

Additionally, it adds a comment about async context behaviour that explains some behavior seen in #1973, #1864 and #1979

### DIFF
--- a/cores/rp2040/lwip_wrap.cpp
+++ b/cores/rp2040/lwip_wrap.cpp
@@ -1038,43 +1038,43 @@ extern "C" {
 
     extern void __real_netif_set_link_up(struct netif *netif);
     void __wrap_netif_set_link_up(struct netif *netif) {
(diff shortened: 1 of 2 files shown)


functionpointer:
> Just out of curiosity, how did you actually write the code for the wrappers? I couldn't get a reliable output from whatever GH copilot was available at the time I implemented them so just had to slog thru the whole list manually...

I just did it manually as well

PR #3374, opened 2026-02-11
Add WIFI AP+STA Support

ayushsharma82:
@earlephilhower  This PR adds WiFi AP+STA support in arduino-pico SDK along with `WIFIATSTA.ino` example to test it.

<ins>Real world test done:</ins>
- Pico connected to STA
- Connected 2 phones to Pico's AP

**Result:** Success (all 3 connections are stable and work concurrently)

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ Read the [Contributing Guide](https://github.com/earlephilhower/arduino-pico/blo
 * Olimex Pico2XL
 * Olimex Pico2XXL
 * Olimex RP2040-Pico30
+* PicoLume Transceiver
 * Pimoroni PGA2040
 * Pimoroni Pico Plus 2
 * Pimoroni Pico Plus 2W
--- a/boards.txt
+++ b/boards.txt
@@ -24505,6 +24505,252 @@ olimex_rp2040pico30.menu.uploadmethod.picoprobe_cmsis_dap.upload.tool=picoprobe_
 olimex_rp2040pico30.menu.uploadmethod.picoprobe_cmsis_dap.upload.tool.default=picoprobe_cmsis_dap
 
 # -----------------------------------
+# PicoLume Transceiver
+# -----------------------------------
+picolume.name=PicoLume Transceiver
+picolume.vid.0=0x1209
+picolume.pid.0=0x2008
+picolume.vid.1=0x1209
+picolume.pid.1=0x2108
+picolume.vid.2=0x1209
+picolume.pid.2=0x6008
+picolume.vid.3=0x1209
+picolume.pid.3=0x6108
+picolume.vid.4=0x1209
+picolume.pid.4=0xa008
+picolume.vid.5=0x1209
+picolume.pid.5=0xa108
+picolume.vid.6=0x1209
+picolume.pid.6=0xe008
+picolume.vid.7=0x1209
+picolume.pid.7=0xe108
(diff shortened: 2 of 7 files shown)


ayushsharma82:
Ready to merge captain.

earlephilhower:
Very, very nice! :rocket: 

I could swear I tried it when the PicoW first came out and when the other virtual channel (what the driver uses to talk to the CYW34 firmware) was opened, the first one was closed.  And at that point the driver and upper layer stack got very upset.  So I hardcoded it on or off and didn't touch it...

Give me a day or two to look this over, it's kind of busy here and not much coding time.

ayushsharma82:
@earlephilhower  I tested this PR with NetWizard, although the core functionality is working again - there are a few minor quirks. For example WiFiClass `disconnect` and other small functions are not behaving same as ESP32. Upon calling `disconnect` the WiFiClass should only disconnect STA in WIFI_AP_STA mode...

I'll look into it more and add some commits today.

ayushsharma82:
@earlephilhower Shall I split the WiFiClass into WiFiSTA and WiFiAP as seen in arduino-esp32 SDK? That pretty much makes more sense and has better separation of logic.

Ref: https://github.com/espressif/arduino-esp32/tree/4e42521cf089e6dac9c9ff472ff15cc4be5b48bf/libraries/WiFi/src

earlephilhower:
If it doesn't break compatibility and it doesn't turn into a massive mess, that would be fine.  I have a feeling it would be kind of a pain, but it's been a while since I wrote the stuff so maybe I'm just scared of opening up something that works. :laughing: 

FWIW, the WiFiClass here was based on the Arduino 'WiFi' API with ESP compatibility hacks so it may not be as simple as separating things and then moving WiFi STA/AP to WiFiSTA/AP and have a dummy top-level class inherit both.

ayushsharma82:
Looks like [Arduino WiFi API](https://github.com/arduino-libraries/WiFi/blob/master/src/WiFi.h) was never intended to even have an "AP" mode. ESP32 style API is far more superior in that case.

I can maintain backwards compatibility with existing functions but there is one thing - All the 'AP' related code has to go into their own "softAPXXXX" functions (just like ESP32 WiFi API) for better separation at some point in future. If I go around doing it, I'll open a separate PR which you can merge when you are ready to add breaking changes.

ayushsharma82:
Done with the changes and ready to merge.

I think this PR has reached a point now which should be good enough for arduino-pico v5. Whenever its time to add breaking changes, I'll suggest copying the ESP32 WiFiSTA and WiFiAP classes and porting CYW43 driver to it. ESP8266 platform is end-of-life and we should deprecate the old logic and move arduino-pico's WiFi API to use latest and proven ESP32 WiFi APIs.

Also tested with NetWizard, everything works flawlessly 🚀 .

earlephilhower:
Awesome.  This past week was crazy hectic and I didn't get a chance to look at this.  I'll pencil it some time this coming weekend to give it a once-over and see that it doesn't break any of the existing examples (probably not, but it's easy to test!).

ayushsharma82:
@earlephilhower No worries. As a solo-maintainer myself of multiple repositories I can understand the situation. Let me know if anything comes up in the testing this weekend.

earlephilhower:
I'm traveling this week for a new job so apologies, again.  It's regularly pinging me so I won't forget when I get home.

PR #3377, opened 2026-02-13
ClientContext.h: Prevent double free by checking in wrapper

functionpointer:
Adds NULL checks in lwip_wrap.cpp to prevent double frees in ClientContext.h

See https://github.com/earlephilhower/arduino-pico/pull/3265 and https://github.com/earlephilhower/arduino-pico/pull/3368
It is an alternative to https://github.com/earlephilhower/arduino-pico/pull/3370

As mentioned in https://github.com/earlephilhower/arduino-pico/pull/3368#issuecomment-3897446991, this is currently incomplete as it does not all accesses of pcb go through the wrapper.

### DIFF
--- a/cores/rp2040/freertos/freertos-lwip.cpp
+++ b/cores/rp2040/freertos/freertos-lwip.cpp
@@ -150,7 +150,9 @@ static void lwipThread(void *params) {
             }
             case __tcp_arg: {
                 __tcp_arg_req *r = (__tcp_arg_req *)w.req;
-                __real_tcp_arg(r->pcb, r->arg);
+				if (r->pcb != NULL) {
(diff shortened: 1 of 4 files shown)


functionpointer:
Maybe we can deal with the direct memory accesses by simply disabling interrupts.
After all, these accesses are fast and can not block. What do you think?

The naive way to implement involves ifdefs in the affected functions, which is ugly and error prone.
So I came up with a cleaner way, by extending LWIPMutex.
If you like it I will fix the errors and test it.

Finally, `getRemoteAddress` and `getLocalAddress` still need a solution.
Most usages immediately make a new `IPAddress` object with the data.
I think we could move that step into ClientContext, where it can happen under the protection of LWIPMutex/noInterrupts.

PR #3397, opened 2026-03-08
Error on BLE builds when BT not enabled

earlephilhower:


earlephilhower:
GH CI is having a bad day, the failures are VMs/containers being killed in mid run.

Issue #2528, opened 2024-10-07
Feature Request: support deep sleep for the RP2350


sillycowvalley:
I wouldn't know where to start so I asked the One Who Is Wrong With Confidence (ChatGPT) for an example and got the example below as a starting point. Would we want to add some RISC-V specific functionality to the RP2040 class for this?

---

Here's an example of how to put the Pico 2 into deep sleep and wake it up using a GPIO pin (You can also use the RTC to wake it up after a specific time interval):

### C++ Example for Deep Sleep on RP2350 (Pico 2)

```cpp
#include "pico/stdlib.h"
#include "hardware/gpio.h"
#include "hardware/clocks.h"
#include "hardware/vreg.h"
#include "hardware/regs/io_bank0.h"

// Set the GPIO pin to wake the system
#define WAKE_PIN 15

void setup_wake_pin() {
    // Enable the wake pin
    gpio_set_dir(WAKE_PIN, GPIO_IN);
    gpio_pull_down(WAKE_PIN);

    // Enable wakeup by pin
    gpio_set_irq_enabled(WAKE_PIN, GPIO_IRQ_EDGE_RISE, true);
}

int main() {
    // Optional: Lower the voltage to reduce power consumption in sleep
    vreg_set_voltage(VREG_VOLTAGE_0_90);

    // Initialize stdio, not needed for actual deep sleep but for testing/debugging
    stdio_init_all();

    // Configure the GPIO pin for waking up
    setup_wake_pin();

    // Print message before going to deep sleep
    printf("Going to sleep now...\n");

    // Enable deep sleep (Note: the Pico 2 allows real deep sleep with RTC wakeup)
    sleep_goto_dormant();

    // This point will only be reached when the system is woken up
    printf("Woke up from sleep!\n");

    return 0;
}
```

### Key Steps in the Example:
1. **GPIO Wake Pin**: Set a GPIO pin (e.g., pin 15) to wake up the device from deep sleep when a rising edge is detected.
2. **Voltage Reduction**: Optionally, you can reduce the core voltage to 0.9V using `vreg_set_voltage()` to further lower power consumption.
3. **GPIO Interrupts**: The code configures the GPIO interrupts to handle the wake-up.
4. **sleep_goto_dormant()**: This is the key function that puts the Pico into deep sleep, cutting power to most peripherals while maintaining the ability to wake up via external signals.

This example ensures the Pico 2 goes into a low-power deep sleep and wakes up based on a GPIO event.

[...]

PR #3399, opened 2026-03-09
Add common PowerManager header location to include list

maxgerhardt:
Makes `#include "hardware/powman.h"` accessible.

Usefull for working with low power functionality as requested in https://github.com/earlephilhower/arduino-pico/issues/2528.

This is RP2350-only hardware but still in the "rp2_common" header. Code like https://github.com/raspberrypi/pico-extras/blob/master/src/rp2_common/pico_sleep/sleep.c makes a compile time decision to use the powman timer (always-on-timer / lposc) or RTC for RP2040.


maxgerhardt:
CI failures seem to be due to overloaded Github actions server, "rerun only failed jobs" should help.

PR #3414, opened 2026-03-29
added buffer allocation to begin() in I2C master mode

Ikram-cat:
when using the Wire library for master MCUs and setting a buffer size using `setBufferSize()`, the following call of `begin()` did not update the buffer's size thus causing issues.
Copied the code for allocation used in `begin(uint8_t adrr)` and inserted it in the `begin()` function.

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -122,15 +122,6 @@ void TwoWire::begin() {
         return;
     }
 
-    // allocate buffer if necessary
-    if (!_buff) {
-        _buff = (uint8_t *)malloc(_buffSize);
-        if (!_buff) {
-            // ERROR
-            return;
-        }
-    }
-
     _slave = false;
     i2c_init(_i2c, _clkHz);
     i2c_set_slave_mode(_i2c, false, 0);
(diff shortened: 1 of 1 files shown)


earlephilhower (changes requested):
Thanks, you definitely did find and fix an issue.  But I think we maybe can be a little smarter about this and avoid duplicating code.  The memory reallocation should really be done in the `setBufferSize` method, I think, then this change (and the one you got it from in `begin(addr)` slave mode) can be deleted.

If you would please:
. Delete this change and the preexisting check for null & allocate in begin(addr)
. Modify setBufferSize to attempt to reallocate the buffer right in that method and if it fails just keep the old buffer size, like:
````
size_t TwoWire::setBufferSize(size_t bSize) {
    if (_running || (_buffSize == bSize)) {
        // Transmission already running or no change. Report back current buffer size
        return _buffSize;
    }
    // Try and resize the buffer to the new size and handle failure
    auto newSize = max(WIRE_BUFFER_SIZE_MIN, int(bSize)); 
    auto save = _buff
    _buff = (uint8_t *)realloc(_buff, newSize);
    if (!_buff) {
        // Can't allocate size requested, keep old size
        _buff = save;
    } else {
        _buffSize = newSize;
    }
    return _buffSize;
}
````

Ikram-cat:
Yeah this is definitely a more clean solution to the issue. Thank you for your suggestion, added the code just like you suggested as there wasen't much to change.

earlephilhower (approved):
Thanks

PR #3419, opened 2026-04-10
fixed typo for variantchip:RP2350B

jeroenzwan:
There is a typo for variantchip RP2350B for the RP2350.


earlephilhower (approved):
Thanks!

Issue #2808, opened 2025-02-18
XIAO RP2350 - Wire1 not working.


metanav:
I am using a [XIAO RP2350](https://jp.seeedstudio.com/Seeed-XIAO-RP2350-p-5944.html) to read a[ Grove - CO2 & Temperature & Humidity Sensor（SCD41）](https://jp.seeedstudio.com/Grove-CO2-Temperature-Humidity-Sensor-SCD41-p-5025.html) over I2C. This sensor works perfectly using the **Wire** but when I use the **Wire1** it does not work. I have done the following changes in the file `rp2040/hardware/rp2040/4.4.0/variants/seeed_xiao_rp2350`.

```
// #define PIN_WIRE1_SDA  (16u)
// #define PIN_WIRE1_SCL  (17u)
#define PIN_WIRE1_SDA  (0u)
#define PIN_WIRE1_SCL  (1u)
```

The I2C scan code recognizes the sensor address using either **Wire** or **Wire1**. I am using the **GPIO6**(SDA) and **GPIO7**(SCK) for Wire and **GPIO0**(SDA) and **GPIO1(SCK)** for Wire1. Also, I tried another I2C sensor but the same behavior.

I am using this example [sketch](https://github.com/Sensirion/arduino-i2c-scd4x/blob/master/examples/exampleUsage/exampleUsage.ino) with following changes.

```
// Wire.begin();
// sensor.begin(Wire, SCD41_I2C_ADDR_62);
Wire1.begin();
sensor.begin(Wire1, SCD41_I2C_ADDR_62);
```

earlephilhower:
Actually, looking into the contributed Xaoi RP2350 pins_arduino.h, it's really messed up.  They've redefined the Wire device (but not Wire1, so both `Wire` and `Wire1` will try and use the same HW) and confused the wire0/1 pins.

The website shows
![Image](https://github.com/user-attachments/assets/cffc77ab-e3da-441d-ad3f-0598da3994c8)
![Image](https://github.com/user-attachments/assets/60de2c94-6e94-41cf-a1f8-906a7c0c04e4)

So no need to swap I2C devices at all.  I will update #2811 to move `Wire` to the sda0/scl0 16/17, remove the swapping of i2c devices, and set `Wire1` to sda1/scl1 on the 6/7 pinout.

metanav:
Thank you, @earlephilhower! Both Wire and Wire1 are working now!

PR #3421, opened 2026-04-12
Swap defined WIRE0/WIRE1 i2c pins for Seeed XIAO rp2350

doctea:
Found that default i2c `Wire` wasn't working on some recently purchased Seeed RP2350 XIAOs I've just opened, tracked it down to these pins apparently being swapped.

Can see this was looked at in #2808 and reported working there, so am a bit puzzled why it isn't working now unless I make this change...

Previously I used some RP2350 XIAOs that I had purchased a year or two ago, and I did not notice this issue, but they're not in my possession anymore so cannot re-check.  Almost wonder if there has been a hardware revision that has affected this somehow!

I should note I've tested this with `Wire` but not `Wire1`.  Without this patch, I could not get either `Wire` or `Wire1` to work in my project.

### DIFF
--- a/variants/seeed_xiao_rp2350/pins_arduino.h
+++ b/variants/seeed_xiao_rp2350/pins_arduino.h
@@ -57,16 +57,16 @@ static const uint8_t D18 = (9u);
 
 // Wire
 #define __WIRE0_DEVICE (i2c0)
-#define PIN_WIRE0_SDA  (16u)
-#define PIN_WIRE0_SCL  (17u)
+#define PIN_WIRE0_SDA  (6u)
+#define PIN_WIRE0_SCL  (7u)
 #define SDA            PIN_WIRE0_SDA
 #define SCL            PIN_WIRE0_SCL
 #define I2C_SDA        (SDA)
 #define I2C_SCL        (SCL)
 
 #define __WIRE1_DEVICE (i2c1)
-#define PIN_WIRE1_SDA  (6u)
-#define PIN_WIRE1_SCL  (7u)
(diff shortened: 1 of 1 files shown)


earlephilhower (changes requested):
This just basically swaps `Wire` and `Wire1`, no?

For some reason the put the `i2c0` I2C bus on 2 landing pads (GP 16/17) on the back or the board and pulled `i2c1` to the castellations (GP 6/7).

I think it's much less confusing to leave `i2c0/i2c1` as unswapped (`Wire` == `i2c0` == (any i2c0 legal pins, not just 16/17, `Wire1` == `i2c1` == (any legal i2c1 pins, not just 6/7).  That capability is there for some boards where they actually label the board "I2C0/Wire" or something but connect it to `i2c1` hardware.  It's confusing, especially if you want to `setSDA/setSCL` since the legal pins change from i2c0/1.

You can just pass in `Wire1` to whatever library or your code instead of `Wire` if you want the castellated pins.

doctea:
Cheers for looking, I'll test it all again and report back. The pull was originally submitted after 2 hours of head scratching, reflowing pins, swapping ADC boards, swapping MCUs (trying rp2040 xiao which worked fine and then a fresh out of the packet rp2350), trying Wire1 and Wire and multiple i2c scanner sketches, and swapping the pin defs was the only way I could get the scanner to work! I'll try and retrace my steps and see if I made a mistake and Wire1 was working after all.

It's a very odd choice by Seeed for sure, it means the xiao rp2350 isn't source -compatible with the rp2040. I wonder which way around they have it in their own boards package?

PR #3422, opened 2026-04-19
[BLEClient] Fix ignoring timeout parameter

aaronjamt:
Currently, the passed `timeoutsec` parameter is ignored when connecting to a BLE device found by scanning. This PR passes the value to the "real" connection method so it can actually be used.


earlephilhower (approved):
Thx!

PR #3423, opened 2026-04-21
[BLE]: add flag for advertising service UUID.

djpearman:
Hi,

When using the BLE library with a 128-bit service UUID, the advertised name is limited to 8 letters due to the length of the advertised service data, thereby potentially truncating the name. This commit adds a boolean flag to control whether the service UUID is included in the advertising data.

This is achieved by conditionally setting the BLEAdvertising UUID so that it is only included in the data when enabled. Not sure if this is the most elegant solution, but it seems to work.

Best regards,
Dominic

### DIFF
--- a/docs/ble.rst
+++ b/docs/ble.rst
@@ -224,12 +224,13 @@ This call returns the current advertising object for any defined servers.  In ge
 sketches do not need to work with this call or ``BLEAdvertising`` at all, but this
 call lets you do things like add URLs or appearance information.
 
-BLE.startAdvertising()
+BLE.startAdvertising(bool advertiseServiceUUID)
 ----------------------
 
 After all services are created (see below) the sketch needs to use this call to start broadcasting
 for BLE clients to find it.  Call this after all services are defined, or after ``BLE.stopAdvertising()``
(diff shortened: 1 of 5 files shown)


earlephilhower (changes requested):
Is it legal not to advertise a service ID?  Doesn't that make you an invalid beacon (w/o manufacturer data) and not generally discoverable by other devices?

OTW, seems fine since by default it's off and when on, it does what it says on the tin.

One nit:  Please update the docs with this new option.  It's not used by the examples and therefore not easily discovered except by looking at the library sources (which Arduino IDEs don't really let you do)

djpearman:
I'm not aware of any legal requirement to advertise a service ID and haven't come across any references to this end in my BLE dealings. A quick manual check of Bluetooth devices in my vicinity also shows about half or so don't advertise a service UUID (among them a Samsung TV). So it doesn't appear to be absolutely necessary, but I also can't rule it out completely.

Updated the BLE doc with the flag.

earlephilhower (approved):
Thanks.  Yeah, I guess you just connect blind and then enumerate the gatt database in this case, same as if you have a hint as to the services via uuid.

PR #3428, opened 2026-04-29
feat: add Seeed XIAO RP2040 Plus board support

cumin777:


earlephilhower:
@cumin777 Thanks, but is this an available board?  I can't find anything about it on Seeed's website.  OTW, looks okay and thanks for keeping it alphabetized!

cumin777:
Hi @earlephilhower,

  Thanks for the review! This is a real board from Seeed — the XIAO RP2040 Plus (an expanded version of the XIAO RP2040 with additional GPIO pins broken out). It's designed and manufactured by Seeed, but hasn't been listed publicly on their store page yet because it's not yet in stock. The product page will go live once inventory is ready.

  The pin definitions in pins_arduino.h are based on the official schematic and have been verified against the actual board.

  I'd appreciate it if this could be merged now, so that when the product launches, users will have out-of-the-box support from day one. Happy to make any changes if you have further feedback. Thanks again for your time!

PR #3435, opened 2026-05-15
Add clock speed reading before PSRAM direct mode

phuzzyday:
Read clock speed before enabling direct mode for PSRAM to ensure proper configuration.

Fix PSRAM init crash when system clock differs from default
Move clock_get_hz() call before QMI direct mode is enabled.
Flash access is unavailable during direct mode, causing a bus fault
on the instruction fetch when clock_configure_undivided is not
in the XIP cache. Confirmed via debugprobe on WeAct RP2350A.

### DIFF
--- a/cores/rp2040/psram.cpp
+++ b/cores/rp2040/psram.cpp
@@ -165,7 +165,7 @@ size_t __no_inline_not_in_flash_func(psram_init)(uint cs_pin) {
     // unavailable while QMI direct mode is enabled)
     const int max_psram_freq = RP2350_PSRAM_MAX_SCK_HZ;
     const int clock_hz = clock_get_hz(clk_sys);
-    
+
     // Enable direct mode, PSRAM CS, clkdiv of 10.
     qmi_hw->direct_csr = 10 << QMI_DIRECT_CSR_CLKDIV_LSB | \
                          QMI_DIRECT_CSR_EN_BITS | \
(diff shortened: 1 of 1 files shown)


earlephilhower:
LGTM.  I deleted the spaced on the blank line after your change.  You'll need to delete the blank spaces after the change like I did in your Micropython PR.  Their commit message stuff, I unfortunately haven't read up on.

Thx!

phuzzyday:
I didn't realize I left a blank. my bad! I am a bit new to this...

earlephilhower:
No worries.  The editor probably did it behind your back.  Most projects have formatting scripts you can run before committing to ensure meaningless whitespace changes don't sneak in.

I'll pull out my Sparkfun board with psram this weekend and do A/B testing on this PR.  Without your changes I should be able to get random hangs on startup with different clocking, which go away with the fix.

phuzzyday:
I did change what I THOUGHT was wrong in Micropython, might have been mistaken, but after a couple edits, I realized the change was in the main repository now, so I guess something went right.

It sure made a difference here! I hope the fact that this was a custom modification to the board doesn't throw things out, but I would suspect the wiring would be the same anyhow? All pins on the same lines other than the CS? Thanks for helping though, Github is a new and strange place for me...  Not all Phils are created equal I guess. ;-)

I looked at the Sparkfun schematics for a couple of their RP2350's, and the connectors ended up being the same as what I slammed on there, the only difference is that I don't have a 10K pullup on my CS line. I mean, Claude's logic was sound anyhow, and the suggested fix worked, so it's probably not that.

earlephilhower (approved):
LGTM!

PR #3438, opened 2026-05-21
feat: Add Soldered NULA Ethernet W55RP20 pin definitions

JosipKuci:
Add pin definitions for the upcoming Soldered NULA Ethernet W55RP20 board by Soldered Electronics
The board is currently still being developed, so I will be attaching the schematic as well as the product image (Note that both the image and schematic call it the "NULA Ether", this will be changed to Ethernet in the final release)

[NULA Ethernet W55RP20 Schematic.pdf](https://github.com/user-attachments/files/28091426/NULA.Ethernet.W55RP20.Schematic.pdf)

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ Read the [Contributing Guide](https://github.com/earlephilhower/arduino-pico/blo
 * Solder Party RP2040 Stamp
 * Solder Party RP2350 Stamp
 * Solder Party RP2350 Stamp XL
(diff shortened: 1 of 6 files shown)


earlephilhower:
Neat.  Check for trailing spaces or tabs in you header file that the CI system is complaining about to get it all green.

JosipKuci:
@earlephilhower I have removed the trailing spaces in a few lines in the newest commit, so it should be good to go!

PR #3452, opened 2026-06-13
Allow disabling UART TX or RX pins

### LINKED ISSUE #3450: Feature Request: Ability to disable RX or TX in Serial1/Serial2
Currently, Serial1 and Serial2 are typically defined some default pins per the board, and then there is no way to disable that pin if you only want receive or transmit, not both. Since you can't do setTX(99) or setTX(-1) or some other way of setting the pin to be ignored, the result is a pin is used when unintended (see issue [#1307](https://github.com/earlephilhower/arduino-pico/issues/1307)), or a pin (from a limited set of pins) must be sacrificed.

It would be nice if there was a sentinel to mean "no pin" or otherwise some way to say transmit only or receive only.

My specific case is talking to an ESC using the KISS telemetry protocol which is a one way serial line.

Thanks.

earlephilhower:
Frees that pin for use by your app in single direction UART protocols. Fixes #3450

### DIFF
--- a/cores/rp2040/SerialUART.cpp
+++ b/cores/rp2040/SerialUART.cpp
@@ -46,9 +46,11 @@ bool SerialUART::setRX(pin_size_t pin) {
                                   };
 #endif
 
-    if ((!_running) && ((1LL << pin) & valid[uart_get_index(_uart)])) {
-        _rx = pin;
-        return true;
+    if (!_running) {
+        if (((1LL << pin) & valid[uart_get_index(_uart)]) || (pin == UART_PIN_NOT_DEFINED)) {
+            _rx = pin;
+            return true;
+        }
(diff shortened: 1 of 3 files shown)


peternlewis:
I tested it by downloading the patch and applying the changes directly to SerialUART.h/cpp.

Looks good, works well. I tested leaving the setTX unset (which on my generic RP2354 board makes GPIO 0 the default TX line). If unset, the line is pulled high. If setTX(UART_PIN_NOT_DEFINED) the line is left unchanged and can be controlled by digitalWrite as normal. If setTX(0) the line is high controlled by the UART.

My only quibble would be the mention of `-1` in the comments, when it is really 255u (UART_PIN_NOT_DEFINED).

PR #3461, opened 2026-06-27
Fix display names for PCBCupid Glyph 2040 and Glyph Mini 2040

Abhi-00047:


earlephilhower:
Err, is this really the name you were going for?  Seems wrong having the same word twice in a row:
> "PCBCupid Pcbcupid GLYPH MINI 2040"

If you wanted to go Initial caps, then just change the manufacturer to `Pcbcupid` from `PCBCupid`, don't add to the name...

earlephilhower (changes requested):
(See comment above)

earlephilhower (changes requested):
Name looks okay now, but I just noticed it's not in alphabetical order.  Can you move your boards up before `PicoLume` and not at the end of the `P`s?

earlephilhower (approved):
Thanks!

PR #3473, opened 2026-07-17
Revert "Check that any sketch with OTA has a FS defined"

earlephilhower:
Reverts earlephilhower/arduino-pico#3472

I'm getting `PicoOTA` pulled in for certain WiFi-only sketches, so this specific change causes build failures where there really is no problem.  Your sketch alone will need to `assert` if it cares that there is a FS.


maxgerhardt:
Still, a macro for the code
```cpp
// Ensure OTA is possible, requires a FS
static_assert(FS_END - FS_START > 127 * 1024, "Need to define a 128K or larger filesystem to use OTA");
```
might be useful, with just configurable filesizes. So, users who want to require their firmwares to be built with a filesystem of a minimum size can do this in a standardized way.

```cpp
#define STRINGIFY_IMPL(x) #x
#define STRINGIFY(x) STRINGIFY_IMPL(x)

#define REQUIRE_FS_SIZE_KB(kb)                                                   \
    static_assert(                                                               \
        (FS_END - FS_START) > ((kb) - 1) * 1024,                                 \
        "Need to define a " STRINGIFY(kb) "K or larger filesystem to use OTA")
```
Possibly with `#include <assert.h>`

earlephilhower:
I'm of two minds on this.  Originally I was hoping to give a default check which "just worked" without people needing to do anything.  But `arduino-cli` seems to pull in `PicoOTA` (could be my WiFi library fault, but didn't really dig since it's was a breaking problem).

A new macro like this seems like it's something that wouldn't get much use.  Anyone who would use it would, I think, already be able to ensure it themselves in some similar way.  And unless you read the docs it's not so easy to discover...

PR #3474, opened 2026-07-18
Adds new board iLabs CPico RP2350 8MB

PontusO:
As the description suggests this adds the new iLabs CPico RP2350 board with 8MByte of RAM and iProbe UART support.


earlephilhower (approved):
LGTM!

PR #3475, opened 2026-07-18
Add support for WeAct Studio RP2350B Core Board

herbkonik:
### Summary

This PR adds board and variant definitions for the **WeAct Studio RP2350B Core Board** (`WeActStudio.RP2350BCoreBoard`), an RP2350B-based board with 48 GPIOs (GP0-GP47), ADC on GP40-GP47 (8 channels).

### Background

We initially suspected our two WeAct RP2350B boards had mislabeled/counterfeit silicon. Symptoms:
- `gpio_pull_down()` worked reliably on GP0-31 (Bank 0) but not on GP32+ (Bank 2).
- External pulldowns on GP32+ didn't fix digital input behavior.
- ADC on GP40-47 worked but floated high without an external pulldown.
- The chip marking (`RP2350B0A4`) should, per Raspberry Pi, already include the Erratum E9 fix, yet behaved as if it didn't.

We considered this consistent with Erratum E9 (pull-down lock-up) despite the revision marking, and were ready to write the board off as unreliable.

### Root cause

It turned out to be a board/SDK configuration issue, not silicon:

- Building with `-DPICO_BOARD=pico2` (generic Pico 2 board header) implicitly sets `PICO_RP2350A=1`, which is correct for the official Pico 2 (RP2350A die, 32 GPIOs) but silently leaves GP32-47 unconfigured on an RP2350B-die board like this one.
- Building with the correct, pico-sdk-provided `weact_studio_rp2350b_core` board header (`-DPICO_BOARD=weact_studio_rp2350b_core`) resolved every symptom above, with no external pulldowns required. This confirmed there was no Erratum E9 issue on Bank 2 (E9 only affects Bank 0 GPIO leakage per the datasheet) and no counterfeit-silicon issue - it was purely a `PICO_RP2350A` misconfiguration.

### Porting to arduino-pico

We initially added a local, unofficial `weact_rp2350b` entry to `boards.txt` and a corresponding `variants/weact_rp2350b/pins_arduino.h`, adapted from an existing working RP2350B variant (`MyRP_2350B`). That first attempt still had `#define PICO_RP2350A 1` at the top (a leftover/copy-paste mistake reproducing the exact same root cause above), with `A4`-`A7` manually patched in afterward to compensate - masking the symptom instead of fixing the cause.

The corrected variant simply sets:
```c
#define PICO_RP2350A 0 // RP2350B
```
and removes the manual `A4`-`A7` patch entirely, since `variants/generic/common.h` already defines `D30`-`D47` and `A0`-`A7` automatically in its `PICO_RP2350A == 0` branch, matching this board's actual pinout (A0=GP40 ... A7=GP47).

### Testing performed

- Digital input test (`pinMode(pin, INPUT)` + `digitalRead()`) across effectively all 48 GPIOs (GP0-GP47), excluding `LED_BUILTIN` (GP25, used as output) and GP23 (tied to the optional onboard PSRAM chip select - not usable as a general GPIO on this board). All confirmed working without any external pulldown, including GP32-47 which previously failed.
- `INPUT_PULLDOWN` verified functional on a sample of pins on both sides of the Bank 0/Bank 2 boundary (5 pins <= GP30, 5 pins > GP30), confirmed by direct multimeter measurement on the open pin (stable ~0V with pulldown enabled vs. floating/unstable without).
- ADC test on all 8 channels (GP40-47) with `analogReadResolution(12)`: full-range values (min ~8, max ~4088) with normal single-sample noise of about +/-2 counts. Note that `INPUT_PULLDOWN` has no effect on `analogRead()` behavior on these pins - this is expected, since the SDK's `adc_gpio_init()` explicitly disables pin pulls before sampling to avoid loading the analog signal; this is unrelated to the Bank 2 digital I/O fix above.
- All of the above verified in a from-scratch PlatformIO environment on a second machine with no prior RP2040/RP2350 project history, using this fork directly via `platform_packages`, to rule out any leftover local package-cache state.

A minimal, reproducible test project (PlatformIO project + `boards/weact_rp2350b.json` + test firmware) is available here: https://github.com/herbkonik/WeAct_RP2350B_Core_Test

### Known open item (not addressed by this PR)

`PINS_COUNT`, `NUM_DIGITAL_PINS`, and `NUM_ANALOG_INPUTS` in `variants/generic/common.h` are currently hardcoded to `30`/`30`/`4` regardless of `PICO_RP2350A`. `D30`-`D47` and `A4`-`A7` are still usable as explicit pin constants, but any code that iterates using these count macros won't see the extra RP2350B pins. Happy to look into fixing this in a follow-up if maintainers agree it's in scope - flagging it here so it's not confused with a limitation of this board's variant specifically.

### Files changed

- `boards.txt`: added `weact_rp2350b.*` entries
- `variants/weact_rp2350b/pins_arduino.h`: new file


earlephilhower:
I think your AI has gone mad here.  Please review the changes it's suggesting and the [contributing docs](https://github.com/earlephilhower/arduino-pico/blob/master/docs/contrib.rst).

If all you're doing is changing an existing pins.h file, there should be no difference in the boards.txt file.  And if you're adding a completely new board, manually or AI editing the boards.txt is not the right way to get it done.  It's all documented in the contrib.rst.

earlephilhower (changes requested):
See prior comment.

herbkonik:
Sorry for creating this PR without reading the guidelines, i did it now. 
I'll use the right way to create the the files again and put a new PR if it's done.

earlephilhower (changes requested):
This again has a bunch of changes that look manually or Ai made to boards.txt.  Please reread the contrib.rst to see how to correctly add a board.

Also make sure you're running against the latest git, I think you're missing a commit or two to the boards file.  You should only see your new board settings in the file, not thousands of lines of deletion and edits.

herbkonik:
Yes, you are right, i uploadet the files and miss to commit... I am not realy familiar with git / github... Sorry again.

earlephilhower (approved):
Looks good now, thx!

PR #3480, opened 2026-07-23
Bundle iLabs_ESP-NOW library and add ESP-NOW to the ESP Wifi Type menu

PontusO:
Part 1: bundle the iLabs_ESP-NOW library as a git submodule under libraries/, the same way ESPHost/AsyncUDP/SdFat ship, pinned to v1.0.0. Users get the ESP-NOW API out of the box. The library's AT+EN protocol uses an "+ENVER" firmware-version token that codespell reads as a typo of "never", so "enver" is added to the codespell ignore_words_list.

Part 2: expose ESP-NOW as a selectable "ESP Wifi Type" for the iLabs boards (tools/makeboards.py + regenerated boards.txt/JSON):
- BuildWifiType gains a 3rd option (ESP-NOW) for the two boards that already have the menu (challenger_2040_wifi_ble, challenger_2350_wifi6_ble5).
- New BuildWifiTypeAT (ESP AT default + ESP-NOW) wired to the five previously-hardcoded WIFIESPAT2 boards (challenger_2040_wifi, challenger_2040_wifi6_ble, challenger_nb_2040_wifi, connectivity_2040_lte_wifi_ble, ilabs_rpico32).
- The hardcoded ["WIFIESPAT2"] extra_flag is removed from those five boards; the menu's ESP AT default now supplies -DWIFIESPAT2, so selecting ESP-NOW yields -DILABS_ESPNOW alone (no double define).

ESP-AT stays the default on every board, so existing builds are unchanged. PlatformIO users of the five converted boards must now set the desired flag in platformio.ini (matching the existing menu boards, whose JSON already carries no espwifitype define).

Part 3: gate the variant ESP reset on ESP-NOW not being selected. When -DILABS_ESPNOW is set the iLabs_ESP-NOW library owns the ESP reset sequence (ESP_NOW.setLink()), so the AT-style reset in initVariant() is compiled out under #ifndef ILABS_ESPNOW:
- challenger_2040_wifi, challenger_2040_wifi_ble, challenger_2040_wifi6_ble, challenger_2350_wifi6_ble5 and challenger_nb_2040_wifi already reset Challenger2040WiFi in initVariant(); that call is now gated.
- ilabs_rpico32 had no board_init.cpp; add one that gates the reset of its Ilabs2040WiFi object the same way.

Part 4: document the feature in docs/ilabs_espnow.rst (linked from the index toctree): what ESP-NOW is, the supported iLabs boards, selecting it from the ESP Wifi Type menu, the two sketch changes versus an ESP32, a first sketch, board reset handling and discovery.


earlephilhower:
@PontusO so is there really no issue with the latest adafruit tinyusb?  I'll undo the reversion #3479 in that case.

PontusO:
i can see that packages are flowing in very differently in the latest adafruit release but with my changes to the sketch it now seemingly works fine. I have run it a couple of 100 times now and besides the odd hang inside the stack which is not related to this it all seems to work fine.

earlephilhower (approved):
LGTM!

So you can use ESP-NOW with the AT FW you're shipping just to "mesh" communicate with other ESP chips, or if you need full fat TCP/IP then the lwip_ESPHost is the way to go, right?

Thx for adding docs, too, since discoverability is always a pain!

PontusO:
Yes, that is the way to go. The ESP-NOW AT FW only handles the ESP-NOW part of the protocol.

PR #3490, opened 2026-08-13
Add Soldered NULA Node RP2040 board definition

JosipKuci:
Add board definition for the upcoming Soldered NULA Node RP2040 dev board, which features the RP2040 microcontroller along with an onboard RF2 Module. It is very similar to the Pico W, but with a different pin assignment and a flash chip of 16MB. The board is still being developed, so I am attaching an image of it as well as its schematic for reference:

[NULA Node RP2040 Schematic.pdf](https://github.com/user-attachments/files/31025658/NULA.Node.RP2040.Schematic.pdf)


earlephilhower:
Looks fine to me.  Let me know if you want to merge or wait for more testing on your end. 

And a general question:  Why stick with the RP2040 and not move to the 2350 for a new board?  The power is way lower with working sleep modes, the same SDK works, you get 2x the SRAM, and much more performance for $0.10 USD increase in BOM, no?

JosipKuci:
> Looks fine to me. Let me know if you want to merge or wait for more testing on your end.
> 
> And a general question: Why stick with the RP2040 and not move to the 2350 for a new board? The power is way lower with working sleep modes, the same SDK works, you get 2x the SRAM, and much more performance for $0.10 USD increase in BOM, no?

We have tested the wifi connectivity as well as the defined default communication protocols, so its good to go!

Regarding your question, we have a lot of these chips left over from other projects and the solder kits we do, so this is a way to use them. I completely agree that the RP2350 is the superior chip, and we do have the NULA Max board with it.

Issue #3491, opened 2026-08-17
Feature Request: Bundling a camera library like esp32-camera in the core.


umeiko:
Hi Earle,

I've written **PicoCamera**, a camera capture library for raspberrypi pico with an
API aligned with esp32-camera, and I'd like to ask whether you'd be open to
bundling it with the core — and if so, in which form.

Repo: https://github.com/umeiko/PicoCamera
(also published in the Arduino Library Manager and the PlatformIO Registry)

What it does:

- Captures frames from parallel-interface (DVP) camera sensors using PIO + DMA,
  with a double-buffered frame queue
- All pins (VSYNC/HREF/PCLK/D0–D7/SCCB/XCLK) are configured at runtime in the
  config struct — no PIO source editing or pioasm build step needed
- esp32-camera-style API (`pico_camera_init()`, `pico_camera_fb_get()`,
  `sensor_t` controls), so existing sketches and knowledge transfer over
- Auto-detects the sensor model over SCCB; OV2640, OV3660 and OV7670 supported,
  RGB565 and JPEG (on sensors with an encoder)
- MIT licensed; only depends on the core and pico-sdk

It already follows the core's conventions: `architectures=rp2040`, 4-space
K&R formatting, and the examples build warning-free with the CI flags
(`-Wall -Wextra -Werror -Wdouble-promotion`). Examples that need third-party
display libraries carry a `.test.skip`.

Two possible forms, happy with whichever you prefer:

1. **Git submodule** under `libraries/`, like AsyncUDP / SdFat /
   iLabs_ESP-NOW — the repo stays independently maintained and released.
2. **In-tree** under `libraries/PicoCamera/`, added to `tests/restyle.sh`,
   maintained through PRs to this repo.

If you're interested I'll prepare a PR for the form you choose. If you'd
rather keep the core lean and leave it as an external library, that's
completely fine too — just let me know.

Thanks!

earlephilhower:
First of all, that's awesome! 👍 

Second of all, that sucks!  I just got my aliexpress package with a set of OV cameras this weekend and was reading the datasheet to write what you seem to have already done. 😆  (Interesting PIO program runtime builder, BTW...I guess there's no "elegant" way to do it like we do here w/the SW Serial port since the instruction count itself is variable).

(Third of all, looks like I need to get working on the OpenCV and Tensorflow Lite Micro ports...)

[...]

PR #3492, opened 2026-08-18
Add PicoCamera camera library as a submodule

umeiko:
As discussed in #3491, this bundles PicoCamera as a git submodule under `libraries/`, the same way Adafruit_TinyUSB_Arduino and friends are handled.

[PicoCamera](https://github.com/umeiko/PicoCamera) is a camera capture library for RP2040/RP2350 with an esp32-camera aligned API:

- PIO + DMA capture from parallel-interface (DVP) sensors; all pins (VSYNC/HREF/PCLK/D0-D7/SCCB/XCLK) configured at runtime — no pioasm build step
- OV2640, OV3660 and OV7670 supported, RGB565 and JPEG output
- Published in the Arduino Library Manager and the PlatformIO Registry; the submodule is pinned to the 0.3.0 release tag
- Version macros (`PICO_CAMERA_VERSION_*`) let sketches detect which copy they compiled against when a Library Manager install overrides the bundled one.

CI notes:

- Excluded from codespell (skip list) and deliberately not added to `tests/restyle.sh`, so the upstream repo keeps its own style, per #3491
- All examples build warning-free with `-Wall -Wextra -Werror
-Wdouble-promotion` on rp2040, rp2350 and rp2350-riscv (verified locally against master); the TFT preview example carries a `.test.skip` marker because it needs the external TFT_eSPI library.


earlephilhower (approved):
LGTM.  If you'd like to list the PicoCamera license in the README.md you can also add it with the others at the end.  (I tried to keep up but am sure I missed a library or two there!)

Let me give this a quick run on my boards for sanity's sake this weekend, and then we'll merge.

umeiko:
> LGTM. If you'd like to list the PicoCamera license in the README.md you can also add it with the others at the end. (I tried to keep up but am sure I missed a library or two there!)
> 
> Let me give this a quick run on my boards for sanity's sake this weekend, and then we'll merge.

Thanks! Added the PicoCamera license line to the README credits as you suggested.  If anything misbehaves on your boards, let me know and I'll fix it upstream and re-point the submodule. 🤝

earlephilhower:
Really nice work!

From finishing wire wrapping the camera and a Pico2 to running the image capture demo and capturing images was under 3 minutes, 2 of which were wasted because I have one of the cameras w/o onboard JPEG so needed to change the well documented `define` in your example.

Nice examples and docs, BTW.  The only thing I needed to change in your example was to fix the RST pin because the Pico PCBs don't break out pin GP29.

umeiko:
Thank you for the quick test and the kind words – much appreciated!
To help other users get started more easily, I’ve just added a dedicated documentation page for the Camera library in #3500.

PR #3500, opened 2026-08-24
docs: add Camera (PicoCamera) documentation page

umeiko:
Documents the newly merged PicoCamera library: camera_config_t configuration with example, SCCB/I2C shared bus, memory planning for RGB565/JPEG frame buffers on RP2040/RP2350 (incl. PSRAM), init / capture / deinit lifecycle, camera_fb_t frame handling, runtime sensor control, full API list, and example sketches.

### DIFF
--- a/docs/camera.rst
+++ b/docs/camera.rst
@@ -1,0 +1,400 @@
+Camera (PicoCamera) Library
+===========================
+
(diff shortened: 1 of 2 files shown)


earlephilhower (approved):
Thx!

PR #3508, opened 2026-09-02
Avoid glitching/reset-to-0 on `pinMode` calls

### LINKED ISSUE #3507: Output glitches are unavoidable when trying to init pin as GPIO output driving HIGH
Hi,

Inside the `pinMode()` implementation, calling `gpio_init(ulPin);` sets the value of the output to 0, making it impossible to set a GPIO HIGH before setting its direction to output.
Normally `digitalWrite()` can be called prior to `pinMode` to accomplish this, and I think its necessary to preserve this behavior.

I noticed the `digitalWrite()` implementation sets the pin function to GPIO_FUNC_SIO (but not `digitalRead()`...why?), so calling `pinMode()` is currently not necessary for outputs... Right now I'm just using `digitalWrite()` for setting pins to outputs, but that's not a clean approach.

I think the crux of the issue is `gpio_init()` calling `gpio_put(gpio,0)`, but thats from the pico sdk so I'm inclined to leave it.. I think we could just replace `gpio_init()` with `gpio_set_function()` in the pinMode OUTPUT cases. If for some reason `gpio_init()` is necessary, we could also read the value in the gpio_out register and write it back before setting the direction:

```
    case OUTPUT_4MA:
    	// get value and write it before setting direction to avoid output glitches
        bool val;
        val = ((sio_hw->gpio_out & (1u << ulPin)) != 0);
        gpio_init(ulPin);
        gpio_set_drive_strength(ulPin, GPIO_DRIVE_STRENGTH_4MA);
        gpio_put(ulPin, val);
        gpio_set_dir(ulPin, true);

earlephilhower:
pinMode was calling gpio_init which overwrites any prior GPIO state on the pin in question.,  This means sequences like
````
pinMode(0, OUTPUT);
digitalWrite(0, HIGH);
pinMode(0, OUTPUT);
````
will leave the GPIO at *0* afterwards which is quite wrong.

Manually set the GPIO MUX to avoid the issue on pinMode calls

Fixes #3507

### DIFF
--- a/cores/rp2040/wiring_digital.cpp
+++ b/cores/rp2040/wiring_digital.cpp
@@ -34,40 +34,40 @@ extern "C" void __pinMode(pin_size_t ulPin, PinMode ulMode) {
 
     switch (ulMode) {
     case INPUT:
-        gpio_init(ulPin);
         gpio_set_dir(ulPin, false);
+        gpio_set_function(ulPin, GPIO_FUNC_SIO);
         gpio_disable_pulls(ulPin);
         break;
     case INPUT_PULLUP:
-        gpio_init(ulPin);
         gpio_set_dir(ulPin, false);
+        gpio_set_function(ulPin, GPIO_FUNC_SIO);
         gpio_pull_up(ulPin);
         gpio_put(ulPin, 0);
         break;
     case INPUT_PULLDOWN:
-        gpio_init(ulPin);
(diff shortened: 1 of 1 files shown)


jcormier:
LGTM

nzins-cl:
Me too :+1:

PR #3511, opened 2026-09-03
Make DEBUGV & friends append a newline, remove newlines from call sites

### LINKED ISSUE #3510: DEBUGV(...) calls are inconsistent about newlines
I know this is internal, but I still like to see the internal chatter! Go ahead and close with prejudice if you can't be bothered to even think about something this trivial.

When `DEBUG_RP2040_PORT` is defined (eg. by the Arduino `Debug Port` menu item), `DEBUGV(...)` and friends just wrap `DEBUG_RP2040_PORT.printf(...)` with no added newline or other jazz.

Some call sites add CR-LF:
```cpp
DEBUGV("Dir::openFile: invalid mode `%s`\r\n", mode);
```

Some call sites add naked LF:
```cpp
DEBUGV("SDFSImpl::open() called on unmounted FS\n");
```

Some call sites add no line ending at all (probably a bug?):
```cpp
DEBUGV("#error: FS: no implementation");
```

Occasionally (rarely) there seems to be a two-part print:
```cpp
DEBUGV("Scan connecting %s at %s ... ", e.name(), e.addressString());
memcpy(a, e.address(), sizeof(a));
if (!a2dp_source_establish_stream(a, &media_tracker.a2dp_cid)) {
    DEBUGV("Connection established\n");
    return true;
}
DEBUGV("Failed\n");
```

I doubt this is high priority for anyone, but I would propose (and can send a PR if this is welcome!) that `DEBUGV` (and `DEBUGCORE` etc) add CR-LF itself, all line endings get removed from the call sites, and the two-part prints are sacrificed and become two lines. That could I suppose add extra line-endings to anyone outside the core using `DEBUGV`. I say "oh well, undocumented API" but others could reasonably disagree.

egnor:
Fixes #3510.

`DEBUGV`, `DEBUGCORE`, `DEBUGWIRE`, `DEBUGSPI` (in `debug_internal.h`) and `DEBUGBLE` (in `BLEDebug.h`) now concatenate `"\r\n"` onto the format string:

```cpp
#define DEBUGV(fmt, ...) do { DEBUG_RP2040_PORT.printf(fmt "\r\n", ## __VA_ARGS__); DEBUG_RP2040_PORT.flush(); } while (0)
```

This requires the format argument to be a string literal, which it already was at every one of the ~350 call sites (a few use `"..." PRIu32 "..."`, which is still a literal after preprocessing). The line ending is a single write, so the other core can't interleave between the text and the newline. `\r\n` matches `Print::println()`.

**Mechanical changes:** every call site has its trailing `\n` or `\r\n` stripped (the tree was a mix of both). The six call sites that had no line ending at all (`FS::begin`, `FatFSImpl::openFile`, `LittleFS` size check, `SDFSImpl::openFile`, two in `UdpContext`) now get one for free.

**Hand-edited call sites, worth a look:**
- `BluetoothHCI::setBLEName()` printed the ATTDB hex dump as `"ATTDB: "` + a loop of `"%02X "` + `"\n"`. It now builds the hex into a `String` and prints it with one `DEBUGV`, under `#if defined(DEBUG_RP2040_PORT)` so no work is done when debugging is off.
- `A2DPSource::connect()` and `BluetoothHIDMaster::connectBLE()` printed `"Scan connecting %s at %s ... "` and then `"Connection established\n"` or `"Failed\n"` on the same line. These are now separate lines, and the bare `"Failed"` was renamed `"Connection failed"` so it reads sensibly on its own.
- `A2DPSink.cpp` had `DEBUGV("\n\n")` after the supported-notifications list, and `A2DPSource.cpp` had `"A2DP Source: Signaling released.\n\n"`. Both now emit one line rather than one line plus a blank one. `BluetoothMediaConfigurationSBC::dump()` ends with `DEBUGV("")`, which still produces a single blank separator line.
- Two format strings had a stray space before the newline (`"... a2dp_cid 0x%02x \n"`, `"bitpool_value [%d, %d] \n"`); the space was dropped too.

**Not touched:** `DEBUG_HTTPCLIENT` (HTTPClient.h) and `DEBUG_BSSL` (WiFiClientSecureBearSSL.cpp, CertStoreBearSSL.cpp) are separate library-local macros with their own conventions and are left alone.

**Verification:** every remaining format-string expression was extracted and compiled on the host as `const char *f = <expr> "\r\n";` with `-Wall -Werror`, plus a check that no `\n` or `\r` remains in any of them. I don't have `arduino-cli` or `astyle` locally, so I'm relying on CI for the full build and style check. Anyone outside the core using `DEBUGV` with their own trailing newline will now get a blank line between messages.

https://claude.ai/code/session_01Sobzdo3ykmeJJRdKSMtMn7

### DIFF
--- a/cores/rp2040/CoreMutex.cpp
+++ b/cores/rp2040/CoreMutex.cpp
@@ -46,7 +46,7 @@ CoreMutex::CoreMutex(mutex_t *mutex, uint8_t option) {
     if (!mutex_try_enter(_mutex, &owner)) {
         if (owner == get_core_num()) { // Deadlock!
             if (_option & DebugEnable) {
-                DEBUGCORE("CoreMutex - Deadlock detected!\n");
+                DEBUGCORE("CoreMutex - Deadlock detected!");
             }
             return;
         }
--- a/cores/rp2040/FS.cpp
+++ b/cores/rp2040/FS.cpp
@@ -248,7 +248,7 @@ File Dir::openFile(const char* mode) {
     OpenMode om;
     AccessMode am;
     if (!sflags(mode, om, am)) {
-        DEBUGV("Dir::openFile: invalid mode `%s`\r\n", mode);
+        DEBUGV("Dir::openFile: invalid mode `%s`", mode);
         return File();
     }
 
@@ -343,7 +343,7 @@ bool FS::begin() {
     }
     _impl->setTimeCallback(_timeCallback);
     bool ret = _impl->begin();
-    DEBUGV("%s\n", ret ? "" : "#error: FS could not start");
+    DEBUGV("%s", ret ? "" : "#error: FS could not start");
     return ret;
 }
 
@@ -393,7 +393,7 @@ File FS::open(const char* path, const char* mode) {
     OpenMode om;
     AccessMode am;
     if (!sflags(mode, om, am)) {
-        DEBUGV("FS::open: invalid mode `%s`\r\n", mode);
+        DEBUGV("FS::open: invalid mode `%s`", mode);
         return File();
     }
     File f(_impl->open(path, om, am), this);
@@ -556,7 +556,7 @@ template<>
 bool mount<FS>(FS& fs, const char* mountPoint) {
     FSImplPtr p = fs._impl;
     if (!p || !p->mount()) {
-        DEBUGV("FSImpl mount failed\r\n");
+        DEBUGV("FSImpl mount failed");
         return false;
     }
 
@@ -578,7 +578,7 @@ File open(const char* path, const char* mode) {
     OpenMode om;
     AccessMode am;
     if (!sflags(mode, om, am)) {
-        DEBUGV("open: invalid mode `%s`\r\n", mode);
+        DEBUGV("open: invalid mode `%s`", mode);
         return File();
     }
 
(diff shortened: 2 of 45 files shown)


egnor:
(Note from the lowly human: Make sure to look at the use of `String` in `BluetoothHCI::setBLEName()`; I'm not sure how you'd feel about that. I could imagine other approaches -- print every byte on its own line, assemble bytes into a fixed-size hex-dump line-buffer, etc.)

earlephilhower:
Very nice, thanks!  I'll look at it tonight.  The CI build went cleanly.

Might want to add something in the `contrib.rst` about the debug formatting rule.  Nobody reads it, but at least I can point to it when a `DEBUGV("blah\n");` appears. :laughing:

earlephilhower:
Oh, and the only debug format I'm really concerned about not changing are the ones in `BluetoothDebug.cpp` because they are used by the BTStack internal debug tools when we have an issue.  And that's not a macro you would have touched, it slots into the BTStack internals.

earlephilhower (approved):
What a trip through memory lane.  Some of these debug messages I or a team member wrote probably a decade ago and they're still kicking around.

The expansion of ATTDB vis `String` is not an issue.  This is a single call and the DB itself is limited in size to 512B or less so no memory issue.

The trailing space issues, that was just a typo or from the BTStack sample.  And the automated replacement seems to not have broken anything else from diff inspection.

LGTM!

BTW, I was greeted with this image after going thru the online diffs...
Image
Never seen it before but I guess someone at MSFT still has a sense of humor! 😃

PR #3518, opened 2026-09-04
uf2conv.py: don't flash the same UF2 drive twice via different paths

egnor:
`uf2conv.py` scans several directories for mounted UF2 drives, and on most Linux systems more than one of them leads to the same mount point, since `/var/run` is a symlink to `/run`. The drive then gets flashed twice:

```
Flashing /run/media/egnor/RPI-RP2 (RPI-RP2)
Wrote 184320 bytes to /run/media/egnor/RPI-RP2/NEW.UF2
Flashing /var/run/media/egnor/RPI-RP2 (RPI-RP2)
Wrote 184320 bytes to /var/run/media/egnor/RPI-RP2/NEW.UF2
```

The second write only works because the board hasn't finished rebooting yet; if it has, the write fails with an I/O error after the upload actually succeeded.

This deduplicates the Linux drive list by `os.path.realpath()`. The Windows and macOS branches aren't touched.

Verified with a Feather RP2040 on Ubuntu 26.04: with the drive mounted at `/media/RPI-RP2` and a symlink `/mnt/RPI-RP2` pointing at it, `uf2conv.py --list` shows both paths before and only `/media/RPI-RP2` after; the udisks case above (`/run/media` + `/var/run/media`) collapses to one entry the same way.

### DIFF
--- a/tools/uf2conv.py
+++ b/tools/uf2conv.py
@@ -300,6 +300,8 @@ def get_drives():
             possibly_any("/var/run/media", drives, u)
             # Add from udisksctl info?
             # Add from /proc/mounts?
+            # avoid uploading twice due to symlinks (eg. /var/run => /run)
+            drives = list(set(os.path.realpath(d) for d in drives))
 
     def has_info(d):
         try:


earlephilhower (approved):
LGTM.  Alternatively, I was going to early exit the write_file `for` loop at the end on a successful write but never got around to it because I could never hit the case repeatably to test any changes.

egnor:
I actually thought it was odd that it would write every UF2 disk it could find, not sure if that was intended behavior... arguably if there are multiple UF2 disks attached an error might make more sense, since there's no way for the user to identify which one they meant, but I wasn't clear if some people like to flash a pile of devices at once with this pathway.

earlephilhower:
This should still allow that multiple uploads, but it's not really an intentional behavior.  If you're flashing new boards in bulk I hope you're not using the IDE to do it!

PR #3519, opened 2026-09-04
uf2conv.py: skip the 2s serial-port wait for the discovered UF2_Board port

egnor:
After flashing, `uf2conv.py` waits up to 2 seconds for the `--serial` port to become writable again, unless the port is the synthetic UF2 drive port, in which case there is nothing to wait for. That check compares against `"UF2 Board"`, but `pluggable_discovery.py` advertises the port with address `UF2_Board` (the *label* is `UF2 Board`), and arduino-cli passes the address. So the check never matches and every upload to a discovered UF2 drive pays the full 2 second timeout.

This accepts both spellings, and skips the `os.access()` loop entirely when `--serial` wasn't given (previously that raised a `TypeError` that the bare `except` swallowed, so the behavior there is unchanged, just explicit).

Measured with a Feather RP2040 in BOOTSEL on Linux: `uf2conv.py --serial UF2_Board --family RP2040 --deploy x.uf2` takes ~5.5 s before and ~3.5 s after.

### DIFF
--- a/tools/uf2conv.py
+++ b/tools/uf2conv.py
@@ -431,9 +431,10 @@ def main():
             sys.stdout.flush()
             write_file(d + "/NEW.UF2", outbuf)
 
-        # Wait until serial port (if defined) re-appears, or 2s timeout unless UF2 drive direct upload
+        # Wait until serial port (if defined) re-appears, or 2s timeout.
+        # Skip the special value "UF2_Board" (see pluggable_discovery.py).
         try:
-            if args.serial != "UF2 Board":
+            if args.serial and args.serial != "UF2_Board":
                 timeout = time.time() + 2.0
                 while time.time() < timeout:
                     if os.access(args.serial, os.W_OK):


egnor:
(By the way, another approach here would be to use the same logic as in the tickle:
```python
if str(args.serial).startswith("/dev/tty") or str(args.serial).startswith("COM") or str(args.serial).startswith("/dev/cu"):
```
...but I figured I'd go with a smaller change.)

earlephilhower (changes requested):
Shouldn't this just be a `s/UF2 Board/UF2_Board/` without checking for one with a space.  That space was just a typo and it's never going to be called with `--serial "UF2 Board"`

egnor:
Changed. (Wasn't sure if something out there used the other spelling.)

earlephilhower (changes requested):
Looks like another formatting tool got to the commit.  Almost every line has minor changes.

egnor:
> Looks like another formatting tool got to the commit. Almost every line has minor changes.

Dammit. Sorry. Fixed.

_(slinks off to make editor only do that in projects which explicitly use ruff)_

earlephilhower (approved):
LGTM

earlephilhower:
Thanks, BTW, for the bundle of cleanup PRs!

PR #3520, opened 2026-09-05
Don't make arduino-cli wait for a serial port after picotool/openocd uploads

egnor:
This is a small quality-of-life change for `arduino-cli` users of the picotool and picoprobe/debugprobe upload methods.

Right now, entries in the generated `boards.txt` file all have `upload.wait_for_upload_port=true`, which causes arduino-cli to wait up to 5 seconds after flashing for a serial port to appear which is "similar enough" to whatever original port was specified for the upload operation (or a blank placeholder if none was). With many of our upload methods, there is not a serial port specified at all, so arduino-cli just times out for five seconds and prints:

```
New upload port:  (default)
```

This timeout is "just" five seconds (five seconds!), but it was bugging me. So this change disables `wait_for_upload_port` when using the picotool/picoprobe upload methods. (It's still on in the default uf2 case.)

The `upload.wait_for_upload_port` flag does these two things:
1. Together with `upload.use_1200bps_touch`, it makes arduino-cli itself do the 1200 bps touch and wait for the port to reappear before running the upload tool. This core never sets `use_1200bps_touch` (uf2conv.py does its own 1200bps touch, the other methods have other ways of rebooting a running core), so that path is inert here.
2. This property changes the time arduino-cli watches serial discovery after flashing to see the board come back (`New upload port: ...`) from 1 second to 5 seconds.

For (2), when the user *had* selected a serial port and it disappeared during the upload, arduino-cli forces the 5 second window regardless of the property (`detectUploadPort` in `commands/service_upload.go`), and returns as soon as a port with the same hardware ID shows up. So IDE users with a port selected are unaffected either way. The property only matters when **no port is selected**, and in that case nothing can ever match, so `true` just adds 4 seconds of idle time before `arduino-cli upload` returns.

**Why only for picotool / picoprobe?**

By the arguments above this is also harmless for the default (uf2conv) method! But I thought I would leave that alone, because at least in that case the port passed from arduino-cli to the uploader is actually used, and in theory in the future arduino-cli could actually pick up a newly appearing port (right now it never does and always times out).

I would be happy to just globally turn off this flag though. (To be clear, I'm okay abandoning this change if you feel like the risk isn't worth it.)

**Testing**

Measured on a Feather RP2040 with `arduino-cli upload --profile ...` (picotool method, no `--port`): 10.3 s → 6.3 s, with the "New upload port" line unchanged. Verified that the menu-level property overrides the board-level `true` for those methods and that the UF2 method still resolves to `true` (`arduino-cli compile --show-properties`).

`boards.txt` is regenerated with `tools/makeboards.py`; the only change is the two new lines per board (302 lines).


earlephilhower:
@egnor I just checked the Arduino spec and this what it says about the variable you're setting

> wait_for_upload_port causes the upload procedure to wait for the serial port to (re)appear before and after the upload. This is only used when use_1200bps_touch is also set. When set, after doing the 1200 bps touch, the development software will wait for a new serial port to appear and use that as the port for uploads. Alternatively, if the original port does not disappear within a few seconds, the upload continues with the original port (which can be the case if the board was already put into bootloader manually, or the the disconnect and reconnect was missed). Additionally, after the upload is complete, the IDE again waits for a new port to appear (or the originally selected port to be present).

So, since we never use that 1200bps reset flag for anything, this setting is ignored, no?

egnor:
> @egnor I just checked the Arduino spec and this what it says about the variable you're setting
> ...
> So, since we never use that 1200bps reset flag for anything, this setting is ignored, no?

Empirically this is for sure not true. Let me dig into the `arduino-cli` code.

egnor:
Ok here's what `arduino-cli` code actually does:

Pre-upload touch-and-wait is here, and like the doc says, `.wait_for_upload_port` only matters when `.use_1200bps_touch` is present, so is inert on all our boards:
https://github.com/arduino/arduino-cli/blob/v1.5.1/commands/service_upload.go#L495-L503

In post-upload port detection, `.wait_for_upload_port` is passed into `detectUploadPort` (as the `waitForUploadPort` parameter) regardless of `.use_1200bps_touch`:
https://github.com/arduino/arduino-cli/blob/v1.5.1/commands/service_upload.go#L472-L476

Finally, here in `detectUploadPort`, it uses that flag to decide whether to wait 1sec or 5sec:
https://github.com/arduino/arduino-cli/blob/v1.5.1/commands/service_upload.go#L641-L644

So if nothing else they have a spec/behavior mismatch and I can open a bug with them about that. Meantime, I still think removing the flag (either in the non-uf2conv case as proposed here, or the general case) is probably good here? Right now empirically it's a 4sec win for no cost in most (all??) cases. What I can't tell you is what the intended significance of all this is.

egnor:
Filed a bug with arduino-cli: https://github.com/arduino/arduino-cli/issues/3297

earlephilhower:
Gotcha, thanks for looking into it.  Seems like it's a fix now and a no-op when the arduino-cli behaviour matches the spec, so LGTM!

PR #3526, opened 2026-09-10
Fix BLEUUID String constructor

### LINKED ISSUE #3524: Bluetooth BLE creates all-zero UUIDs, or UUIDs that don't match what was requested
I'm trying to use the BLE stuff, and it's mostly working, but I'm seeing something strange.  The UUIDs don’t match what is requested.  Sometimes I get all zeroes, other times they are partially correct – the first 10 bytes match, the last 6 contain something, but not what was requested.

After poking at this for a while, I believe the problem is how the BLEUUID class’s constructor (see file arduino-pico/libraries/BLE/src/BLEUUID.h line 54) parses the input string – the first 10 bytes are parsed with sscanf as unsigned int, and the last 6 bytes are parsed as a "llx" format long long (8 bytes).  It looks like the function returns garbage in the variable e – if the garbage has magnitude too large it causes the all-zero case, but otherwise it just uses whatever garbage is there.

It doesn’t appear that there is anything wrong with the way sscanf is called, so I think there are a couple of possible causes:

First – a compiler bug.  That seems kind of unlikely to me.

Second – a google search on “raspberry pi pico c++ using Arduino IDE sscanf llx format fails” returns something plausible (and oddly specific!): 

The failure occurs because the Raspberry Pi Pico Arduino core (specifically Earle Philhower's arduino-pico core) compiles using Newlib-nano. [1] (https://github.com/earlephilhower/arduino-pico/issues/960), [2] (https://community.st.com/stm32cubeide-mcus-28/newlib-nano-is-unable-to-process-sscanf-input-after-the-1st-unsigned-char-value-hhu-or-hhx-123722)To minimize code size for microcontrollers, Newlib-nano strips out C99 and C11 format modifiers by default. This means the ll (long long) size modifier is not supported within standard sscanf() or sprintf() functions, causing %llx to fail silently or behave unpredictably. [1] (https://community.st.com/stm32cubeide-mcus-28/newlib-nano-is-unable-to-process-sscanf-input-after-the-1st-unsigned-char-value-hhu-or-hhx-123722)

I don’t know enough to confirm whether that’s true, but assuming it is, the google result also includes a possible fix ...

If you like, I could try it out and submit a pull request ...

Thanks!

earlephilhower:
Nano-printf doesn't sscanf %ll properly, so manually parse the UUID byte-by-byte.

Fixes #3524

### DIFF
--- a/libraries/BLE/src/BLEUUID.h
+++ b/libraries/BLE/src/BLEUUID.h
@@ -41,38 +41,28 @@ public:
     }
 
     BLEUUID(String s) {
-        const char *str = s.c_str();
         is16 = false;
         bzero(uuid128, sizeof(uuid128));
         // 0000FF11-0000-1000-8000-00805F9B34FB
-        if (strlen(str) != 36) {
+        if (s.length() != 36) {
             return;
         }
-        // Maybe not prettiest or most efficient, but not critical in a constructor
-        unsigned int a, b, c, d;
-        unsigned long long e;
-        if (5 != sscanf(str, "%x-%x-%x-%x-%llx", &a, &b, &c, &d, &e)) {
-            return;
-        }
-        if ((b > 0xffff) || (c > 0xffff) || (d > 0xffff) || (e > 0xffffffffffffll)) {
-            return;
-        }
-        uuid128[15] = e >> 0ll;
-        uuid128[14] = e >> 8ll;
-        uuid128[13] = e >> 16ll;
-        uuid128[12] = e >> 24ll;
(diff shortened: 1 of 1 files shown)


earlephilhower:
````
#include <BLE.h>

void setup() {
  delay(5000);
  BLEUUID u("0300FF11-8800-1000-8000-00805F9B34FB");
  Serial.println(u.toString());  
}

void loop() {
}
````

Now gives proper result
````
0300ff11-8800-1000-8000-00805f9b34fb
````

PR #3528, opened 2026-09-14
LEAmDNS: defer packet processing while a user call is sending

bdraco:
`UdpContext` calls the LEAmDNS receive handler from the network driver. On the Pico W that is the CYW43 low priority interrupt, and also the user thread, because `async_context_threadsafe_background` runs pending driver work when the outermost lwIP lock is released, including inside the `udp_sendto` that `UdpContext::sendTimeout` makes. Either way `_callProcess()` can re-enter the responder while `update()`, `close()` or `announce()` is in the middle of a send on the same `UdpContext`. The ESP8266 core had the same kind of double free ([esphome/esphome#18760](https://github.com/esphome/esphome/issues/18760)).

Calls that send now set `m_bUserCallActive` through a small scope object; while it is set, `_callProcess()` only notes that packets are waiting, and the outermost scope parses them on the user thread when it ends. The scope covers `update()`, `close()`, `announce()`, `removeService()` and `_sendMDNSServiceQuery()`, which leaves the answer wait in `queryService()` unguarded.

SimpleMDNS avoids the problem, but it can't replace LEAmDNS for us. lwIP [limits each TXT item to 63 bytes and a service's TXT data to 255 bytes](https://github.com/lwip-tcpip/lwip/blob/77dcd25a72509eb83f72b033d219b1d40cd8eb95/doc/mdns.txt#L95-L97), and ESPHome's records (friendly name, encryption, project, import URL) go past that, which already drops records on LibreTiny ([esphome/esphome#11281](https://github.com/esphome/esphome/issues/11281)); the lwIP fix ([lwip-tcpip/lwip#83](https://github.com/lwip-tcpip/lwip/pull/83)) is still open.

Tested on a Pico W running ESPHome, flooding it with mDNS queries while restarting it so the flood covered the window where ESPHome calls `update()`. Before this change the receive handler re-entered `update()` in every window and single `update()` calls stalled for up to 4 s; with it those packets were deferred instead and the longest call was under 1 s. Neither build crashed during the test. After every restart the device still announced itself and resolved with all of its TXT records, and a query sent once a second after boot got an answer every second with this change, while the unpatched build went silent for up to 21 s.


earlephilhower:
Thanks for the debug and fix on this!  I've not had to do much with LEAmDNS here other than pull in from the 8266, so let me just look at the code tonight before merging.  But at first glance this patch seems very clean and neat.

CI fails seem to be GH Actions going bad like usual.

earlephilhower (approved):
Very nice way of working around the issue.  LGTM.  Thanks!

bdraco:
Thanks

PR #3529, opened 2026-09-14
Fix unwrapped functions that cause a SIGTRAP if Wifi.begin is called …

wgm4321:
NOTE: These changes were proposed by Claude.
In short: if  wifi.begin is call off the initial task results in a SIGTRAP.   I have NO understanding of how LWIP anbd the CYN43 driver work.    Please review in DETAIL!
Claude suggests this description:
WiFi.begin() → CYW43::begin() calls cyw43_ioctl()/cyw43_wifi_update_multicast_filter() directly, unwrapped, while cyw43_wifi_join() right after them is wrapped; calling WiFi.begin() from any FreeRTOS task other than the internal "LWIP" task races those two calls against cb_cyw43_do_poll() on that task, corrupting cyw43 driver state — you hit it as a hardfault inside memcmp() in cyw43_ll_wifi_update_multicast_filter() when moving a reconnect-retry WiFi.begin() call from setup() into a periodic FreeRTOS task. Worth linking your crash trace as reproduction evidence.:

````
Stack trace of the SIGTRAP:
isr_hardfault@0x1000311c (Unknown Source:349)
<signal handler called>@0xfffffffd (Unknown Source:0)
memcmp@0x1002bb9c (\memcmp.dbgasm:15)
cyw43_ll_wifi_update_multicast_filter@0x1002acf8 (\cyw43_ll_wifi_update_multicast_filter.dbgasm:69)
cyw43_wifi_update_multicast_filter@0x10029216 (\cyw43_wifi_update_multicast_filter.dbgasm:10)
CYW43::begin@0x1002f304 (c:\Users\Internet\.platformio\packages\framework-arduinopico\libraries\lwIP_CYW43\src\utility\CYW43shim.cpp:85)
LwipIntfDev<CYW43>::begin@0x10012e6c (c:\Users\Internet\.platformio\packages\framework-arduinopico\libraries\lwIP_Ethernet\src\LwipIntfDev.h:399)
WiFiClass::_beginInternal@0x100131e6 (c:\Users\Internet\.platformio\packages\framework-arduinopico\libraries\WiFi\src\WiFiClass.cpp:107)
WiFiClass::begin@0x1001321e (c:\Users\Internet\.platformio\packages\framework-arduinopico\libraries\WiFi\src\WiFiClass.cpp:150)
dcc::WifiManager::poll@0x100069c0 (c:\Users\Internet\OneDrive\Documents\PlatformIO\Projects\RP2350-DCC\src\net\wifi_manager.cpp:84)
(anonymous namespace)::wifiManagerTask@0x1000556a (c:\Users\Internet\OneDrive\Documents\PlatformIO\Projects\RP2350-DCC\src\main.cpp:269)
prvDoorbellInterruptHandler@0x10024664 (c:\Users\Internet\.platformio\packages\framework-arduinopico\FreeRTOS-Kernel\portable\ThirdParty\GCC\RP2350_ARM_NTZ\non_secure\port.c:1711)
````


earlephilhower (commented):
Do you have a persistently failing small MCVE (test code) case you can share?

My own quick testing doesn't show any issues with the existing code if I take the `WiFiClient` example and rename `setup/loop` to `setup1/loop1`, pinning them to the 2nd core.  The `ioctl` calls, IIRC, will use the async context which should be safe for either core.

Also, there are no signals on RP2040, it's not running UNIX.  Your trace does show a HW exception, but without more digging the stack trace itself doesn't really say why.

That said, I *think* the change proposed is safe.  The core is the only bit of code to call the multicast-filter, so all's fine there.  The ioctl call is used in the SDK and in the core, but it's safe to call a wrapped fcn from another, either the barebones LWIPMutex will handle it properly, or the LWIP task in FreeRTOS should short-circult and just call the fcn, not place it on the work queue (where it'd deadlock the task).

earlephilhower (approved):
I don't have a failing case to test against, but I've tested what I can by modding the examples to use core1 and FreeRTOS and at worst this is a no-op.  Logically it seems like a good thing to do, so I'll pull this in and we'll see how it's received on the next release.

PR #3532, opened 2026-09-15
Bound the 1200-baud BOOTSEL request so a failed reboot is not a hang

hippo5329:
checkSerialReset() tears USB down and then relies on "WDT will fire here", but nothing arms a watchdog unless the sketch called rp2040.wdt_begin().

On RP2040 that is survivable: reset_usb_boot() is the ROM entry point and does not come back. On RP2350 the SDK maps it to

    rom_reboot(REBOOT2_FLAG_REBOOT_TYPE_BOOTSEL |
               REBOOT2_FLAG_NO_RETURN_ON_SUCCESS, 10, ...)

and discards the result, because the wrapper is declared noreturn. When that reboot does not take, USBCTRL_IRQ is already disabled and the USB block has already been reset, so the while(1) below is permanent. The board stays enumerated, is mute at every baud, and answers no control transfer -- the host sees ETIMEDOUT on the very SET_CONTROL_LINE_STATE that triggered the touch, so a successful touch and a hung one are indistinguishable from the host. Only the RESET button recovers it.

Observed on a Pico 2 doing repeated flash cycles: most touches reach BOOTSEL, an occasional one leaves the board in that state until it is reset by hand.

Arming the watchdog immediately before the call makes the existing comment true in every case, and costs nothing on the working path: a successful reset_usb_boot() reprograms the watchdog for its own 10 ms reboot long before this one could fire.


earlephilhower (changes requested):
Can you please provide an MCVE that shows the problem?

If I understand what you're saying, Pico2 uploads should fail 99% of the time because most sketches don't ever use the WDT.  But I know that's not the case.

The comment `// wdt will fire here` is probably old and out of date.  In the initial SDK release I *think* that the reset_usb_boot actually returned after setting the WDT or whatever (or I misunderstood it to begin with).

If `reset_usb_boot` doesn't work for some cases on the RP2350, then that's a SDK problem and the R.PI guys will still need an MCVE to reproduce it.

